PR #52, opened 2016-04-30
Geocoding Observerable NullPointerException Fix

### LINKED ISSUE #51: NullPointerExcetion When Using MapboxGeocoding.getObservable()
The `service` object is always null when `MapboxGeocoding.getObservable()` is called which results in a NullPointerException.  This is due to the the [internal `service` object being referenced directly](https://github.com/mapbox/mapbox-java/blob/c4215912dfd22ee3ce8a189b8ad409cc4c25b472/libjava/lib/src/main/java/com/mapbox/services/geocoding/v5/MapboxGeocoding.java#L97) instead of calling `getService()` which is what's responsible for the object's creation.

```java
java.lang.NullPointerException: Attempt to invoke interface method 'rx.Observable com.mapbox.services.geocoding.v5.GeocodingService.getObservable(java.lang.String, java.lang.String, java.lang.String, java.lang.String, java.lang.String)' on a null object reference
    at com.mapbox.services.geocoding.v5.MapboxGeocoding.getObservable(MapboxGeocoding.java:96)
```

screen shot 2016-04-30 at 5 29 33 pm

/cc @zugaldia

bleege:
Closes #51

### DIFF
--- a/libjava/lib/src/main/java/com/mapbox/services/directions/v4/MapboxDirections.java
+++ b/libjava/lib/src/main/java/com/mapbox/services/directions/v4/MapboxDirections.java
@@ -99,7 +99,7 @@ public class MapboxDirections implements MapboxService<DirectionsResponse> {
         // No need to recreate it
         if (observable != null) return observable;
 
-        observable = service.getObservable(
+        observable = getService().getObservable(
                 builder.profile,
                 builder.getWaypoints(),
                 builder.accessToken,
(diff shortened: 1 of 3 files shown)


zugaldia:
@bleege Looking good, I also noticed the same error in MapboxDirections v4 too. Two birds, one PR.

bleege:
💥

PR #64, opened 2016-05-18
mapmatching - initial implementation

ivovandongen:
Initial implementation for the Map Matching service.

Could do with some more unit tests on response body eventually.


zugaldia:
Reviewing now.

/cc: @cammace

zugaldia:
@ivovandongen The implementation is looking good.

Like we discussed, I'm only undecided about one thing. Right now we mimic the API response in Java (like we do with other services). However, in this case, in order for that to work we need to change the constructor visibility of `FeatureCollection` to `protected` in order to be able to extend it to include the `code` field.

This is fine, but I'm thinking of making an exception here and changing the deserializer to have the `FeatureCollection` object as a field in the `MapMatchingResponse` instead. Let me think about it.

zugaldia:
All good, thanks @ivovandongen for the contribution! Merging now.

PR #89, opened 2016-06-30
Geocoder client improvements

### LINKED ISSUE #57: Geocoder wikidata
Geocoder now includes wikidata information. MAS needs to include getter and setter methods. [API-documentation](https://www.mapbox.com/api-documentation/#response-format)

cc: @zugaldia

### LINKED ISSUE #59: Support geocoding bbox parameter

zugaldia:
This PR brings a number of fixes and improvements :
- Fixes #57: Support geocoder `wikidata` response value
- Fixes #59: Support geocoder `bbox` request parameter
- `dataset` request parameter renamed to `mode`
- Bring other previously unsupported request parameters (`country`, `autocomplete`)
- Update test fixtures to current response format
- Response classes now explicit Carmen's implementation

/cc: @mattficke @cammace


zugaldia:
Local tests are passing (Bitrise is acting up). Merging.

bleege:
Per discussion with @zugaldia this afternoon we'll need to bump MAS to `2.0.0` instead of `1.2.0` as the data model for the GeoJSON is a breaking change.

zugaldia:
I'm torn here:
* Other [Mapbox](https://github.com/mapbox/mapbox-sdk-py/blob/master/CHANGES) [SDKs](https://github.com/mapbox/MapboxDirections.swift/releases) are not bumping their versions yet
* Geocoding parameters got updated but we still are in `v5` (server-side)
* The low-level `libjava` changes don't affect the higher level Android functionality

How about we maintain 1.2.x and flag the changes in the release notes?

bleege:
Yeah, I'm torn with ya @zugaldia.  This is definitely a server side thing in terms of a driving the change, and they're still calling it V5 on that end.  That said, the reason why the Python and iOS SDKs can get by not bumping a full rev number is that they're still under the magical `1.0.0` release were breaking changes can be expected (`0.9.0` for Python and `0.6.0` for iOS) and MAS is now at `1.1.0`.  It's your call as I know the next semver is coming very soon down the line.

PR #91, opened 2016-06-30
Customize user agent string

### LINKED ISSUE #74: Customize user agent string
This library should set the `User-Agent` HTTP header on requests it sends, so that Mapbox Studio’s Stats page can identify this library under the “Browsers” column instead of saying “Other”.

/cc @zugaldia

zugaldia:
For now I'm setting: `MapboxAndroidServices/1.0`.

@1ec5 : Do you know if there is a better rule for this?

Fixes #74.

/cc: @cammace

### DIFF
--- a/libjava/lib/src/main/java/com/mapbox/services/commons/MapboxService.java
+++ b/libjava/lib/src/main/java/com/mapbox/services/commons/MapboxService.java
@@ -1,6 +1,10 @@
 package com.mapbox.services.commons;
 
+import com.mapbox.services.Constants;
+import com.mapbox.services.commons.utils.TextUtils;
+
 import java.io.IOException;
+import java.util.Locale;
 
 import okhttp3.OkHttpClient;
 import okhttp3.logging.HttpLoggingInterceptor;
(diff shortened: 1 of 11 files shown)


libjava/lib/src/main/java/com/mapbox/services/commons/MapboxService.java
```diff
@@ -14,6 +14,8 @@
  */
 public abstract class MapboxService<T> {
 
+    public final static String HEADER_USER_AGENT = "User-Agent: MapboxAndroidServices/1.0";
```
1ec5: Is there a way to dynamically update the version to match the library’s actual version? Otherwise, the source code should be updated as part of the build or packaging process.

1ec5:
On the four Apple platforms, the Mapbox map SDK and Swift libraries all use a similar format. But in addition to the library’s name and version, the user agent string also indicates the name and version of the application (which can be useful for identifying bugs or abuse in server logs), as well as the device’s architecture and operating system version. So an equivalent user agent string here would look something like:

```
SanDiego/1.2.3 MapboxAndroidServices/1.2.0 Android/6.0.1 (mips64)
```

Finally, libjava can be used in Java SE applications, right? It wouldn’t make sense for the user agent string to report “Mapbox Android Services” in that case.

zugaldia:
@1ec5 Much appreciated, these are great tips. Based on this I've done the following changes:

* I've replaced `MapboxAndroidServices` with `MapboxJava` (the name of the repo) to reflect that the HTTP requests actually come from the Java library (this is also true on Android).
* Added (manually for now) the library version number.
* Added OS information.

The string currently looks like this: `MapboxJava/1.2.0 Mac OS X/10.11.5 (x86_64)` and it's added as a header to every request.

It doesn't include app info for now as this is a Java library that works outside Android. If useful, we could explore in the future ways to modify this header when invoked from an Android environment.

1ec5:
:+1:

Issue #87, opened 2016-06-30
Map Matching Activity in Test App


zugaldia:
Map matching support is coming to v1.2.0 via https://github.com/mapbox/mapbox-java/pull/64, we need to include a new activity to show this functionality in action.

One potential source for a sample trace is http://www.openstreetmap.org/traces.

/cc: @cammace

zugaldia:
Fixed via https://github.com/mapbox/mapbox-java/pull/94.

PR #94, opened 2016-07-01
[READY FOR REVIEW] Map matching example

cammace:
Placeholder for the map matching example


cammace:
I can change the trace file if we need something showing more of a difference. Otherwise this example should be ready to merge

cc: @zugaldia

cammace:
closes #87

zugaldia:
@cammace Is the trace file in https://www.mapbox.com/blog/map-matching-api/ any better?

cammace:
Here is what the demo looks like now, you get a perfect match when precision is maxed out at 10. I added a check in the code to make sure the response is "OK", `if(response.code() == 200)`.

ezgif com-video-to-gif 1

cammace:
Changed the trace again, added a missing log output when/if an error occurs and commented more of the code. The trace now looks like:

ezgif com-video-to-gif 3

Examples ready to 🚢, merging now!

zugaldia:
@cammace Love the new route.

Issue #69, opened 2016-05-19
MapboxStaticImage - add precision setter on builder


ivovandongen:
In order to make the returned images better cacheable on the client, it would be nice to set the precision in decimals on the Builder instead of manually rounding the latitude/longitude.

Issue #67, opened 2016-05-19
MapboxStaticImage - validate image dimensions on value range


ivovandongen:
Currently the image dimensions are only validated on `null` but not the correct range. With some image loading libraries this results in the image just not appearing and no obvious error being logged. (Maybe because the api returns a status code 422, which is not that common).

Issue #65, opened 2016-05-19
MapboxStaticImage - provide option to specify position as LatLng


ivovandongen:
We have been using the [mapbox-gl-native](https://github.com/mapbox/mapbox-gl-native) for our main apps and our now integrating the static map api. It would be great if we can use the same `LatLng` location objects here to keep consistency. Especially as we built out our support methods based on the `LatLng` and like to re-use those.

zugaldia:
Unfortunately we don't have access to `LatLng` objects from the main SDK in MAS. But what we can do is to add a `setLocation(Position position)` method in addition to the the individual coordinate setters.

PR #96, opened 2016-07-02
Improvements to Static Image API service

zugaldia:
This PR fixes three proposed improvement to the Static Image API service. Namely:

1. Provide option to specify position as Position #65
2. Validate image dimensions on value range #67
3. Add precision setter on builder #69

/cc: @ivovandongen @cammace


zugaldia:
Merging.

PR #98, opened 2016-07-04
Fixed parameter ordering issue

lyptt:
Hi there!

I've noticed a bug in the 1.2.0 SNAPSHOT with the MapboxGeocoding class:

```java
    //MapboxGeocoding.java:67
    public Call<GeocodingResponse> getCall() {
        // No need to recreate it
        if (call != null) return call;

        call = getService().getCall(
                getHeaderUserAgent(),
                builder.getMode(),
                builder.getQuery(),
                builder.getCountry(),
                builder.getAccessToken(),
                builder.getProximity(),
                builder.getGeocodingTypes(),
                builder.getAutocomplete(),
                builder.getBbox());

        return call;
    }
```

This all appears fairly mundane until you realise that the 4th parameter of `getCall()` is `accessToken`, not `country`, which causes any request in this library to fail due to having a bad access token.

This fix swaps around the parameters in the call to  `getCall()`, so it should now match up to the retrofit service.


zugaldia:
Great eye @lyptt , this was totally my fault, I believe I broke this as part of https://github.com/mapbox/mapbox-java/pull/89. Merging your fix now.

zugaldia:
One more switch in https://github.com/mapbox/mapbox-java/commit/4ad3195791b709353e2762e374834ffbee53226b to update the `getObservable` method as well.

lyptt:
Thanks for the quick update! Saves me using dirty reflection hacks as a workaround 😉

zugaldia:
@lyptt We definitely wouldn't want to force you to use reflection =:-)

PR #99, opened 2016-07-05
Fix geocoder issues in libandroid

cammace:
This PR resolves issues in libandroid since updates were made in libjava. 

cc: @zugaldia


zugaldia:
@cammace Looks good to me, thanks for updating the code. I'll merge it now.

BTW, let's remember adding this in the release notes. This is a great example of how to update the code to the latest geocoder models (ref https://github.com/mapbox/mapbox-java/pull/89#issuecomment-230560942).

PR #103, opened 2016-07-06
Replaced emerald for outdoor style in static image

cammace:
cc: @zugaldia


zugaldia:
Sounds great, merging.

Issue #95, opened 2016-07-01
Decouple the snapshot creation between libandroid and libjava 


cammace:
When something changes in libjava, we are breaking libandroid because we are using the SNAPSHOT that's only updated nightly. 

cc: @zugaldia

zugaldia:
I'm thinking that during development we could have something like this in `settings.gradle`:

```
include ':libjava'
project(':libjava').projectDir = new File(settingsDir, '../libjava')
```

and then in the `build.gradle` file something like:

```
dependencies {
    compile project(':libjava')
}
```

That way we'd be developing and building from master. Only at release time, we'd force the dependency to match a published version of the library.

cammace:
Just tried your suggestion out briefly @zugaldia and it looks promising. One thing to note is that I had to change the file path to `'../libjava/lib'`.

zugaldia:
@cammace Glad to hear. PR'ed in https://github.com/mapbox/mapbox-java/pull/113.

zugaldia:
> It also now includes the libjava module in the libandroid project?

Yeah, it does for me too. Even more convenient :-)

PR #113, opened 2016-07-11
Use master branch from libjava during dev

zugaldia:
Helps with https://github.com/mapbox/mapbox-java/issues/95.

/cc: @cammace


cammace:
Just tested adding a random method in libjava and was instantly able to use it in libandroid. This is fantastic! Ready to merge when you are @zugaldia

PR #122, opened 2016-07-13
Test app using account manager

cammace:
Added application object and removed all mapview set access token usages.

cc: @zugaldia


zugaldia:
Merging.

PR #124, opened 2016-07-13
Fixes checkstyle warning

cammace:
I've run through and fixed most of the warning found in the MAS testapp.


cammace:
Closing since we are switching to a different checkstyle.

Issue #76, opened 2016-06-01
Add Turf.within() Module


bleege:
This will allow for POIs to be filtered by a given region.

https://github.com/Turfjs/turf/tree/master/packages/turf-within

bleege:
Requires Turf.inside() as a dependency.

https://github.com/Turfjs/turf/tree/master/packages/turf-inside

PR #128, opened 2016-07-14
Bring Turf `within` to libjava

zugaldia:
In order to bring `within` to Java, we need to bring its dependencies:
- [x] `invariant`
- [x] `inside`
- [x] `within`

Fixes https://github.com/mapbox/mapbox-java/issues/76.

/cc: @cammace @bleege


zugaldia:
@cammace Implementation is now finished. Do you wanna add a test activity before we finally merge?

bleege:
> Do you wanna add a test activity before we finally merge?

From the outsider with an opinion department, I strongly vote YES.  😄

cammace:
I've added an example using turf inside:

ezgif com-video-to-gif

I also added an additional method making it easier to directly pass in position's instead of having to convert to point and polygon first (test was added also). I'm ready to merge.

PR #132, opened 2016-07-15
Turf midpoint added

cammace:
Added turf midpoint and all the test. Also added an additional method that converts position to point making it easier on the user and added a custom test to make sure it works.

cc: @zugaldia


cammace:
added an activity in the testapp showing off Turf midpoint. The markers currently don't line up perfectly due to the anchoring bug in `4.1.0` https://github.com/mapbox/mapbox-gl-native/issues/5513

ezgif com-video-to-gif 2

zugaldia:
LGTM. Merging.

PR #133, opened 2016-07-15
Geocoding autocomplete widget bbox

### LINKED ISSUE #130: Goecoding autocomplete widget should expose bounding box
Now that `setBbox();` is exposed in the geocoder in #89, users should be able to set a bounding box using the Android autocomplete widget. We should also make sure all other available options are exposed.

cc: @zugaldia

cammace:
Closes #130. I've added `setbbox()` to the geocoding widget. 

cc: @zugaldia


cammace:
Added a method that allows you to pass in two position called northeast and southwest instead of:
`setBbox(double minX, double minY, double  maxX, double  maxY)`

PR #139, opened 2016-07-18
Added missing Turf javadoc

cammace:
All Turf methods now have appropriate javadoc. Going ahead and merging!

cc: @zugaldia


zugaldia:
@cammace Fan. Tas. Tic.

PR #142, opened 2016-07-19
Javadoc added and fixed

cammace:
libjava now produces 0 errors or warnings. The problem was missing tags and tag descriptions among other things. Every public method now has javadoc.

I've also added a task in gradle to build the java doc:

```groovy
task JavadocGeneration(type: Javadoc) {
    source = sourceSets.main.allJava
    failOnError = false
    classpath = configurations.compile
}
```

cc: @zugaldia

### DIFF
--- a/libjava/lib/build.gradle
+++ b/libjava/lib/build.gradle
@@ -16,5 +16,11 @@ dependencies {
     testCompile 'com.squareup.okhttp3:mockwebserver:3.3.1'
 }
 
+task JavadocGeneration(type: Javadoc) {
+    source = sourceSets.main.allJava
+    failOnError = false
+    classpath = configurations.compile
+}
+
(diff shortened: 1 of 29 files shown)


zugaldia:
> libjava now produces 0 errors or warnings.

💥 💥 💥 💥 💥 

> I've also added a task in gradle to build the java doc

I thought we already had a task in place (that we've been using to generate the docs in the past). Why is this one required?

cammace:
I saw one in the libandroid but not for libjava build.gradle file. If we already have one performing the same task, i'll remove though.

zugaldia:
@cammace Yeah, it seems that the `java` plugin already includes a `javadoc` task by default: https://docs.gradle.org/current/userguide/java_plugin.html#sec:javadoc. 

For Android, I believe I added it manually because it didn't include the equivalent task.

PR #153, opened 2016-07-25
Enable CarmenFeature to be used as a standalone object

### LINKED ISSUE #148: Allow the standalone use of the geocoding model objects
Opening this ticket to discuss the issue we are currently having updating to the latest version of MAS and correct me if I'm wrong @bleege. Previously we have been saving search history like so:

```java
public GeocodingFeature adaptTo() {
        GeocodingFeature feature = new GeocodingFeature();
        feature.setText(key);
        feature.setPlaceName(key);

        FeatureGeometry geometry = new FeatureGeometry();
        List<Double> coordinates = new ArrayList<>();
        coordinates.add(longitude);
        coordinates.add(latitude);
        geometry.setCoordinates(coordinates);

        feature.setGeometry(geometry);
        return feature;
    }
```
With the latest MAS release, `1.2.1` the data model is setup so that it can only be worked with through GSON and not as standalone objects making the above code not work anymore. Some options here would be to just add the models in the app and use them instead of the ones found in MAS or make the ones in MAS usable. 

cc: @zugaldia @bleege

zugaldia:
Fixes #148.

@cammace Feel free to merge if this is what you're looking for.

### DIFF
--- a/libjava/lib/src/main/java/com/mapbox/services/commons/geojson/Feature.java
+++ b/libjava/lib/src/main/java/com/mapbox/services/commons/geojson/Feature.java
@@ -17,9 +17,9 @@ import com.mapbox.services.commons.models.Position;
 public class Feature implements GeoJSON {
 
     private final String type = "Feature";
-    private final Geometry geometry;
+    private Geometry geometry;
     private JsonObject properties;
-    private final String id;
+    private String id;
 
     /**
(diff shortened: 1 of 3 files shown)


cammace:
Thanks for doing this @zugaldia, we might need to make `GeocodingResponse` public as well. Merging and will check.

PR #155, opened 2016-07-26
Bring geojson-tidy to Java library

### LINKED ISSUE #151: Bring geojson-tidy to Map Matching client
The Map Matching API has a limit of 100 coordinates for input. As [explained in the docs](https://www.mapbox.com/api-documentation/#retrieve-a-match), there're some some pre-processing tips that can help achieve better results:

> * Clusters of points (like a person waiting at a train crossing for a few minutes) often don't add more information to a trace and should be removed.
> * Map matching works best with a sample rate of 5 seconds between points. If your trace has a higher sample rate, you may want to downsample your trace.
> * geojson-tidy removes clusters and re-samples traces, which can be useful for preprocessing.

This is something the SDK can help with by bringing `geojson-tidy` as a convenience method to our Map Matching client.

/cc: @cammace

zugaldia:
Fixes #151.

/cc: @cammace


zugaldia:
Merging after CI approval.

Issue #154, opened 2016-07-26
libandroid snapshot not working


cammace:
[Documentation](https://www.mapbox.com/android-sdk/#android_library) points to using android-services snapshot but it can't be found when building project using:
```groovy
compile ('com.mapbox.mapboxsdk:mapbox-android-services:2.0.0-SNAPSHOT@jar'){
        transitive=true
    }
```
[Checked here](https://oss.sonatype.org/content/repositories/snapshots/com/mapbox/mapboxsdk/mapbox-android-services/2.0.0-SNAPSHOT/), but didn't see anything unusual. Using the java-services snapshot works as expected:

```groovy
compile ('com.mapbox.mapboxsdk:mapbox-java-services:2.0.0-SNAPSHOT@jar'){
        transitive=true
    }
```

cc: @zugaldia

zugaldia:
> com.mapbox.mapboxsdk:mapbox-android-services:2.0.0-SNAPSHOT@jar

It should be `aar`:

> com.mapbox.mapboxsdk:mapbox-android-services:2.0.0-SNAPSHOT@aar

Did that do the trick?

cammace:
sorry I was trying with:
```groovy
compile ('com.mapbox.mapboxsdk:mapbox-android-services:2.0.0-SNAPSHOT@aar'){
        transitive=true
    }
```
and getting this error
```
Error:Could not find libandroid:mapbox-java-services:unspecified.
Required by:
    sirius-android:Sirius:unspecified > com.mapbox.mapboxsdk:mapbox-android-services:2.0.0-SNAPSHOT
<a href="searchInBuildFiles">Search in build.gradle files</a>
```

zugaldia:
This was my bad. When I released `1.2.1` I left the following line active in `build.gradle`:

```
compile project(':libjava')
```

Which should _only_ be used during local development. For Maven deps to work, we need to use this instead when building the actual package:

```
// Mapbox Java Services (release)
compile ('com.mapbox.mapboxsdk:mapbox-java-services:2.0.0-SNAPSHOT@jar') {
    transitive=true
}
```

I'm gonna fire another `SNAPSHOT` now to double check.

zugaldia:
@cammace Confirmed that the `SNAPSHOT` built on `Wed Jul 27 21:43:47 UTC 2016	52408` works as expected. Reopen otherwise.

PR #160, opened 2016-07-27
Make snapshots build from mapbox-java-services:2.0.0-SNAPSHOT@jar

zugaldia:
Instead of from FS.

Fixes https://github.com/mapbox/mapbox-java/issues/154.

/cc: @cammace


zugaldia:
CI approved, merging.

PR #161, opened 2016-07-28
fixes #156: add Feature.hasNonNullValueForProperty()

### LINKED ISSUE #156: Convenient methods for geocoder response properties
The geocoder offers telephone numbers, maki icon suggestions, wikidata, address, and category depending on the geocoder filter search. To access these currently you have to do something like this in your app:

```java
if(feature.getProperties().has("tel")) {
    String phone = feature.getProperties().get("tel").getAsString();
}
```

But this doesn't check if the value's null ([more info here](https://github.com/mapbox/sirius-android/issues/329)). We should expose methods that make accessing these elements easier and gives this feature more exposure in the SDK.

cc: @zugaldia

incanus:
Again, bear with me and my meager Java skills for a moment. 

Would something like this work? The caller could more simply check for `null` like so: 

```java
if(feature.hasNonNullValueForProperty("tel")) {
    String phone = feature.getStringProperty("tel");
}
```

Versus something like this now: 

```java
if(feature.getProperties().has("tel")) {
    String phone = feature.getProperties().get("tel").getAsString();
    // or String phone = feature.getStringProperty("tel");
}
```

Which can throw a`JsonNull` exception. 

@zugaldia @cammace

### DIFF
--- a/libjava/lib/src/main/java/com/mapbox/services/commons/geojson/Feature.java
+++ b/libjava/lib/src/main/java/com/mapbox/services/commons/geojson/Feature.java
@@ -283,4 +283,15 @@ public class Feature implements GeoJSON {
     public boolean hasProperty(String key) {
         return getProperties().has(key);
     }
+
+    /**
+     * Convenience method to check for a member by name as well as non-null value.
+     *
+     * @param key name of the member
+     * @return true if member is present with non-null value, false otherwise.
+     * @since 2.0.0
+     */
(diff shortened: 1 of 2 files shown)


incanus:
Screwed up branch a bit. Fix was in https://github.com/mapbox/mapbox-java/commit/62244ee9ac83e59872bb27641e213e5c7e5678d1.

zugaldia:
@incanus 👍

PR #165, opened 2016-08-04
Remove RxJava dependencies

### LINKED ISSUE #131: Support RxJava
/cc: @cammace @bleege

zugaldia:
The plan is to extract this functionality as an optional module for devs.

Fixes #131.

/cc: @ivovandongen


zugaldia:
@cammace Could you compute the new method count when you have a chance.

@ivovandongen CI has approved, please merge whenever you're ready.

cammace:
It isn't possible to check the method count using the website with the SNAPSHOT but I added a gradle plugin and ran it for the libandroid, results can be found in #170

bleege:
@zugaldia Just checking, but do we have an example of how to make use of RxJava / RxAndroid with MAS now that it's formally been removed the from the library?

PR #169, opened 2016-08-05
Be explicit about features in the README

tmcw:
cc @cammace for the review


cammace:
I've added the rest of the features mapbox java has for devs. Ready to merge when you are @tmcw

zugaldia:
@cammace Let's add under the Android section replacement for the built-in Android `Geocoder`.

tmcw:
Made the Geocoder mention more explicit. Merging once green.

zugaldia:
> Geocoder input view with autocompletion (drop-in replacement for stock Geocoder view)

Sorry, my bad explaining, this is not accurate. These should be two different items:

- Ready-to-go geocoder input view with autocompletion.
- Drop-in replacement for stock Android Geocoder object.

The first item is a UI element, the second one is a drop in replacement for the [`Geocoder`](https://developer.android.com/reference/android/location/Geocoder.html) object.

Issue #171, opened 2016-08-10
GeoJson: Add factory methods to create Geometries from primitive arrays


ivovandongen:
When using the geojson classes from jni, there is a performance hit as all factory methods are now neatly typed. For example: `MultiPolygon#fromCoordinates(List<List<List<Position>>> coordinates)`. This means creating multiple nested arrays of `Position` objects and turning all of those into `List`s. This is great from Java, very messy from jni.

Let's add methods like: `MultiPolygon#fromCoordinates(double[][][][] coordinates)` (already lost count of the `[]`'s)

cc @zugaldia

PR #172, opened 2016-08-10
Add support for primitive arrays

zugaldia:
Fixes https://github.com/mapbox/mapbox-java/issues/171

@ivovandongen Is it something like this that you had in mind? The conversions feel hacky for some reason, I want to see if there's a better way to accomplish this.

Either way, before merging, this needs:

- [ ] Tests.


ivovandongen:
Thanks @zugaldia

ivovandongen:
Merged 

cc @mapbox/android

ivovandongen:
@zugaldia I think this is perfectly fine as it reflects the geojson specification. I wouldn't advertise it for "normal" use cases, but in our case (jni), it is preferable.

PR #176, opened 2016-08-14
reworded libandroid features in readme

cammace:
Fixes `libandroid` features in readme pointed out by @zugaldia in #169.


zugaldia:
@cammace 👍

PR #180, opened 2016-08-23
Bring RouteUtils class to simplify work with RouteLeg objects

zugaldia:
This PR brings a new `RouteUtils` class that brings some Turf-powered methods to obtain situational information from a `RouteLeg` object. To some extend, this a partial refactor and port to Java of @bsudekum 's work on [`navigation.js`](https://github.com/mapbox/navigation.js).

Particularly, this PR includes the following methods:

- `isInStep`: Given a position, a route, and a step index, determines if the position is close to the given step (within a customizable threshold).
- `getDistanceToStep`: Given a position, a route, and a step index, determines the distance between the position and the closest point in the given step.
- `getSnapToRoute`: Given a position, a route, and a step index, determines the closest point in the given step
- `isOffRoute`: Given a position and a route, determines if the position is off-route (within a customizable threshold).
- `getClosestStep`: Given a position and a route, determines the closest step in the route.

It also has a basic Activity to showcase its functionality. It currently only outputs information via `Log`, we should think how to better show this in the UI.

#### Next

- [x] Improve TestApp's Activity UI
- [x] Include tests

/cc: @cammace

### DIFF
--- a/libandroid/app/src/main/AndroidManifest.xml
+++ b/libandroid/app/src/main/AndroidManifest.xml
@@ -36,6 +36,11 @@
             android:parentActivityName=".MainActivity"
             android:theme="@style/AppTheme.NoActionBar" />
         <activity
+            android:name=".directions.RouteUtilsV5Activity"
+            android:label="Route Utils v5"
+            android:parentActivityName=".MainActivity"
+            android:theme="@style/AppTheme.NoActionBar" />
+        <activity
             android:name=".icons.DirectionsIconsActivity"
             android:label="Directions icons"
             android:parentActivityName=".MainActivity"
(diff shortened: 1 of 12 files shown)


zugaldia:
The `RouteUtilsV5Activity` now has a better UI:

video

1. It shows each step of the route in alternating colors to distinguish them better.
2. Fix a bug with 1-element steps.
3. Draws the lines to the closest point on the step.

To use it: tap once to set the `from`, tap a second time to set the `to` and compute the route. Every following tap will show information about the route in regard with the "user location" (the place where you tap the map).

Next are tests.

zugaldia:
Merging.

PR #200, opened 2016-09-20
updated method count badge

cammace:
I've switched to using mapbox-java instead to better represent this library.

cc: @zugaldia


cammace:
Old badge was only displaying libandroid method count (which should be around 124) i've switched to libjava to more accurately reflect this SDK.

zugaldia:
@cammace Makes sense, thanks. I see OkHttp and Gson being the main two contributors to that count. I believe both are a reasonable use and assumption nowadays.

PR #204, opened 2016-09-27
Improve User-Agent string

### LINKED ISSUE #92: Improve user agent string
Once https://github.com/mapbox/mapbox-java/pull/91 landed, we started setting a string that looks like this: `MapboxJava/1.2.0 Mac OS X/10.11.5 (x86_64)`

We'd also like to include:
* The name of the Android app (if any)
* Automate the inclusion of the version number in the Java code to avoid inconsistencies.

/cc: @1ec5 @cammace

ivovandongen:
Fixes #92

### DIFF
--- a/libjava/.gitignore
+++ b/libjava/.gitignore
@@ -5,3 +5,4 @@
 .DS_Store
 /build
 /captures
+BuildConfig.java
--- a/libjava/lib/build-config.gradle
+++ b/libjava/lib/build-config.gradle
@@ -1,0 +1,101 @@
+import org.gradle.api.DefaultTask
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.Task
+import org.gradle.api.tasks.Input
+import org.gradle.api.tasks.OutputFile
+import org.gradle.api.tasks.TaskAction
+
+// Plugin to generate a BuildConfig class from Gradle //
+
+/**
+ * Main plugin class
+ */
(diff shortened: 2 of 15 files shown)


zugaldia:
Looking good though CI is complaining:

> A problem occurred evaluating project ':libjava'.
> Could not get unknown property 'BuildConfigPlugin' for project ':libjava' of type org.gradle.api.Project.

ivovandongen:
>Looking good though CI is complaining

Yeah, having trouble including the project in such a way that both projects can use it (apparently needed when depending on another project directly). Also, generating the `BuildConfig` in any other directory than the main sources doesn't work in Android Studio, but does cmd-line.

For now, I'm going to generate the `BuildConfig` in the main sources with a git ignore entry. This appeases AS, bitrise and cmd line invocation.

ivovandongen:
@zugaldia Added a setter on the builders to add an App name optionally. This results in 2 possible user agents:

- No app name set: `MapboxJava/2.0.0-SNAPSHOT (975537e) Mac OS X/10.11.6 (x86_64)`
- With app name set: `AppName MapboxJava/2.0.0-SNAPSHOT (975537e) Mac OS X/10.11.6 (x86_64)`

ivovandongen:
> Javadoc ain't happy

Should be fxed

zugaldia:
@ivovandongen 👍 Ready to merge then.

PR #209, opened 2016-10-30
add poi.landmark type to geocoder

sbma44:
dropped in `poi.landmark`. Hoping to see what CI tests say, then take it from there. Eyes appreciated, I'm totally unfamiliar with Java development.


sbma44:
Thanks @ivovandongen! Anything I need to do in connection to the version bump when I hit merge?

cammace:
Versioning's fine, we can always bump later down the road if needed. Thanks!

PR #214, opened 2016-11-08
Added direction v5 lanes and updated fixture

cammace:
This adds lanes to the direction v5 models along with updating the fixtures to run test on this. Can you review please @ivovandongen.

cc: @zugaldia


cammace:
Thanks for pointing out the Makefile @zugaldia, I updated the fixture using the same API call and reset one of the test geometries back to what it was on master.

PR #217, opened 2016-11-23
Updates Geocoding to match API

cammace:
We are missing a few options/parameters the geocoder gives you. This PR adds `limit` and exposes `country` in the auto complete widget. Still need to add `landmark`.

cc: @zugaldia

### DIFF
--- a/libandroid/lib/src/main/java/com/mapbox/services/android/geocoder/ui/GeocoderAutoCompleteView.java
+++ b/libandroid/lib/src/main/java/com/mapbox/services/android/geocoder/ui/GeocoderAutoCompleteView.java
@@ -108,16 +108,27 @@ public class GeocoderAutoCompleteView extends AutoCompleteTextView {
   }
 
   /**
-   * parameter limits results to a set of one or more countries
-   * <a href="https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2">ISO 3166 alpha 2</a> country codes,
-   * separated by commas.
+   * Parameter limits results to a country. Use one of the
+   * <a href="https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2">ISO 3166 alpha 2</a> country codes. The country code is
+   * case sensitive and needs to be all lowercase.
    *
-   * @param country <a href="https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2">ISO 3166 alpha 2</a>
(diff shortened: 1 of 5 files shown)


cammace:
> ~Still need to add landmark.~

This is a property and can be acquired by the user using `feature.getProperties().get("landmark").getAsBoolean()`.

cammace:
Could I get someone on @mapbox/android for review :)

libandroid/lib/src/main/java/com/mapbox/services/android/geocoder/ui/GeocoderAutoCompleteView.java
```diff
@@ -108,6 +108,19 @@ public void setType(String type) {
...
   /**
+   * parameter limits results to a set of one or more countries
+   * <a href="https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2">ISO 3166 alpha 2</a> country codes,
+   * separated by commas.
+   *
+   * @param country <a href="https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2">ISO 3166 alpha 2</a>
+   *                country codes, separated by commas.
+   * @since 2.0.0
+   */
+  public void setCountry(String country) {
```
ivovandongen: We might want to offer a varargs method here for convenience so that the comma separated format is abstracted:

```
public void setCountries(String... countries)
```

sbma44:
> This is a property and can be acquired by the user using 

`landmark` is both a property on the response and (part of) a valid value for the `types` param. e.g. `types=place,poi.landmark`

libandroid/lib/src/main/java/com/mapbox/services/android/geocoder/ui/GeocoderAutoCompleteView.java
cammace: Great point @ivovandongen, adding.

cammace:
@ivovandongen: I've added a countries option which takes in a String array, array was used here since it was already implemented that way in the [`libjava` countries option](https://github.com/mapbox/mapbox-java/blob/master/libjava/lib/src/main/java/com/mapbox/services/geocoding/v5/MapboxGeocoding.java#L256-L258). 

@sbma44 Happy to help out on the geocoding Android side of things. For types, we allow `poi.landmark` to be passed in to filter the results.

PR #222, opened 2016-11-29
Turf Point On Line fixed and test added

cammace:
We were missing test for  [`Turf.Misc#pointOnLine()`](https://github.com/mapbox/mapbox-java/blob/master/libjava/lib/src/main/java/com/mapbox/services/commons/turf/TurfMisc.java#L90).

cc: @ivovandongen @zugaldia


cammace:
Resolved a few errors made when originally porting `point-on-line`, all test are passing now and using a demo activity, it appears to be working correctly now. 

cc: @ivovandongen can you review this 😃

tobrun:
Tests are looking good!

PR #226, opened 2016-11-30
Added cancel method to autocomplete widget

### LINKED ISSUE #206: Expose API to cancel retrofit call for the geocoding widget
We need to expose a way to cancel the geocoder widget API call and add in documentation to implement it in the `onDestroy` method of activity so when the device is rotated there isn't a memory leak.

cc: @zugaldia

cammace:
Closes #206

cc: @ivovandongen for review

### DIFF
--- a/libandroid/lib/src/main/java/com/mapbox/services/android/geocoder/ui/GeocoderAdapter.java
+++ b/libandroid/lib/src/main/java/com/mapbox/services/android/geocoder/ui/GeocoderAdapter.java
@@ -31,8 +31,6 @@ import retrofit2.Response;
  */
 public class GeocoderAdapter extends BaseAdapter implements Filterable {
 
-  private static final String TAG = "GeocoderAdapter";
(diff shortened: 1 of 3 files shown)


libandroid/lib/src/main/java/com/mapbox/services/android/geocoder/ui/GeocoderAdapter.java
```diff
@@ -225,6 +230,20 @@ public void setLimit(int limit) {
...
   }
 
+  /**
+   * Can be used to cancel any calls currently in progress. It's a good idea to include in onDestroy() to prevent
+   * memory leaks
+   *
+   * @since 2.0.0
+   */
+  public void cancelApiCall() {
+    if (call == null) {
```
ivovandongen: Don't think we need to log this as it's not an uncommon occurrence. At most, log at debug I think.

PR #237, opened 2016-12-06
fixed javadoc errors

### LINKED ISSUE #121: Javadoc errors
When building the javadoc today for `1.2.1` release, errors were found. Many of the packages are missing info as well. i'd like to go in and fix all these issues and make sure everything is documented well. 

On a side note, I think we should try using the `@since` tag to make it easier for users to know when things have been added to the SDK. 

cc: @zugaldia

cammace:
Closes #121 

@tobrun for review

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
 javadoc:
 	# Output is ./libjava/lib/build/docs/javadoc
-	cd libjava; ./gradlew clean javadoc
+	cd libjava; ./gradlew javadocGeneration
 
 	# Output is ./libandroid/lib/build/docs/javadoc/release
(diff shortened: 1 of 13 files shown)


cammace:
One thing to point out is that you have to switch from `compile project(':libjava')` in libandroid lib's build.gradle file to use the maven dependency:

```groovy
// Mapbox Java Services (release)
    compile ('com.mapbox.mapboxsdk:mapbox-java-services:2.0.0-SNAPSHOT@jar') {
        transitive=true
    }
```

Otherwise when generating the javadoc for libandroid it won't find libjava as a dependency and a bunch of errors will occur.

PR #259, opened 2017-01-06
Round position to 7 digits

cammace:
For part of the compliance with RFC 7946 for GeoJSON, we shouldn't write longitude and latitude values with more than 7 decimal places of precision. This can be applied to our Position object since at a 7 decimal precision, it gives a 11mm accuracy.


zugaldia:
I'm not sure about this approach. `Position` can be used beyond GeoJSON serialization. Why don't we apply this formatting only to the GeoJSON classes instead? We'll probably need a custom Gson rule.

mapbox/libjava-geojson/src/main/java/com/mapbox/services/commons/geojson/custom/PositionSerializer.java
```diff
@@ -37,9 +38,10 @@ public JsonElement serialize(Position src, Type typeOfSrc, JsonSerializationCont
     JsonArray rawCoordinates = new JsonArray();
 
     BigDecimal lat = new BigDecimal(src.getLatitude());
-    lat = lat.round(new MathContext(7));
+    lat = lat.setScale(7, RoundingMode.HALF_UP).stripTrailingZeros();
```
zugaldia: @cammace Doesn't this need a call to `asDouble()` to make sure we write a `double` value to the `JsonArray`? Does it change the behavior in any way?

cammace:
I was able to figure out the difference between BigDecimal round and scale. Rounding will round considering all significant figures in the number. So say I had `105.01234567`, rounding would make the digit `105.0123` which is not what we want. Scale only effects the number of digits to the right of the decimal point which _is_ what we want. using the same example above would become `105.0123457` due to rounding. I also added `stripTrailingZeros` to remove any trailing zeros to the rounded number. 

In addition, this PR moves all the geojson test to the correct package and adds an additional test file called `GeojsonTest` which we should use for all compliance test to the spec.

Issue #278, opened 2017-01-17
Make Carmen feature easier to save in Realm database


cammace:
Currently, Realm offers saving of [mostly primitive types](https://realm.io/docs/java/latest/#field-types) therefore it isn't possible to directly save a Carmen Feature. Saving Carmen Feature is ideal when you want to store recent queries. Thoughts @zugaldia?

zugaldia:
@cammace You can always serialize/deserialize any of our API models to a `String` (with Gson)?

PR #279, opened 2017-01-17
Added fromjson test and method for carmen feature

cammace:
Closes: https://github.com/mapbox/mapbox-java/issues/278

Carmen Feature was using the `fromJson()` method found in the Feature.class. This method was returning a Feature object and not a Carmen. To get around this, I added a fromJson method to Carmen Feature.

### DIFF
--- a/mapbox/libjava-services/src/main/java/com/mapbox/services/api/geocoding/v5/models/CarmenFeature.java
+++ b/mapbox/libjava-services/src/main/java/com/mapbox/services/api/geocoding/v5/models/CarmenFeature.java
@@ -142,6 +142,9 @@ public class CarmenFeature extends Feature {
    */
   public static CarmenFeature fromJson(String json) {
     GsonBuilder gson = new GsonBuilder();
+    gson.registerTypeAdapter(Position.class, new PositionDeserializer());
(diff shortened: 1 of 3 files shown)


mapbox/libjava-services/src/main/java/com/mapbox/services/api/geocoding/v5/models/CarmenFeature.java
```diff
@@ -129,6 +134,18 @@ public void setRelevance(double relevance) {
...
   /**
+   * Create a CarmenFeature object from JSON.
+   *
+   * @param json String of JSON making up a carmen feature.
+   * @return Carmen Feature
+   * @since 2.0.0
+   */
+  public static CarmenFeature fromJson(String json) {
+    GsonBuilder gson = new GsonBuilder();
+    return gson.create().fromJson(json, CarmenFeature.class);
```
zugaldia: @cammace Don't you need to attach our custom deserializer here `gson.registerTypeAdapter(Position.class, new PositionDeserializer());`?

cammace: Good point, need to make my test more robust to catch this also.

PR #285, opened 2017-01-24
bbox and coordEach added to MAS

### LINKED ISSUE #196: Bring Turf.bbox to libjava
Original documentation is [here](http://turfjs.org/docs/#bbox) and code is [here](https://github.com/Turfjs/turf/blob/master/packages/turf-bbox/index.js).

This can be used, for example, to create a `LatLngBounds` object as input to `CameraUpdate` to fit an object in the screen (e.g. a route).

cammace:
Closes #196 

All test are passing, I modified coordEach to simply return a list of positions instead of an iterable list. Difference between coordEach and [CoordAll](http://turfjs.org/docs#coordall) would be the boolean `excludeWrapCoord` parameter which can be used to remove the last point in a polygon or multipolygon (since they are the same point as the first in geometry). 

cc: @zugaldia


mapbox/libjava-services/src/main/java/com/mapbox/services/api/utils/turf/TurfMeta.java
```diff
@@ -0,0 +1,108 @@
...
+      for (g = 0; g < stopG; g++) {
+
+        geometry = isGeometryCollection
+          ? (((GeometryCollection) geometryMaybeCollection).getGeometries().get(g))
+          : (Geometry) geometryMaybeCollection;
+
+        int wrapShrink = (excludeWrapCoord
+          && (geometry.getType().equals("Polygon") || geometry.getType().equals("MultiPolygon")) ? 1 : 0);
+
+        if (geometry.getType().equals("Point")) {
```
zugaldia: I'm not completely satisfied with this approach as it goes around Java's type system. How about instead of having one only `TurfMeta.coordEach(GeoJSON layer, boolean excludeWrapCoord)` that takes a generic `GeoJSON` object, we have different `TurfMeta.coordEach()` methods for each supported GeoJSON type? Something like [what we do here](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libjava-services/src/main/java/com/mapbox/services/api/utils/turf/TurfMeasurement.java) with `destination()`.

cammace: see [lineDistance](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libjava-services/src/main/java/com/mapbox/services/api/utils/turf/TurfMeasurement.java#L213) 😄 

Will look into splitting it up into different methods.

zugaldia: @cammace Damn, that's not right either.

cammace:
Closing in favor of #366

PR #228, opened 2016-12-01
[WIP] Work on route utils

cammace:
Closes https://github.com/mapbox/mapbox-java/issues/207

PR #287, opened 2017-01-24
Adds support for polyline6

### LINKED ISSUE #207: Change RouteUtils "route" parameter name
Most of the [RouteUtils](http://mapbox.github.io/mapbox-java/api/libjava/1.3.1/com/mapbox/services/navigation/v5/RouteUtils.html) methods take in a RouteLeg parameter named `route`. The parameter would be more appropriately named `leg` or `routeLeg`.

### LINKED ISSUE #288: Supports Directions precision-6
Need to add geometry option precision-6 and allow decoding of precision-6 strings.

cc: @zugaldia

cammace:
Closes #207 
Closes #288

Bring over the work done in #228 and completes it.

### DIFF
--- a/mapbox/libjava-core/src/main/java/com/mapbox/services/Constants.java
+++ b/mapbox/libjava-core/src/main/java/com/mapbox/services/Constants.java
@@ -28,7 +28,6 @@ public class Constants {
   public static final int GOOGLE_PRECISION = 5;
   public static final int OSRM_PRECISION_V4 = 6;
   public static final int OSRM_PRECISION_V5 = 5;
-  public static final int OSRM_PRECISION_6_V5 = 6;
 
   // Use these if you want to be explicit about the precision
   public static final int PRECISION_5 = 5;
--- a/mapbox/libjava-core/src/main/java/com/mapbox/services/Constants.java
+++ b/mapbox/libjava-core/src/main/java/com/mapbox/services/Constants.java
@@ -28,6 +28,7 @@ public class Constants {
   public static final int GOOGLE_PRECISION = 5;
   public static final int OSRM_PRECISION_V4 = 6;
   public static final int OSRM_PRECISION_V5 = 5;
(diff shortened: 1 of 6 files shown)


zugaldia:
@cammace Should we close https://github.com/mapbox/mapbox-java/pull/228 then?

cammace:
I've fixed up this PR to only add support for precision 6 in the directions api. The route utils refactor will come laster down the road but I don't expect it to break the API.

mapbox/libjava-core/src/main/java/com/mapbox/services/Constants.java
```diff
@@ -26,6 +26,7 @@
   public static final int GOOGLE_PRECISION = 5;
   public static final int OSRM_PRECISION_V4 = 6;
   public static final int OSRM_PRECISION_V5 = 5;
+  public static final int OSRM_PRECISION_6_V5 = 6;
```
zugaldia: @cammace Let's not do this, let's use `PRECISION_6 ` from https://github.com/mapbox/mapbox-java/pull/312 instead.

mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
```diff
@@ -148,12 +148,24 @@ public Builder() {
       // Set defaults
       this.user = DirectionsCriteria.PROFILE_DEFAULT_USER;
 
-      // We only support polyline encoded geometries to reduce the size of the response.
-      // If we need the corresponding LineString object, this SDK can do the decoding with
-      // LineString.fromPolyline(String polyline, int precision).
+      // by defauly the geometry is polyline with precision 5.
       this.geometries = DirectionsCriteria.GEOMETRY_POLYLINE;
```
zugaldia: @cammace Let's default to precision 6. That's what I do on MapMatching and the payload isn't much bigger but there's a gain in precision.

PR #292, opened 2017-01-27
mapbox/driving-traffic support added to the list of directions profil…

srabenja:
…es, and checks added for maximum number of coordinates that can be provided, as discussed with @zugaldia


mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
```diff
@@ -464,6 +465,18 @@ public MapboxDirections build() throws ServicesException {
           "You should provide at least two coordinates (from/to).");
       }
 
+
+      if (profile != null && profile.equals(DirectionsCriteria.PROFILE_DRIVING_TRAFFIC)
```
zugaldia: @srabenja Considering that `profile` and `coordinates` are required parameters, how about we add a check before these two raise an exception if either is `null` (or empty)?

srabenja:
@zugaldia I added the changes you mentioned

zugaldia:
@srabenja I see the fixture, but I see no tests using it?

cammace:
I have fixed a few documentation things, cleaned up the fixture, rewrote some exception messages to make it more clear what error occurred and lastly, added a few test. @zugaldia can you review this again?

Issue #296, opened 2017-01-27
Add convenience method for roundabout exit


cammace:
If the maneuver type is a roundabout, an exit is provided in the json properties. 

cc: @zugaldia

PR #303, opened 2017-02-03
added exit convenience method

ghoshkaj:
#296


mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/models/StepManeuver.java
```diff
@@ -76,6 +77,23 @@ public String getType() {
...
+   * <br>
+   * <ul>
+   * <li>else - indicates the number of intersections passed until the turn.
+   * Example instruction: "at the fourth intersection, turn left"
+   * <li>roundabout - traverse roundabout</li>
+   * <li>rotary - a traffic circle</li>
+   * </ul>
+   *
+   * @return String .
+   * @since 1.0.0
```
cammace: This should be 2.0.0 since it is just now getting introduced :)

Issue #283, opened 2017-01-23
Rename Position.fromCoordinate() to fromLngLat()


1ec5:
`Position.fromCoordinate()` should be deprecated in favor of a `fromLngLat()` method that takes the same parameters.

`Position.fromCoordinate()` accepts parameters in longitude, latitude order, consistent with GeoJSON’s `Position` object. Unfortunately, the Mapbox Android SDK’s `LatLng` object’s constructor accepts parameters in latitude, longitude order. As a consequence, map camera changes, annotations, anything involving user location, and offline maps all use latitude, longitude order while runtime styling and feature querying both use longitude, latitude order.

It’s quite easy to imagine a scenario where one order is required on one line and the reverse order is required on the following line. But without labeled parameters, it’s too easy for copy-pasta to occur between the two without any way for a code reviewer to spot the mistake. At least if the code reviewer sees `LatLng` on one line and `LngLat` on the next line, they’ll know to expect reversed parameters.

/cc @zugaldia @cammace

Issue #266, opened 2017-01-10
Add Position check for correct order


cammace:
When creating a Position object, many users accidentally switch up the latitude and longitude values. I propose adding a check so that if the latitude is over `90.0` we throw an error. While this wont always catch the error, it might help out users. We can also make the `invalid position` error more clear on this issue.

zugaldia:
I like it. I wouldn't throw an error but a warning, as the user might be handling coordinates that need to be wrapped.

Related: https://github.com/mapbox/mapbox-java/issues/283.

PR #309, opened 2017-02-07
Conveniences to Position model

zugaldia:
This PR does two things:

- Adds a new `Position.fromLngLat(double longitude, double latitude)` method the be more explicit about the lat/lon order. It does not deprecate the `Position.fromCoordinates()` method as this is used for other cases too (e.g. set an altitude or a double array). Fixes https://github.com/mapbox/mapbox-java/issues/283.
- Adds a warning to the `Position` constructor if the lat/lon values seem out of range. Fixes https://github.com/mapbox/mapbox-java/issues/266.

This last item is tested via the Test App. In `MainActivity` we create an object with values out of range and the warning can be seen as usual in the Android log window:

image

@cammace Please review.


1ec5:
Position is still used with the Android SDK's runtime styling API, right? Note that out-of-range longitudes are expected for features that straddle the antimeridian.

zugaldia:
@1ec5 Right, we're tracking a couple of [antimeridian-related issues here](https://github.com/mapbox/mapbox-java/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20antimeridian).

1ec5:
Yes, but doesn’t antimeridian straddling already work in the Android SDK? mapbox/mapbox-gl-native#6088 Or does that affect only annotations, which would use the SDK-specific model objects?

Issue #281, opened 2017-01-18
Add optional language parameter to geocoder


cammace:
iOS allows the option to [pass in a language](https://github.com/mapbox/MapboxGeocoder.swift/blob/master/MapboxGeocoder/MBGeocodeOptions.swift#L101) parameter. My understanding of it is that it isn't part of the geocoder API but rather something done inside the library itself. @1ec5 can you elaborate a bit more on the behavior on this and whether we should implement it in MAS?

cc: @zugaldia

1ec5:
Here's [the documentation for this option](https://github.com/mapbox/MapboxGeocoder.swift/blob/af6119f6a253b2ab60c69e4970f64dfbc3a390a5/MapboxGeocoder/MBGeocodeOptions.swift#L56-L65).

MapboxGeocoder.swift lets the developer specify an NSLocale (a standard Apple-platform object representing a locale); it extracts the language code from that NSLocale and passes it into the API. Currently, the default leaves the locale unspecified, because the option is experimental, but in internal applications we've been passing in the system locale.

1ec5:
The Geocoding team is fine with us exposing this option, as long as we mark it as experimental and clarify its purpose (just the display language).

PR #315, opened 2017-02-08
Add optional language parameter to geocoder

zugaldia:
Fixes https://github.com/mapbox/mapbox-java/issues/281.


cammace:
We should also expose this in our geocodingAutoCompleteWidget correct?

zugaldia:
@cammace good call. just added it via https://github.com/mapbox/mapbox-java/pull/315/commits/740f850d85f233c4f04ea7ccbc103d6cc6f0a15e. Care to review again?

Issue #202, opened 2016-09-20
Fix CI for javadoc PRs


cammace:
When opening a PR to merge into `gh-pages` CI fails (possibly not running the correct build script).

cammace:
I've added a `bitrise.yml` file to the `gh-pages` branch and edited the triggers which seems to have fixed the issue. Closing

PR #326, opened 2017-02-12
2.0 beta1 javadoc

cammace:
Left the old module Javadoc modules and just labeled them as `old` and the new module javadocs have been added.


zugaldia:
Looks good though we're still hitting https://github.com/mapbox/mapbox-java/issues/202.

Issue #293, opened 2017-01-27
Add convenience methods for rotary properties


cammace:
`rotary_name` and `rotary_pronunciation` are potentially given to use with the maneuver type is a rotary. 

cc: @zugaldia

cammace:
Resolved in #329 correct?

ghoshkaj:
yes ^

PR #329, opened 2017-02-15
added rotary convenience methods

ghoshkaj:
#293


mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/models/LegStep.java
```diff
@@ -78,4 +84,36 @@ public StepManeuver getManeuver() {
...
+    return rotaryPronunciation;
+  }
+
+  /**
+   * The pronunciation hint of the way name. Will be undefined if no pronunciation is hit.
+   *
+   * @return String with the pronunciation
+   * @since 2.0.0
+   */
+  public String getPronunciation() {
```
zugaldia: @ghoshkaj why two getters for the same field `getRotaryPronunciation` and `getPronunciation`?

ghoshkaj: because I messed up :D fixed it!

PR #337, opened 2017-02-20
Added bearing query parameter 

cammace:
We were missing this parameter in our builder.


mapbox/app/src/main/java/com/mapbox/services/android/testapp/directions/DirectionsV5Activity.java
```diff
@@ -135,6 +136,8 @@ public void call(DirectionsResponse response) {
     client.enqueueCall(new Callback<DirectionsResponse>() {
       @Override
       public void onResponse(Call<DirectionsResponse> call, Response<DirectionsResponse> response) {
+        System.out.println(call.request().url().toString());
```
zugaldia: `Log.d` is better here.

PR #352, opened 2017-02-22
Fix versioning

zugaldia:
This PR does two things:

1. Centralize all deps in `dependencies.gradle`
2. Bump a few stable versions


zugaldia:
As a nice side effect, upgrading to the latest Retrofit makes us support RxJava2. cc: @Guardiola31337 .

PR #356, opened 2017-02-23
AppEngine support (and II)

zugaldia:
Turns out that the approach in https://github.com/mapbox/mapbox-java/pull/344 wouldn't work if both the client and the factory were set at the same time. This PR makes sure we only set one.

It also adds a number of no-args constructors to make Gson happy.


zugaldia:
A sample `AppengineFactory` can be found [here](https://gist.github.com/zugaldia/e4ba2e758678b8702824c6a5e38eb925).

PR #373, opened 2017-03-02
make service and turf RuntimeExceptions

### LINKED ISSUE #66: libjava - make ServiceException an unchecked exception
As the current validations that throw the Service Exception are mostly programming errors, it makes sense to make ServiceException an unchecked exception to avoid unneeded boilerplate in the consuming code.

One exception (no pun intended to this) is the the image size validation as these could potentially be more data driven. Although not checking for 0/negative integers on the caller side might as well be considered a programming error.

cammace:
Closes #66


zugaldia:
@cammace could you take care of the checkstyle issues? otherwise it all looks 👍 .

Issue #368, opened 2017-02-28
Add setters for Directions API


cammace:
Currently we offer an empty constructor but don't expose setters for the user to build off of.

zugaldia:
Moving it out of 2.0 as this won't affect SEMVER.

cammace:
setters for directions have been added. closing.

PR #374, opened 2017-03-02
Port of `osrm-text-instructions` to Java

zugaldia:
For easier iterations, let's include it in the navigation package. Once it's stable, let's spin it off as its own package / repo possibly like https://github.com/Project-OSRM/osrm-text-instructions.swift.

cc: @freenerd @frederoni @bsudekum @1ec5  @cammace


Makefile
```diff
@@ -108,3 +108,8 @@ distance-fixtures:
 	curl -X POST --header "Content-Type:application/json" -d @mapbox/libjava-services/src/test/fixtures/distance_coordinates.json \
 		"https://api.mapbox.com/distances/v1/mapbox/driving?access_token=$(MAPBOX_ACCESS_TOKEN)" \
 		-o mapbox/libjava-services/src/test/fixtures/distance_v1.json
+
+libosrm:
+	rm -rf mapbox/libjava-services/src/main/resources/translations
+	mkdir -p mapbox/libjava-services/src/main/resources/translations
+	cp ../osrm-text-instructions/languages/translations/* mapbox/libjava-services/src/main/resources/translations
```
1ec5: This assumes the developer has a clone of osrm-text-instructions at a specific location on disk, and people with different commits checked out will end up with different JSON files in this repository. Can we make this dependency less fragile, perhaps using a submodule?

zugaldia:
The good news is that the core logic as well as tests have now been ported. The bad news is that the tests for the `compile` method are not passing. We're getting different values for the compiled instructions vs fixtures:

image

Makefile
zugaldia: @1ec5 totally, though I personally would love to avoid submodules at all costs, the final set up is gonna depend on how we decide to distribute and host this library. cc: @freenerd

1ec5: If you check in the JSON files, then the submodule can be something only contributors to this project have to worry about; someone incorporating the library would use the checked-in JSON files but never have to worry about how they got there.

1ec5: The current code’s disadvantage is twofold: it assumes a suprarepository file structure, and it doesn’t record versions or commit hashes, so it’ll be more difficult to bisect should something go wrong that depends on specific translations.

cammace:
It looks like you've added additional constructors to some of the direction model classes. Would this resolve https://github.com/mapbox/mapbox-java/issues/368 or do we still want to add setters?

zugaldia:
@cammace I'm afraid this PR won't bring all the getters/setter needed for https://github.com/mapbox/mapbox-java/issues/368, we'll need to address that one in a separate PR. Otherwise, after the fixes we're now good for your review.

PR #379, opened 2017-03-07
[WIP] Add sample activity implementing Navigation SDK

zugaldia:
This PR brings a new sample activity to the test app that shows how to use the new navigation SDK on Android.

cc: @cammace


cammace:
Note that the MapboxTelemetry object needs to be reset to master before merging.

PR #383, opened 2017-03-08
Fix set origin and destination

### LINKED ISSUE #378: Resetting origin or destination in same directions builder, adds to list instead of replacing
Described in the title, if you setOrigin twice, it will only add to the list instead of replacing the origin found in the list.

@zugaldia

Guardiola31337:
Fixes #378 

Fix `setOrigin` and `setDestination` replacing beginning and end positions instead of adding them.

👀  @cammace @zugaldia

### DIFF
--- a/mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
+++ b/mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
@@ -133,7 +133,7 @@ public class MapboxDirections extends MapboxService<DirectionsResponse> {
 
     private static final int TWO_COORDINATES = 2;
     private static final int END = 1;
-    public static final int BEGINNING = 0;
+    private static final int BEGINNING = 0;
     // We use `Boolean` instead of `boolean` to allow unset (null) values.
     private String user = null;
     private String profile = null;
--- a/mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
+++ b/mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
@@ -3,9 +3,9 @@ package com.mapbox.services.api.directions.v5;
 import com.mapbox.services.api.MapboxBuilder;
 import com.mapbox.services.api.MapboxService;
 import com.mapbox.services.api.ServicesException;
+import com.mapbox.services.api.directions.v5.models.DirectionsResponse;
 import com.mapbox.services.commons.models.Position;
 import com.mapbox.services.commons.utils.TextUtils;
-import com.mapbox.services.api.directions.v5.models.DirectionsResponse;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -131,6 +131,9 @@ public class MapboxDirections extends MapboxService<DirectionsResponse> {
    */
   public static class Builder<T extends Builder> extends MapboxBuilder {
 
+    private static final int TWO_COORDINATES = 2;
+    private static final int END = 1;
+    public static final int BEGINNING = 0;
     // We use `Boolean` instead of `boolean` to allow unset (null) values.
     private String user = null;
     private String profile = null;
@@ -223,11 +226,7 @@ public class MapboxDirections extends MapboxService<DirectionsResponse> {
         coordinates = new ArrayList<>();
       }
 
-      // The default behavior of ArrayList is to inserts the specified element at the
-      // specified position in this list (beginning) and to shift the element currently at
-      // that position (if any) and any subsequent elements to the right (adds one to
-      // their indices)
-      coordinates.add(0, origin);
+      insertOrigin(origin);
 
       return (T) this;
     }
@@ -246,9 +245,7 @@ public class MapboxDirections extends MapboxService<DirectionsResponse> {
         coordinates = new ArrayList<>();
       }
 
-      // The default behavior for ArrayList is to appends the specified element
-      // to the end of this list.
-      coordinates.add(destination);
+      insertDestination(destination);
 
       return (T) this;
     }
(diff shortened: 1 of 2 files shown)


mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
```diff
@@ -131,6 +131,9 @@ public void cancelCall() {
    */
   public static class Builder<T extends Builder> extends MapboxBuilder {
 
+    private static final int TWO_COORDINATES = 2;
```
cammace: Constants can go in the Constants.java file

mapbox/libjava-services/src/test/java/com/mapbox/services/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -168,27 +171,30 @@ public void noProfileProvided() throws ServicesException {
...
       .setProfile(DirectionsCriteria.PROFILE_DRIVING)
       .setBaseUrl(mockUrl.toString())
       .build();
+
     Response<DirectionsResponse> response = client.executeCall();
-    assertEquals(response.code(), 200);
-    assertEquals(response.body().getCode(), DirectionsCriteria.RESPONSE_OK);
+
+    assertEquals(200, response.code());
+    assertEquals("Ok", response.body().getCode());
```
zugaldia: @Guardiola31337 why replacing the constant with its value?

mapbox/libjava-services/src/test/java/com/mapbox/services/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -290,300 +296,317 @@ public void testBearingArrayLengthNotExactlyTwo() throws ServicesException {
...
       .build();
-    Response<DirectionsResponse> response = client.executeCall();
 
+    Response<DirectionsResponse> response = client.executeCall();
     DirectionsResponse body = response.body();
-    assertEquals(body.getCode(), DirectionsCriteria.RESPONSE_OK);
-    assertEquals(body.getRoutes().size(), 1);
-    assertEquals(body.getWaypoints().size(), 2);
+
+    assertEquals("Ok", body.getCode());
```
zugaldia: @Guardiola31337 same comnent as above.

mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
Guardiola31337: @cammace I don't like the approach of having this kind of _all-constant_ classes. Constants should be in their pertinent classes, basically because it's where they belong and provide more context. Actually, [using an _all-constant_ class approach is a well known antipattern](https://softwareengineering.stackexchange.com/questions/49572/is-it-a-bad-practice-to-have-an-interface-to-define-constants#answer-49579).

mapbox/libjava-services/src/test/java/com/mapbox/services/api/directions/v5/MapboxDirectionsTest.java
Guardiola31337: @zugaldia it's a technique what I learnt from the book [Working Effectively With Unit Tests](https://leanpub.com/wewut). I really like the techniques described on it in favor of DAMP (Descriptive And Maintainable Procedures). Mainly because they increase maintainability and readability of tests.
In this concrete case, why I prefer _expecting literals_?

- The main advantage of expecting literals is that you only need to focus on the actual value of the assertion, increasing readability and traceability.
If neither your expected nor actual values are literals you’re forced to determine the value of both when a test fails. Conversely, if your expected value is a literal, only the actual value will need to be traced when a test is failing.
- Literals will reflect the examples from the business.

zugaldia: @Guardiola31337 Thanks for the explanation. While agree with these general goals, I disagree in this particular case. This test really intends to test `response.body().getCode().equals(DirectionsCriteria.RESPONSE_OK)` which is what the developer would type in their code. The fact that `DirectionsCriteria.RESPONSE_OK` has the value `Ok` or `OK` or something else is secondary in this case. In fact, if the server happened to change its response from `Ok` to `OK` and we simply updated the literal here, the test would now pass, but we'd miss updating the value of `DirectionsCriteria.RESPONSE_OK`.

mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
```diff
@@ -131,6 +131,9 @@ public void cancelCall() {
...
+    private static final int END = 1;
+    public static final int BEGINNING = 0;
```
zugaldia: @Guardiola31337 why `public`?

mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
```diff
@@ -578,5 +575,25 @@ public MapboxDirections build() throws ServicesException {
       }
       return new MapboxDirections(this);
     }
+
```
zugaldia: On a second thought, I think we should fix this in a simpler way.
- `setOrigin(Position origin)` should set a private `Position origin` value.
- `setDestination(Position destination)` should set a private `Position destination` value.
- `setCoordinates(List<Position> coordinates)` should set a private `List<Position> coordinates` value.

Then, at build time we check if both (`origin`, `destination`) and  `coordinates` have been set. If the developer has mixed them we through a `ServicesException`.

mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
Guardiola31337: You're right 😅

mapbox/libjava-services/src/test/java/com/mapbox/services/api/directions/v5/MapboxDirectionsTest.java
Guardiola31337: @zugaldia Ok agreed. It's also explanatory using the constant so I leave it as it was 👍

mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/MapboxDirections.java
Guardiola31337: @zugaldia Since I don't know in depth the SDK, does it mean that `coordinates` would be a list of coordinates in between the `origin` and the `destination`? At the moment, `origin` is the first element of the list and `destination` is the last one. When you try to add a new `origin`, if you already had more than two `coordinates`, the new `origin` is prepended to the `coordinates` list. Similarly, when you try to add a new `destination`, if you already had more than two `coordinates`, the new `destination` is appended. But, if you called `setCoordinates` the whole `coordinates` is reset. Is that the behavior that we are looking for?
cc/ @cammace

cammace:
closing this in favor of https://github.com/mapbox/mapbox-java/pull/420.

PR #402, opened 2017-03-21
added setters to location engine and javadoc

### LINKED ISSUE #395: Add setters for location engine interval and fastest interval.
We only expose a setter for priority meanwhile interval and fastest interval are also quite important and should be allowed to set depending on the developers application.

cc: @zugaldia

cammace:
Closes #395

### DIFF
--- a/mapbox/libandroid-services/src/main/java/com/mapbox/services/android/location/LostLocationEngine.java
+++ b/mapbox/libandroid-services/src/main/java/com/mapbox/services/android/location/LostLocationEngine.java
@@ -89,9 +89,9 @@ public class LostLocationEngine extends LocationEngine implements
   public void requestLocationUpdates() {
     // Common params
     LocationRequest request = LocationRequest.create()
-      .setInterval(1000)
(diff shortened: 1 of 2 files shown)


mapbox/libandroid-services/src/main/java/com/mapbox/services/android/location/LostLocationEngine.java
```diff
@@ -89,9 +89,9 @@ public Location getLastLocation() {
   public void requestLocationUpdates() {
     // Common params
     LocationRequest request = LocationRequest.create()
-      .setInterval(1000)
-      .setFastestInterval(1000)
-      .setSmallestDisplacement(3.0f);
+      .setInterval(interval)
```
zugaldia: @cammace I like where this is going, particularly the great javadoc added. I'd like to make this fields optional to set, like GMS does ([example](https://developers.google.com/android/reference/com/google/android/gms/location/LocationRequest.html#setFastestInterval(long)). How about we make them objects instead with a default value of `null`?

PR #416, opened 2017-03-27
Added infinity exception to if statement inside Position object

cammace:


mapbox/libjava-core/src/main/java/com/mapbox/services/commons/models/Position.java
```diff
@@ -31,15 +31,15 @@ private Position(double longitude, double latitude, double altitude) {
     this.latitude = latitude;
     this.altitude = altitude;
 
-    if (latitude < -90 || latitude > 90) {
+    if (latitude < -90 || latitude > 90 && latitude != Double.POSITIVE_INFINITY) {
```
zugaldia: Let's do it a bit more readable with `latitude != Double.POSITIVE_INFINITY && (latitude < -90 || latitude > 90)`

zugaldia:
@cammace Note the `()` in my comment: `latitude != Double.POSITIVE_INFINITY && (latitude < -90 || latitude > 90)`

Issue #338, opened 2017-02-20
When building the directions API call we add trailing zeros to digits


cammace:
This request
```
directions/v5/mapbox/driving/13.430100,52.510900;13.426500,52.508000;13.431600,52.502100?access_token=pk.XXX&geometries=polyline&bearings=60.000000,45.000000;;45.000000,45.000000
```
shows that we are forcing 6 digits and causing trailing zeros.

cc: @zugaldia

PR #424, opened 2017-03-30
Remove trailing zeros from coordinates

cammace:
closes https://github.com/mapbox/mapbox-java/issues/338

### DIFF
--- a/mapbox/libjava-core/src/main/java/com/mapbox/services/commons/models/Position.java
+++ b/mapbox/libjava-core/src/main/java/com/mapbox/services/commons/models/Position.java
@@ -1,8 +1,8 @@
 package com.mapbox.services.commons.models;
 
 import com.mapbox.services.Constants;
+import com.mapbox.services.commons.utils.TextUtils;
 
-import java.text.DecimalFormat;
 import java.util.logging.Logger;
 
 /**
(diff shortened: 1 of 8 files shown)


mapbox/libjava-core/src/main/java/com/mapbox/services/commons/models/Position.java
```diff
@@ -27,24 +28,27 @@
    * @since 1.0.0
    */
   private Position(double longitude, double latitude, double altitude) {
+    // Used to remove any trailing zeros and prevent a coordinate being over 7 significant figures.
+    DecimalFormat decimalFormat = new DecimalFormat("0.######");
```
zugaldia: Because this same logic is used in several place, how about exposing it as static utility method?

mapbox/libjava-core/src/main/java/com/mapbox/services/commons/utils/TextUtils.java
```diff
@@ -47,4 +51,10 @@ public static String join(CharSequence delimiter, Object[] tokens) {
     return sb.toString();
   }
 
+  public static String formatCoordinate(double coordinate) {
+    // Used to remove any trailing zeros and prevent a coordinate being over 7 significant figures.
```
zugaldia: @cammace Let's make it a javadoc instead.

mapbox/libjava-core/src/main/java/com/mapbox/services/commons/utils/TextUtils.java
```diff
@@ -47,4 +51,10 @@ public static String join(CharSequence delimiter, Object[] tokens) {
...
+    DecimalFormat decimalFormat = new DecimalFormat("0.######", new DecimalFormatSymbols(Locale.US));
+    return String.format(Locale.US, "%s",
```
zugaldia: @cammace We have a constant for the default locale already.

PR #431, opened 2017-04-06
Add local aar and jar generation support

Guardiola31337:
- [x] Added local `aar` and `jar` generation support
In order to build locally (same procedure to both Android and Java libraries):
1. Set `IS_LOCAL_DEVELOPMENT` gradle property to `true` (is `false` by default)
2. `$> gradlew uploadArchives`
3. `aar` or `jar` will be automatically installed in your Maven local repo folder (`.m2`/`M2_HOME`)
4. Remember to add `mavenLocal()` in `repositories{}` (`build.gradle`) in the project in which you'd like to use Maven local repository
- [x] Fixed javadoc issue

👀  @zugaldia

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,25 +41,25 @@ javadoc:
 	cd mapbox; ./gradlew :libandroid-ui:javadocrelease
 
 publish-java:
-	cd mapbox; ./gradlew :libjava-core:uploadArchives
-	cd mapbox; ./gradlew :libjava-geojson:uploadArchives
-	cd mapbox; ./gradlew :libjava-services:uploadArchives
-	cd mapbox; ./gradlew :libjava-services-rx:uploadArchives
+	cd mapbox; export IS_LOCAL_DEVELOPMENT=false; ./gradlew :libjava-core:uploadArchives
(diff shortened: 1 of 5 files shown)


zugaldia:
Bravo. This is a great solution.

> Set IS_LOCAL_DEVELOPMENT gradle property to true (is false by default)

Per chat, let's explore how to set `IS_LOCAL_DEVELOPMENT` as an environment variable, with the plugin defaulting to `true` to make public releases more explicit. That way, we'd need to update our Bitrise rules to be like `IS_LOCAL_DEVELOPMENT=false ./gradlew uploadArchives` before uploading the artifacts.

Guardiola31337:
@zugaldia done 😉 
Now if you want to install `aar`s or `jar`s in your Maven local repository you should define `IS_LOCAL_DEVELOPMENT` environment variable like this:

`$> export IS_LOCAL_DEVELOPMENT=true`

And `$> export IS_LOCAL_DEVELOPMENT=false` if you want same behavior as before. If `IS_LOCAL_DEVELOPMENT` is not exported also works as before.

zugaldia:
Next, I want `isLocalBuild` to return `true` by default to make releases more explicit. Let's ticket that separately.

PR #450, opened 2017-04-20
Added toString, hashcode, and equal methods

cammace:
This PR adds the standard `toString` `hashcode` and `equal` methods to model classes. Previously, if you tried comparing two matching objects, it would return false, this fixes that.

- [ ] Test should be added
- [ ] Javadoc needs to be added

### DIFF
--- a/mapbox/libjava-geojson/src/main/java/com/mapbox/services/commons/geojson/BaseFeatureCollection.java
+++ b/mapbox/libjava-geojson/src/main/java/com/mapbox/services/commons/geojson/BaseFeatureCollection.java
@@ -56,4 +56,30 @@ public class BaseFeatureCollection implements GeoJSON {
     return gson.create().toJson(this);
   }
 
+  @Override
+  public boolean equals(Object o) {
+    if (this == o) {
+      return true;
+    }
+    if (o == null || getClass() != o.getClass()) {
+      return false;
+    }
+
+    BaseFeatureCollection that = (BaseFeatureCollection) o;
+
+    return type.equals(that.type);
+
+  }
+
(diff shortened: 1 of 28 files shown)


cammace:
> I'd like to propose an alternative approach to automate this process

My original hesitation for doing this was the overhead that comes with maintaining all these methods. Any means to automate this process sounds great to me.

cammace:
Since AutoValue will require a bunch of API breaking changes we are holding off till 3.0.0 gets released (discussed in https://github.com/mapbox/mapbox-java/issues/499). Moving forward with this PR in the meantime. Can I get your review again @zugaldia

zugaldia:
Were all tests and Javadoc added to the PR, or is this still pending as per OP?

cammace:
I chose not to include test and Javadoc since we plan to switch to AutoValues very soon (which won't require us to test this).

cammace:
If you agree that we don't need to add test/javadoc for this, could I get another round of reviews?

cammace:
Removing from the 2.2.0 milestone since this isn't an immediate feature needed and most likely the 3.0 release will bring AutoValues which handles this for us.

Guardiola31337:
Noting here that when using [Kotlin Data Classes](https://kotlinlang.org/docs/reference/data-classes.html) the compiler automatically derives `equals()`/`hashCode()` and `toString()` providing everything you need to make AutoValue (in most cases) not necessary (avoiding a 3rd party dependency).

cammace:
AutoValue works coming in 3.0.0 release which resolves all the things mentioned in this PR, closing.

PR #469, opened 2017-05-26
added congestion annotation

cammace:


cammace:
Spoke with @ericrwolfe and since this doesn't expose much other than getters and setters, going ahead and merging into master.

PR #476, opened 2017-06-08
mock location

cammace:
This moves the mockLocationEngine to a more central location rather than including it in several different projects and having to maintain them independently.


zugaldia:
Sounds good. Let's remove it then from any other places it's being used like `mapbox/app/src/main/java/com/mapbox/services/android/testapp/location/MockLocationEngine.java`.

cammace:
Removed the old mock location and changed locationEngineActivity to use the new one.

Issue #419, opened 2017-03-30
Improve permissions management


Guardiola31337:
Per [Permissions wrong requested for devices with API level 22 or lower](https://github.com/mapbox/mapbox-gl-native/issues/8424) we should improve how permissions are managed and ensure consistency between apps.

We should consider the following options:

- Use a 3rd party library like [Dexter](https://github.com/Karumi/Dexter)
  - Not reinventing the wheel :white_check_mark:
  - Simplifies the process of requesting permissions at runtime :white_check_mark:
  - Adds a new dependency to SDK 🤔 
  - [Caveat](https://commonsware.com/blog/2015/11/09/you-cannot-hold-nonexistent-permissions.html) ❌ 
- Improve [`PermissionsManager`](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/permissions/PermissionsManager.java)
  - Reinventing the wheel? 🤔 
  - Not adding a new dependency :white_check_mark:
  - Deal with the [caveat](https://commonsware.com/blog/2015/11/09/you-cannot-hold-nonexistent-permissions.html) mentioned before ❌ 

In both options Android deals with when [permissions](https://developer.android.com/training/permissions/requesting.html) are requested (pre/post Android 6.0) and shows the dialogs wisely :white_check_mark:

Thoughts? Comments?

@mapbox/android

cammace:
Note that both [coarse and fine location permissions](https://developer.android.com/reference/android/Manifest.permission.html#ACCESS_COARSE_LOCATION) have been around since API 1 and shouldn't fall into the issue described in the article above. Perhaps we should narrow the scope of what the permission manager can handle or indeed fix this up. usage of the permission manager is optional anyways, and the developer has the option to handle it manually or use a third party library such as Dexter.

zugaldia:
I agree with @cammace above. Let's start by listing which specific features we need, that `PermissionsManager` isn't currently providing. Depending on the list of requirements, we could then decide whether an external dependency is worth it.

zugaldia:
From what I can see in https://github.com/mapbox/mapbox-gl-native/pull/8559/files, those two methods can easily be brought into `PermissionsManager` and an external dependency wouldn't be required (unless there are additional requirements).

cammace:
@Guardiola31337 what's the status of this issue? I am going ahead and removing it from the 2.2.0 milestone.

[...]

PR #477, opened 2017-06-08
Add check to determine if runtime permissions are required

tobrun:
Refs #419

### DIFF
--- a/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/permissions/PermissionsManager.java
+++ b/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/permissions/PermissionsManager.java
@@ -4,6 +4,7 @@ import android.Manifest;
 import android.app.Activity;
 import android.content.Context;
 import android.content.pm.PackageManager;
+import android.os.Build;
 import android.support.v4.app.ActivityCompat;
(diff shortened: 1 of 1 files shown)


cammace:
Looks like CI's failing with this error:
```
:libandroid-telemetry:compileReleaseJavaWithJavac - is not incremental (e.g. outputs have changed, no previous execution, etc.).
/bitrise/src/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/permissions/PermissionsManager.java:55: error: package Build does not exist
    return android.os.Build.VERSION.SDK_INT >= Build.VERSION_CODES.M;
                                                    ^
1 error
:libandroid-telemetry:compileReleaseJavaWithJavac FAILED
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':libandroid-telemetry:compileReleaseJavaWithJavac'.
> Compilation failed; see the compiler error output for details.
```

tobrun:
Hah, it seems that there aren't any android dependency available there, instead of checking against BUILD.x.xx, we can check against a constant int instead.

zugaldia:
Weird, this is an Android library targeting Nougat, why wouldn't it have `Build.VERSION_CODES` available?

cammace:
The class was just missing the build import. fixed.

PR #479, opened 2017-06-12
Fix static marker annotation url generation

Guardiola31337:
Fixes [mapbox/mapbox-gl-native#9107](https://github.com/mapbox/mapbox-gl-native/issues/9107)
- [x] Fix tests
- [x] Add new tests
- [ ] Verify when a label is not a valid Maki icon (not blocker)

👀 @zugaldia @cammace

### DIFF
--- a/mapbox/libjava-services/src/main/java/com/mapbox/services/api/staticimage/v1/models/StaticMarkerAnnotation.java
+++ b/mapbox/libjava-services/src/main/java/com/mapbox/services/api/staticimage/v1/models/StaticMarkerAnnotation.java
@@ -44,7 +44,7 @@ public class StaticMarkerAnnotation {
           TextUtils.formatCoordinate(builder.getLat(), builder.getPrecision())
         );
       } else {
-        marker = String.format(Constants.DEFAULT_LOCALE, "%s-%s+%s(%s,%s)",
+        marker = String.format(Constants.DEFAULT_LOCALE, "%s%s%s(%s,%s)",
           builder.getName(),
           builder.getLabel(),
           builder.getColor(),
@@ -58,7 +58,7 @@ public class StaticMarkerAnnotation {
         marker = String.format(Constants.DEFAULT_LOCALE, "url-%s(%s,%s)",
           builder.getUrl(), builder.getLon(), builder.getLat());
       } else {
-        marker = String.format(Constants.DEFAULT_LOCALE, "%s-%s+%s(%f,%f)", builder.getName(),
+        marker = String.format(Constants.DEFAULT_LOCALE, "%s%s%s(%f,%f)", builder.getName(),
           builder.getLabel(), builder.getColor(), builder.getLon(), builder.getLat());
       }
     }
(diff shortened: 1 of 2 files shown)


mapbox/libjava-services/src/main/java/com/mapbox/services/api/staticimage/v1/models/StaticMarkerAnnotation.java
```diff
@@ -167,7 +177,11 @@ public Builder setLabel(String label) {
      * @since 2.1.0
      */
     public String getColor() {
-      return color;
+      if (color.isEmpty()) {
+        return EMPTY;
```
cammace: Why not just return null if a color isn't set?

mapbox/libjava-services/src/main/java/com/mapbox/services/api/staticimage/v1/models/StaticMarkerAnnotation.java
```diff
@@ -144,12 +149,17 @@ public Builder setName(String name) {
      * @since 2.1.0
      */
     public String getLabel() {
-      return label;
+      if (label.isEmpty()) {
+        return EMPTY;
```
cammace: Why not just return null if a label isn't set?

mapbox/libjava-services/src/test/java/com/mapbox/services/api/staticimage/v1/StaticMarkerAnnotationTest.java
```diff
@@ -65,8 +65,76 @@ public void requireMarkerLon() throws ServicesException {
     new StaticMarkerAnnotation.Builder().setName(Constants.PIN_SMALL).setLat(2.0).build();
   }
 
+  @Test(expected = ServicesException.class)
```
cammace: Should we be for explicit on which ServiceException we expect to occur here?

mapbox/libjava-services/src/main/java/com/mapbox/services/api/staticimage/v1/models/StaticMarkerAnnotation.java
Guardiola31337: @cammace `null` is not valid because `getLabel` is called when parsing (`String.format`) and `label` is an optional param (note that `-`was removed from the default string because if the `label` was empty url was incorrect. A [Null Object pattern](https://sourcemaking.com/design_patterns/null_object) could be implemented though.

BTW `null` is considered _The worst mistake of computer science_: [Null References: The Billion Dollar Mistake](https://www.infoq.com/presentations/Null-References-The-Billion-Dollar-Mistake-Tony-Hoare) 😉

mapbox/libjava-services/src/main/java/com/mapbox/services/api/staticimage/v1/models/StaticMarkerAnnotation.java
Guardiola31337: Same here.

mapbox/libjava-services/src/test/java/com/mapbox/services/api/staticimage/v1/StaticMarkerAnnotationTest.java
Guardiola31337: IMO just testing that an exception is thrown when an url is not valid is 👌
If more explicitness needed, what about creating a new specific exception?
BTW, is that explicitness really needed in this case?

mapbox/libjava-services/src/main/java/com/mapbox/services/api/staticimage/v1/models/StaticMarkerAnnotation.java
cammace: lol, I've read that interview 😄 . Sounds good.

PR #491, opened 2017-06-22
AutoValue

cammace:
Looking into the AutoValue suggestion in https://github.com/mapbox/mapbox-java/pull/450, I successfully converted two model classes to see how this would work. There are going to be significant API breaking changes if we want to fully implement this. I also didn't mess around with adding Builders for each class since you can call `create()` passing in all the params and receive a built object.

Using AutoValues also enforces classes to be immutable which makes sense for these models, you can currently call `setRoute` at any time after constructed.

Would love to hear your thoughts on this @zugaldia

### DIFF
--- a/mapbox/build.gradle
+++ b/mapbox/build.gradle
@@ -1,34 +1,35 @@
 // Top-level build file where you can add configuration options common to all sub-projects/modules.
 
 buildscript {
-    repositories {
-        jcenter()
-        mavenCentral()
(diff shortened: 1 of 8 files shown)


zugaldia:
@cammace Please let me know once CI is happy and I'll give this one a pass.

cammace:
I'm going to go ahead and close this PR in favor of a new one which moves logic over to AutoValue. This PR was more of an experiment to learn about the cost and benefits of using AutoValue.

PR #493, opened 2017-06-30
downgrade LOST to v1.1.1

tobrun:
Copied from downstream ticket in https://github.com/mapbox/mapbox-gl-native/pull/9394:

> We have been having some issues with recent releases of the LOST library. To be able to ship a next final release for 5.1.0. We are downgrading back to `v1.1.1`. This version shipped as part of our 4.x series.

This PR requires a newer version of the SDK first to pass CI, excluding lost from the maps sdk makes the test application crash at startup.

### DIFF
--- a/mapbox/app/src/main/java/com/mapbox/services/android/testapp/location/GoogleLocationEngine.java
+++ b/mapbox/app/src/main/java/com/mapbox/services/android/testapp/location/GoogleLocationEngine.java
@@ -88,7 +88,7 @@ public class GoogleLocationEngine extends LocationEngine implements
 
   @Override
   public Location getLastLocation() {
-    if (googleApiClient.isConnected() && PermissionsManager.areLocationPermissionsGranted(context.get())) {
+    if (googleApiClient.isConnected()) {
       //noinspection MissingPermission
       return LocationServices.FusedLocationApi.getLastLocation(googleApiClient);
     }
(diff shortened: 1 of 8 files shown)


mapbox/libandroid-services/src/main/java/com/mapbox/services/android/location/LostLocationEngine.java
```diff
@@ -44,50 +38,63 @@ public static synchronized LocationEngine getLocationEngine(Context context) {
...
+   *
+   * @return boolean true if the location engine has been activated/connected, else false.
+   */
   @Override
   public boolean isConnected() {
     return lostApiClient.isConnected();
   }
 
+  /**
+   * Returns the Last known location is the location provider is connected and location permissions are granted.
```
Guardiola31337: Minor typo `if` instead of `is`.

tobrun: great catch

mapbox/app/src/main/java/com/mapbox/services/android/testapp/location/GoogleLocationEngine.java
```diff
@@ -88,7 +87,7 @@ public void onConnectionFailed(@NonNull ConnectionResult connectionResult) {
 
   @Override
   public Location getLastLocation() {
-    if (googleApiClient.isConnected() && PermissionsManager.areLocationPermissionsGranted(context.get())) {
+    if (googleApiClient.isConnected()) {
```
tobrun: `PermissionsManager.areLocationPermissionsGranted(context.get()` is broken for pre Marshmallow devices

PR #513, opened 2017-07-18
added test and exception if null point exist during operations

### LINKED ISSUE #459: Throw exception when origin and destination positions are the same
mirror ticket in navigation SDK repo: https://github.com/mapbox/mapbox-navigation-android/issues/40

If destination and origin are identical, directions API and turf methods might fail (test need to be added to confirm this).

cc: @zugaldia

cammace:
Closes #459


zugaldia:
Fix checksyle and 🚀 .

PR #530, opened 2017-08-02
Add exit field in LegStep

anant2614:
Add exit field in LegStep class

Test added:
testLegStepExit() in MapboxDirectionsTest.java


mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/models/LegStep.java
```diff
@@ -325,4 +326,19 @@ public String getPronunciation() {
...
+  /**
+   * A String with the exit numbers or names of the way. Optionally included, if data is available
+   * @return String depicting exit
+   */
+  public String getExit() {
+    return exit;
+  }
+
+  /**
+   * A String with the exit numbers or names of the way. Optionally included, if data is available
```
cammace: Missing `@param` javadoc and `@since 3.0.0` tags

mapbox/libjava-services/src/main/java/com/mapbox/services/api/directions/v5/models/LegStep.java
```diff
@@ -325,4 +326,19 @@ public String getPronunciation() {
   public void setPronunciation(String pronunciation) {
     this.pronunciation = pronunciation;
   }
+
+  /**
+   * A String with the exit numbers or names of the way. Optionally included, if data is available
+   * @return String depicting exit
```
cammace: check style will complain about there not being a line between the summary and `@return` tag.

mapbox/libjava-services/src/test/fixtures/directions_v5_fixtures_rotary.json
```diff
@@ -190,7 +190,8 @@
               "duration": 150.6,
               "rotary_name": "Dupont Circle Northwest",
               "distance": 189.9,
-              "name": "P Street Northwest"
+              "name": "P Street Northwest",
+              "exit": "M Street"
```
cammace: Was the fixture updated using `make directions-fixtures` or was the value manually placed in here? If the latter, it will be difficult later down the road when we need to update the fixtures again and the `exit` value's missing resulting in a failing test. @1ec5 Do you guys have a fixture which includes the step exit?

mapbox/libjava-services/src/test/java/com/mapbox/services/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -455,6 +455,29 @@ public void testLegStep() throws ServicesException, IOException {
...
+            .setBaseUrl(mockUrl.toString())
+            .build();
+    Response<DirectionsResponse> response = client.executeCall();
+
+    StepManeuver maneuver = response.body().getRoutes().get(0).getLegs().get(0).getSteps().get(1).getManeuver();
+    assertEquals(maneuver.getExit(), Integer.valueOf(3));
+    LegStep step = response.body().getRoutes().get(0).getLegs().get(0).getSteps().get(1);
+    assertEquals(step.getRotaryName(), "Dupont Circle Northwest");
+    assertEquals(step.getExit(),"M Street");
+    assertEquals(step.getRotaryPronunciation(), null);
```
cammace: Is there a reason for checking that `getRotaryPronunciation` and `getPronunciation` are null? Since the test is named `testLegStepExit` it should only be checking that the `step.getExit()` is correct.

cammace:
Closing since we have added this in 3.0.0

PR #544, opened 2017-08-23
Add setter for session rotation time

zugaldia:
cc: @Guardiola31337 @tobrun @cammace


zugaldia:
@cammace Ready for you. Thanks @Guardiola31337 for the quick 👀 .

PR #545, opened 2017-08-23
3.0-changes

cammace:
AutoValues being introduced in 3.0 for a number of reasons. This PR is where we will track all the changes and progress while making the conversion. Aside from AutoValues, we are also using Android's Annotation support library for limiting parameter scope, annotating null/nonnull, etc.


cammace:
The latest implementation of this code has been added into https://github.com/mapbox/mapbox-java/pull/580, closing.

PR #549, opened 2017-08-29
Introduce setDebugLoggingEnabled API

zugaldia:
This API lets us manually control the verbosity of HTTP requests.

### DIFF
--- a/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/http/TelemetryClient.java
+++ b/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/http/TelemetryClient.java
@@ -227,7 +227,7 @@ public class TelemetryClient {
     RequestBody body = RequestBody.create(JSON, payload);
     String url = getEventsEndpoint() + "/events/v2?access_token=" + getAccessToken();
 
-    // Extra debug in staging mode
+    // Extra debug if manually enabled or if in staging mode
     if (isDebugLoggingEnabled() || isStagingEnvironment()) {
       Log.d(LOG_TAG, String.format("Sending POST to %s with %d event(s) (user agent: %s) with payload: %s",
         url, events.size(), getUserAgent(), payload));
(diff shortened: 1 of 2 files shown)


mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/http/TelemetryClient.java
```diff
@@ -219,7 +228,7 @@ private void sendEventsWrapped(Vector<Hashtable<String, Object>> events, Callbac
     String url = getEventsEndpoint() + "/events/v2?access_token=" + getAccessToken();
 
     // Extra debug in staging mode
```
Guardiola31337: This comment is now lying. This is an example of why I'm not a big fan of comments :trollface:

mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/http/TelemetryClient.java
```diff
@@ -219,7 +228,7 @@ private void sendEventsWrapped(Vector<Hashtable<String, Object>> events, Callbac
...
-    if (isStagingEnvironment()) {
+    if (isDebugLoggingEnabled()) {
```
Guardiola31337: @zugaldia 
If we're going to go this way, we need to change

https://github.com/mapbox/mapbox-java/blob/e4842cbe98840451503a9d3153ab4124226e2211/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java#L255

with `MapboxTelemetry#setDebugLoggingEnabled(true)` and deprecate/adapt

https://github.com/mapbox/mapbox-java/blob/e4842cbe98840451503a9d3153ab4124226e2211/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/http/TelemetryClient.java#L38

its getter and its setter

https://github.com/mapbox/mapbox-java/blob/e4842cbe98840451503a9d3153ab4124226e2211/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/http/TelemetryClient.java#L76-L82

Another possibility is adding `|| isStagingEnvironment()` and not break anything 😉 
Right now, we're not logging when in staging mode if we don't set `debugLoggingEnabled` to `true`.

zugaldia: Good call, let's do `|| isStagingEnvironment()` for now. Fixing up in a bit.

PR #556, opened 2017-09-05
Map out all navigation events

cammace:
Adds a for loop which adds all the metadata for the events, not sure why this wasn't the way it was done the first place.


zugaldia:
> not sure why this wasn't the way it was done the first place.

I believe that the original author wanted to have more control of which events were added and its formatting. What are the performance penalties of now (trying) to add every key twice?

cammace:
> What are the performance penalties of now (trying) to add every key twice?

I say we remove the old keys and simply use the for loop but for the sake of getting this done and moving onto navigation SDK without messing anything up for the map sdk I left them there. I'd prefer the for loop since we have so many meta fields for nav which would need to be handwritten otherwise and leaves the potential of leaving one out... Not sure about the performance hit though.

PR #561, opened 2017-09-13
[WIP] Added hashcode, equals and toString implementations

osanchik:
1) Added toString implementation to LocationEngine.

2) Added hashCode, equals implementations to LocationEngine. 
This is not enough though. For equals to work correctly EACH of the subclasses need to override it and call super.equals().  Otherwise we could have false equals.

3) Added EqualsTester to test equals() on two Objects. It would be good to have a HashCodeTester as well. Added LocationEngineTest which tests and compares custom LocationEngine instances.

cc: @zugaldia , @cammace


cammace:
@Guardiola31337 do we want to port this over to the telemetry library repo?

cammace:
Go ahead and closing this since I never got a response from @Guardiola31337 on the telemetry side. Feel free to port this code over to the new repo whenever.

PR #568, opened 2017-09-19
Introduce new user agent API

Guardiola31337:
- Adds `newUserAgent(String userAgent)`

👀 @cammace @ericrwolfe @zugaldia

### DIFF
--- a/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
+++ b/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxTelemetry.java
@@ -35,7 +35,9 @@ import com.mapbox.services.android.telemetry.utils.TelemetryUtils;
 
 import java.io.IOException;
 import java.math.BigDecimal;
+import java.util.ArrayList;
 import java.util.Hashtable;
+import java.util.List;
 import java.util.Timer;
 import java.util.TimerTask;
 import java.util.Vector;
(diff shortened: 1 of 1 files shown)


danesfeder:
@ericrwolfe @Guardiola31337 When I create a new instance `MapboxNavigation`, it will initialize the telem library.  

Will it be initialized with `userAgent` set to "mapbox-navigation-android"?  

If someone launches the drop-in UI, it should be be changed to "mapbox-navigation-android-ui", so I will need to add a line in the UI code after I initialize `MapboxNavigation` to set it to "mapbox-navigation-android-ui" right?

ericrwolfe:
> If someone launches the drop-in UI, it should be be changed to "mapbox-navigation-android-ui", so I will need to add a line in the UI code after I initialize MapboxNavigation to set it to "mapbox-navigation-android-ui" right?

Yep that's correct. Here's where it's done on iOS: 
* Setting the user agent: https://github.com/mapbox/mapbox-navigation-ios/blob/master/MapboxCoreNavigation/RouteController.swift#L222-L223
* Setting the user interface flag: https://github.com/mapbox/mapbox-navigation-ios/blob/e4eb39bdece779b29f3c19b916d0eca16b39064c/MapboxNavigation/NavigationViewController.swift#L341-L343

Per my comment above https://github.com/mapbox/mapbox-java/pull/568#pullrequestreview-63761426, you'll probably need to pass a full user agent string with a version, such as `mapbox-navigation-ui-android/0.6.0-SNAPSHOT`, into `newUserAgent()`.

danesfeder:
Okay sounds good - adding to list of things to do in the Nav SDK once MAS lands

Guardiola31337:
> you'll probably need to pass a full user agent string with a version, such as `mapbox-navigation-ui-android/0.6.0-SNAPSHOT`, into `newUserAgent()`.

That's correct.

After initializing telemetry you should add a call to `MapboxTelemetry.getInstance().newUserAgent("<fullUserAgent>");`

Merging it.

Issue #575, opened 2017-09-25
Support API level 14


zugaldia:
We currently support API level 15 as our min SDK. However, we hear from developers that it'd be useful to lower that requirement to API level 14. This will require changes to some of our dependencies that also require 15+. Let's use this ticket to research whether this is a possibility.

cc: @cammace

cammace:
In 3.0 no all our modules will not have any dependency on Android, therefore, this ticket can be closed. This is also being tracked in the telem repo in https://github.com/mapbox/mapbox-telemetry-android/issues/100.

PR #576, opened 2017-09-25
Lower API level min SDK to 14

zugaldia:
Fixes https://github.com/mapbox/mapbox-java/issues/575.

### DIFF
--- a/mapbox/app/src/main/AndroidManifest.xml
+++ b/mapbox/app/src/main/AndroidManifest.xml
@@ -1,5 +1,6 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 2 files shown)


zugaldia:
CI is currently failing due to issues with the Test App depending on libraries (Maps SDK, LOST) that requires 15+. Looks like otherwise the library is in good shape to support 14. Let's revisit once https://github.com/mapbox/mapbox-gl-native/issues/9403 lands.

cc: @cammace

langsmith:
Let's make sure to let the support team know about this because documentation might have to be updated and they periodically answer Help Scout tickets related to min version

cammace:
Note that in 3.0.0 this is a pure java repo and won't have an API level :)

zugaldia:
> Note that in 3.0.0 this is a pure java repo and won't have an API level :)

Correct. This requirement is now for the standalone telemetry library. @electrostat could you keep track of this one?

cammace:
Closing since in 3.0 there will be no API level to lower :)

PR #591, opened 2017-10-06
Location serializer added

cammace:


mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/navigation/NavigationLocation.java
```diff
@@ -0,0 +1,58 @@
...
+    }
+    return jsonArray;
+  }
+
+  private JSONObject serializeToJson(Location location) {
+    formatLocationTime(location);
+
+    HashMap<String, Object> map = new HashMap<>();
+    // Potentially null values
+    map.put(ALTITUDE, location.hasAltitude() ? location.getAltitude() : JsonToken.NULL);
```
Guardiola31337: Why `JsonToken.NULL` instead of `JSONObject.NULL`?

mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/navigation/NavigationLocation.java
```diff
@@ -0,0 +1,58 @@
...
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.TimeZone;
+
+public class NavigationLocation {
+
+  private static final String UTC_TIME_FORMAT = "yyyy-MM-dd'T'HH:mm:ss.SSSZ";
+  private static final String ACCURACY = "horizontalAccuracy";
```
Guardiola31337: Almost all of these variables are duplicated.
👀 [`MapboxEvent`](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/MapboxEvent.java) and [`MapboxNavigationEvent`](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/navigation/MapboxNavigationEvent.java)

mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/navigation/NavigationLocation.java
cammace: Because the values null, not the entire object. It doesn't really matter though, they result in the samething.

mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/navigation/NavigationLocation.java
cammace: doesn't matter, this is temporary

mapbox/libandroid-telemetry/src/main/java/com/mapbox/services/android/telemetry/navigation/NavigationLocation.java
Guardiola31337: Wondering why not keeping consistency with the rest of the code then 🤔

PR #592, opened 2017-10-06
Step Metadata is serialized improperly

### LINKED ISSUE #585: Navigation event step isn't getting serialized correctly
In a reroute event part of the metadata sent over is step information which should be in its own json object. Right now however, it is being serialized as a string.

```
"step": "{distance=23, previousType=depart, upcomingType=turn, previousModifier=right, duration=14, upcomingModifier=left, distanceRemaining=23, previousName=West 21st Place, upcomingName=South Rockwell Street, durationRemaining=13, upcomingInstruction=Turn left onto South Rockwell Street, previousInstruction=Head east on West 21st Place}",
```

cc: @Guardiola31337 @electrostat @zugaldia

electrostat:
Closes #585

Current iteration returns a string instead of a json object.


electrostat:
closes #585

PR #593, opened 2017-10-06
placed things back to their original location to not break API

cammace:
LOST is added as a dependency again inside libandroid-services and the LOST location provider is duplicated back into it's original package but deprecated now.

### DIFF
--- a/mapbox/libandroid-services/build.gradle
+++ b/mapbox/libandroid-services/build.gradle
@@ -42,6 +42,11 @@ dependencies {
     // Timber for logging
     compile rootProject.ext.dep.timber
 
+    // LOST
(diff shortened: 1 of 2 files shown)


mapbox/libandroid-services/build.gradle
```diff
@@ -42,6 +42,11 @@ dependencies {
     // Timber for logging
     compile rootProject.ext.dep.timber
 
+    // LOST
+    compile(rootProject.ext.dep.lost) {
```
Guardiola31337: @cammace 
This could be `provided` and per Lost 3.x no need to exclude guava.

```
provided rootProject.ext.dep.lost
```

PR #598, opened 2017-10-09
Upgrade Lost version

Guardiola31337:
- Upgrades Lost version to `3.0.3` but [`LostLocationEngine`](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libandroid-services/src/main/java/com/mapbox/services/android/location/LostLocationEngine.java) (now `@Deprecated`) from `libandroid-services`

👀 @zugaldia @cammace @danesfeder @electrostat

### DIFF
--- a/mapbox/app/build.gradle
+++ b/mapbox/app/build.gradle
@@ -60,9 +60,7 @@ dependencies {
     compile rootProject.ext.dep.gmsLocation
 
     // LOST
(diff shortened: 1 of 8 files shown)


Guardiola31337:
Fixed via https://github.com/mapbox/mapbox-java/pull/636

Closing.

PR #622, opened 2017-11-01
mapbox-android-core dependency

Guardiola31337:
- Adapts code so that it uses `mapbox-android-core:0.1.1` dependency

👀 @zugaldia @electrostat @lilykaiser @cammace


circle.yml
```diff
@@ -110,13 +110,13 @@ jobs:
       - deploy:
           name: Publish Java Services To Maven Central
           command: |
-            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+            if [ "${CIRCLE_BRANCH}" == "pg-android-core" ]; then
```
zugaldia: @Guardiola31337 I believe this needs to be changed back to `master` before merging.

Issue #623, opened 2017-11-01
Release 2.2.7


Guardiola31337:
### Release checklist

- [x] Change version name to `2.2.7` in `gradle.properties`
- [x] Publish `2.2.7` artifact to Maven Central
- [x] Change version name back to `2.3.0-SNAPSHOT`
- [x] Update `CHANGELOG.md`
- [x] Tag `v2.2.7` in GitHub
- [ ] Update dependencies in the Maps SDK

### Post-release @langsmith

- [ ] ~_If_ there's a new SNAPSHOT version, then update this repo's `README.md` file~
- [ ] Update Javadoc for the MAS docs section
- [ ] Update the `constants.js` file in `android-docs` repo so that `compile 'com.mapbox.mapboxsdk:mapbox-android-services:x.x.x'` line is updated in [the Gradle section of the MAS overview page](https://www.mapbox.com/android-docs/mapbox-services/overview/)
- [ ] Update the `compile 'com.mapbox.mapboxsdk:mapbox-android-ui:X.X.X'` line in the Android demo app's Gradle file
- [ ] Update version # in config files in /help @langsmith
- [ ] Update version # in [Store Locator kit](https://github.com/mapbox/store-locator-android)

/cc: @mapbox/mobile

Guardiola31337:
👉 #628 
Closing.

PR #624, opened 2017-11-01
Release 2.2.7

Guardiola31337:
- Addresses #623 

Tagging @mapbox/android because this release will need the Maps SDK and the Navigation SDK to bump MAS version number to `2.2.7`.

After merging https://github.com/mapbox/mapbox-java/pull/622 builds are failing because now the `LocationEngine` is taken from `mapbox-android-core` so we need to update the dependencies in all the projects that use the one included in `mapbox-telemetry-android`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 Mapbox welcomes participation and contributions from everyone.
 
+### v2.2.7
+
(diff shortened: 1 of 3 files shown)


Guardiola31337:
#622 was reverted in #627 
Closing here.

PR #647, opened 2017-11-19
Fix sonarqube issues

cammace:


samples/src/main/java/com/mapbox/samples/BasicDirections.java
```diff
@@ -1,60 +1,140 @@
...
+      + response.body().routes().get(0).distance()));
+
+
+
+//
+//
+//
+//
+//
+//    return MapboxDirections.builder()
```
tobrun: do we need to keep this commented out codeblock?

PR #665, opened 2017-11-28
Fix CI on master branch

### LINKED ISSUE #648: Fix transifex translating ns
Since switching the code base over to what will become 3.0. a bunch of filepaths have been broken including the transfax ones.

cammace:
Closes #648 

Also enforces we only send new info to SonarQube that's coming from master and not any branch.


circle.yml
```diff
@@ -33,16 +33,9 @@ jobs:
...
       - run:
           name: SonarQube upload
-          command: make sonarqube
-      - run:
-          name: Update Transifex
           command: |
-            if [ "${CIRCLE_BRANCH}" == "${DEPLOY_BRANCH}" ]; then
-              pip install transifex-client
-              echo $'[https://www.transifex.com]\nhostname = https://www.transifex.com\nusername = '"${TRANSIFEX_USER}"$'\npassword = '"${TRANSIFEX_API_TOKEN}"$'\ntoken = '""$'\n' > ~/.transifexrc
-              tx push -s
```
1ec5: Instead of pushing explicitly, you can set up Transifex to autoupdate the resources [from the Web interface](https://www.transifex.com/mapbox/mapbox-java/content/). It updates at least daily based on the raw URLs you provide.

cammace: Do we want to do this now or in the future once we actually have strings to localize? We currently have nothing to translate in 3.0.

1ec5: What’ll happen to [these resources](https://www.transifex.com/mapbox/mapbox-java/content/) in 3.0? Does 3.0 have hard-coded strings in the code, or are there no user-facing strings in the first place?

cammace: no user-facing strings (besides of course the ones coming from APIs). There are no UI components in 3.0 now.

PR #670, opened 2017-11-29
Added testLocale test back into class

cammace:


services-directions/src/test/java/com/mapbox/directions/v5/MapboxDirectionsTest.java
```diff
@@ -450,4 +448,19 @@ public void noValidRouteTest() throws Exception {
     assertThat(response.body().message(), containsString("No route found"));
     assertThat(response.body().code(), containsString("NoRoute"));
   }
+
+  @Test
+  public void testLocale() {
+    // Locale shouldn't matter (#39)
```
Guardiola31337: I'd fix the name of the test based on the comment (also following the same nomenclature as in the rest of the tests ☝️).
This way the test will be more descriptive and maintainable and comment will become superfluous so it won't be necessary.

PR #680, opened 2017-12-06
Adds javac lint and more test

cammace:
Fixes a bug which caused Point serializing not to include altitude in JSON output.


services-geojson/src/test/java/com/mapbox/geojson/gson/GeometryDeserializerTest.java
```diff
@@ -0,0 +1,12 @@
+package com.mapbox.geojson.gson;
+
+import com.mapbox.core.TestUtils;
+import org.junit.Test;
+
+public class GeometryDeserializerTest extends TestUtils {
```
Guardiola31337: This is currently doing nothing. We should remove it.

services-geojson/src/main/java/com/mapbox/geojson/gson/BoundingBoxSerializer.java
```diff
@@ -16,14 +16,43 @@
...
+   *                  invocation of its {@link com.google.gson.JsonDeserializationContext} method
+   * @return a JsonArray containing the raw coordinates
+   * @since 3.0.0
+   */
   @Override
   public JsonElement serialize(BoundingBox src, Type typeOfSrc, JsonSerializationContext context) {
-
     JsonArray bbox = new JsonArray();
+
+    // Southwest
```
Guardiola31337: What about extracting this block of code into a private method and give it a name based on the comment? This way `serialize` will be easier to read and understand and comment will become superfluous so it won't be necessary.

services-geojson/src/main/java/com/mapbox/geojson/gson/BoundingBoxSerializer.java
```diff
@@ -16,14 +16,43 @@
...
     bbox.add(new JsonPrimitive(src.southwest().longitude()));
     bbox.add(new JsonPrimitive(src.southwest().latitude()));
+    if (src.southwest().hasAltitude()) {
+      bbox.add(new JsonPrimitive(src.southwest().altitude()));
+    }
+
+    // Northeast
```
Guardiola31337: Same here ☝️

PR #686, opened 2017-12-08
RxJava Module

cammace:


services-rxbinding/src/main/java/com/mapbox/rx/api/directions/v5/EnqueueCallObservable.java
```diff
@@ -0,0 +1,91 @@
...
+      if (isDisposed()) {
+        return;
+      }
+
+      try {
+        observer.onNext(response.body());
+        if (!call.isCanceled()) {
+          terminated = true;
+          observer.onComplete();
+        }
```
pakoito: What happens if the call is cancelled? Does the observable never complete?

cammace: hmm I might have switched up the call object coming from onResponse and the `isDisposed` methods, do you see an issue here?

pakoito: I'm trying to understand the logic. You always get the body, even in a cancellation, but if it is cancelled it just doesn't complete the Observable, which means several operators like toList() or concatMap won't work with this API.

cammace:
Going to go ahead and close this PR. Little discussion has occurred on our end on whether or not we still want to have a RxJava module specific to this library and until a compelling reason comes across as to why we need to support this, this PR will just continue sitting here stale. 

@zugaldia @lilykaiser @osana let me know if you guys want to sit down and discuss further actions on this.

Issue #633, opened 2017-11-08
Test that PolylineUtils.decode never returns null but rather an empty list


cammace:
and annotate it with `@NonNull`. This will simplify and prevent errors in the nav SDK.

PR #687, opened 2017-12-11
Adds annotations and test for polylineUtils

cammace:
Closes https://github.com/mapbox/mapbox-java/issues/633

Even with an empty list or path string, the returned results for all PolylineUtils results in a nonnull value, thus annotating the methods will help developers know they can skip over this check.

### DIFF
--- a/services-geojson/src/test/java/com/mapbox/geojson/utils/PolylineUtilsTest.java
+++ b/services-geojson/src/test/java/com/mapbox/geojson/utils/PolylineUtilsTest.java
@@ -1,12 +1,17 @@
 package com.mapbox.geojson.utils;
 
-import com.mapbox.core.constants.Constants;
-import com.mapbox.core.TestUtils;
-import com.mapbox.geojson.Point;
+import static com.mapbox.geojson.utils.PolylineUtils.decode;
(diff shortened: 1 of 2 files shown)


services-geojson/src/test/java/com/mapbox/geojson/utils/PolylineUtilsTest.java
```diff
@@ -1,24 +1,30 @@
...
   public void testDecodePath() {
-    List<Point> latLngs = PolylineUtils.decode(TEST_LINE, Constants.PRECISION_5);
+    List<Point> latLngs = decode(TEST_LINE, Constants.PRECISION_5);
 
     int expectedLength = 21;
-    Assert.assertEquals("Wrong length.", expectedLength, latLngs.size());
+    assertEquals("Wrong length.", expectedLength, latLngs.size());
 
     Point lastPoint = latLngs.get(expectedLength - 1);
     expectNearNumber(37.76953, lastPoint.latitude(), 1e-6);
```
osana: should we use a Constants.PRECISION_6 ?

cammace: The `TEST_LINE` was originally written for precision 5 a long, long time ago ;)

PR #689, opened 2017-12-11
add check and test for pointOnLine list less than 2

cammace:
I've also added a `checkNotNull` so we can use it throughout the SDKs.

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
@@ -1,11 +1,12 @@
 package com.mapbox.turf;
 
-import android.support.annotation.NonNull;
+import static com.mapbox.core.internal.Preconditions.checkNotNull;
 
-import com.mapbox.turf.models.LineIntersectsResult;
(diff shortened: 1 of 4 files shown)


services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
```diff
@@ -36,7 +37,10 @@ private TurfMisc() {
    * @see <a href="http://turfjs.org/docs/#lineslice">Turf Line slice documentation</a>
    * @since 1.2.0
    */
-  public static LineString lineSlice(Point startPt, Point stopPt, Feature line) {
+  @NonNull
+  public static LineString lineSlice(@NonNull Point startPt, @NonNull Point stopPt,
+                                     @NonNull Feature line) {
+    checkNotNull(line.geometry(), "Feature.geometry() == null");
```
osana: Can line.geometry().getType() be null?

cammace: The Feature geometry can be nullable but if the geometry isn't null, the type will never be null.

PR #690, opened 2017-12-11
Adds missing docs and cleans things up in static map module

cammace:


services-core/src/main/java/com/mapbox/core/utils/ColorUtils.java
```diff
@@ -0,0 +1,13 @@
+package com.mapbox.core.utils;
+
+import java.awt.Color;
+
+public final class ColorUtils {
+  public static String toHexString(Color color) {
+    String hexColour = Integer.toHexString(color.getRGB() & 0xffffff);
```
osana: Would something like this be better:
String hex = String.format("#%02x%02x%02x", r, g, b);
String hex = String.format("#%02X%02X%02X", color.getRed(), color.getGreen(), color.getBlue());

PR #692, opened 2018-01-08
Adds new static initializers to geojson collection objects

### LINKED ISSUE #691: Varargs support for FeatureCollection.fromFeatures
being able to pass in a single Feature would be a nice enhancement to the API. 
Could be applicable to other parts in the codebase of geojson.

cammace:
Closes #691

### DIFF
--- a/services-geojson/src/main/java/com/mapbox/geojson/GeometryCollection.java
+++ b/services-geojson/src/main/java/com/mapbox/geojson/GeometryCollection.java
@@ -17,6 +17,7 @@ import com.mapbox.geojson.gson.PointSerializer;
 import com.mapbox.geojson.gson.BoundingBoxSerializer;
 
 import java.io.Serializable;
+import java.util.ArrayList;
 import java.util.List;
 
 /**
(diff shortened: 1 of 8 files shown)


services-geojson/src/main/java/com/mapbox/geojson/GeometryCollection.java
```diff
@@ -99,6 +100,20 @@ public static GeometryCollection fromGeometries(@NonNull List<Geometry> geometri
...
+   *
+   * @param geometry a non-null object of type geometry which makes up this collection
+   * @return a new instance of this class defined by the values passed inside this static factory
+   *   method
+   * @since 3.0.0
+   */
+  public static GeometryCollection fromGeometry(@NonNull Geometry geometry) {
+    List<Geometry> geometries = new ArrayList<>();
+    geometries.add(geometry);
+    return new AutoValue_GeometryCollection(null, geometries);
```
tobrun: Since there is not much logic to these three lines, you can alternatively just do a one liner instead: `return new AutoValue_GeometryCollection(null, new ArrayList<>(){geometry});`

cammace: tried making this change but `new ArrayList<>(){geometry}` doesn't work. The closest thing I got to work was `new ArrayList<>(1){Geometry geometry;}` but in that case, I think I'm just creating a new geometry object? Have a better example of this working properly? I can fix it in a future PR. Merging this one now.

cammace: ah makes sense, I'll make sure to fix that up in the next sweep

Issue #695, opened 2018-01-10
SerializedName not working in geocoder CarmenFeature class


cammace:
I'm not sure why this is happening but the `@SerializedName` annotation is supposed to work for both serialization and deserialization but it appears to not work at all. It wasn't noticed until now since we set `FieldNamingPolicy.LOWER_CASE_WITH_UNDERSCORES` on the `gson` object. but if you look at the [`rawCenter`](https://github.com/mapbox/mapbox-java/blob/master/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/CarmenFeature.java#L207) method line, we use `@SerializedName("center")` but the output when serializing the object is `raw_center` still. Any ideas on what the issue could be here @zugaldia @tobrun @osana @danesfeder?

cammace:
This is a result of not using the generated type adapter that the AutoValue Gson uses. Since it's a static method, it can be accessed really easily:

```java
Gson gson = new Gson();
String json = CarmenFeature.typeAdapter(gson).toJson(carmenFeature);
System.out.println(json);
```

This results in using the annotations as expected but requires a bit more work removing and changing the code I didn't know was getting generated. I predict a few of the `TypeAdapterFactory`s can be removed. In 0.7 of A-V-G, there's also a `@GenerateTypeAdapter`  annotation which should clean the code up further.

PR #696, opened 2018-01-16
Add fromJson method to DirectionsRoute

danesfeder:
For converting a JSON `String` representing a `DirectionsRoute` to an actual `DirectionsRoute` object


cammace:
While this works, let's wait and see what https://github.com/mapbox/mapbox-java/issues/695 pans out to be, probably won't need to register as many typeAdapters.

PR #707, opened 2018-01-24
Adding BSD-2 license to repo

langsmith:
This repo needs a license. @kathleenlu09 confirmed and said it needs a BSD-2 license. This pr adds that.

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,0 +1,11 @@
+BSD 2-Clause License
+
+Copyright (c) 2018, Mapbox All rights reserved.
+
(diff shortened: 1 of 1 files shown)


cammace:
I added back the old LICENCE we were using in 2.x but it is MIT, this one looks different. if @kathleenlu09 confirmed it, we can merge this one though and close my PR.

kathleenlu09:
MIT would be fine too, @cammace if that's what we were using before. Not sure how it got lost 🤷‍♀️

langsmith:
Didn't even see your pr from a couple of days ago @cammace 🤦‍♂️ 

I'll close this one and just go with yours.

PR #719, opened 2018-02-12
Moved MathUtils from mapbox-java2.x telemetry to mapbox-java3.0 servi…

osana:
Add MathUtils to services-core.
It used to be part of telemetry

### DIFF
--- a/services-core/src/main/java/com/mapbox/core/utils/MathUtils.java
+++ b/services-core/src/main/java/com/mapbox/core/utils/MathUtils.java
@@ -1,0 +1,48 @@
+package com.mapbox.core.utils;
+
+public class MathUtils {
+
+  /**
+   * Test a value in specified range, returning minimum if it's below, and maximum if it's above
+   *
+   * @param value Value to test
+   * @param min   Minimum value of range
+   * @param max   Maximum value of range
+   * @return value if it's between min and max, min if it's below, max if it's above
+   */
+  public static double clamp(double value, double min, double max) {
+    return Math.max(min, Math.min(max, value));
+  }
+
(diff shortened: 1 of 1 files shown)


services-core/src/main/java/com/mapbox/core/utils/MathUtils.java
```diff
@@ -0,0 +1,78 @@
...
+    return secondMod + min;
+  }
+
+  /**
+   * Convert bearing from core to match Android SDK value.
+   *
+   * @param nativeBearing bearing value coming from core
+   * @return bearing in degrees starting from 0 rotating clockwise
+   */
+  public static double convertNativeBearing(double nativeBearing) {
```
Guardiola31337: Not used in `MapboxGLAndroidSDK` so no need to expose it.
We should remove it.

services-core/src/main/java/com/mapbox/core/utils/MathUtils.java
```diff
@@ -0,0 +1,78 @@
...
+  }
+
+  /**
+   * Returns the smallest angle between two angles.
+   *
+   * @param alpha First angle in degrees
+   * @param beta  Second angle in degrees
+   * @return Smallest angle between two angles.
+   */
+  public static double differenceBetweenAngles(double alpha, double beta) {
```
Guardiola31337: Same here ☝️

Guardiola31337:
`MathUtils` will remain as part of the Maps SDK 👀 https://github.com/mapbox/mapbox-gl-native/issues/11182#issuecomment-365255109

Closing here without merging.

Issue #710, opened 2018-02-02
Low level GeoJson implementation


tobrun:
This is a request for either a rewrite or alternative implementation of the current GeoJson module found [here](https://github.com/mapbox/mapbox-java/blob/master/services-geojson). We need a high performance implementation that has a minimal memory footprint. The reasoning  behind that is that geojson models should be scalable (eg. downstream we query the map for 1000 of features, the path currently taken to create those and the intermediate objects created for doing that isn't optimal atm).

Couple of thoughts to improve this:
 - use primitive instead of objects where possible
 - no need for AutoValue, implementation is based on a specification that doesn't change (often)
 - if geojson models are immutable, use arrays instead of List
 - limit method invocations & don't allocate memory if you can avoid it
 - don't over abstract concepts if they can be expressed more simple 

More information on performance tips [here](https://developer.android.com/training/articles/perf-tips.html)
RFC of GeoJSON can be found [here](https://tools.ietf.org/html/rfc7946).

cc @mapbox/android

zugaldia:
Thank you @tobrun for framing the issue and proposing some solutions. By looking at your list, looks like at list the following items would change the public API of the library:

> use primitive instead of objects where possible
> no need for AutoValue, implementation is based on a specification that doesn't change (often)
> if geojson models are immutable, use arrays instead of List

Am I reading it right? If so, we need to 1) either wait until the next MAS SEMVER release (4.x) or 2) develop this new library as a new product with a different namespace.

tobrun:
Notes in OP were mostly about the internal management so probably we can keep the same public api though this is something that needs to be determined when scoping out. As first steps I would recommend writing up a test suite to tracks memory allocation / cpu usage for 1 feature up to a million features and compare that with the 2.2.9 version. After, we can scoop out what the impact would  be to bring the notes from OP into play and use that same test suite as reference during implementation.

osana:
I did a bit of testing on how much memory Point takes in 2.x.x based code vs 3.0 (with AutoValue).

osana:
Also tried comparing loading of sample json with FeatureCollection: 

{
  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "geometry": {
        "type": "Point",
        "coordinates": [
          102,
          0.5
        ]
      },
      "properties": {
        "prop0": "value0"
      }
    },
    {
      "type": "Feature",
      "geometry": {
        "type": "LineString",
        "coordinates": [
          [
            102,
[...]

PR #722, opened 2018-02-21
Added back fromLngLats() that take double arrays for coordinates

osana:
This is partially addressing https://github.com/mapbox/mapbox-java/issues/710


osana:
@cammace  @tobrun  
The real test for this will be when I integrate into sdk. 
Although I can add some tests as well.

Create Polygon etc. using different methods and then compare

PR #727, opened 2018-02-27
moved toHumanReadableAscii method into library

cammace:
In OkHttp 3.10.0, `toHumanReadableAscii` util method has been moved to `OkHttpURLConnection` and marked as private. Since we previously relied on this method, I've ported over the same method into our utils class.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -30,9 +30,9 @@ jobs:
           paths:
             - ~/.gradle
           key: jars-{{ checksum "mapbox/build.gradle" }}-{{ checksum  "mapbox/app/build.gradle" }}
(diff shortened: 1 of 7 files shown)


cammace:
commit https://github.com/mapbox/mapbox-java/pull/727/commits/b1f4a950e3e611bb3fe1768a53b8bfd54330f7d2 only disables checkstyle, not all of circle ci.

cammace:
The UI tests are failing since we also are using the same internal method in the map SDK, bumping OkHttp back down to 3.9 confirms this. Going ahead and merging to get the fix in the map SDK.

PR #737, opened 2018-03-05
changed compile into implemenation

### LINKED ISSUE #717: Leaking some unnecessary dependencies to the end users
When solving https://github.com/mapbox/mapbox-gl-native/issues/11125 found that we're leaking some unnecessary dependencies to the end users 👀

**5.4.0 - MAS 2.x - Telem 2.2.9**

```
\--- com.mapbox.mapboxsdk:mapbox-android-sdk:5.4.0
     +--- com.android.support:support-annotations:25.4.0
     +--- com.android.support:support-fragment:25.4.0 (*)
     +--- com.jakewharton.timber:timber:4.5.1
     +--- com.squareup.okhttp3:okhttp:3.9.1
     |    \--- com.squareup.okio:okio:1.13.0
     +--- com.mapbox.mapboxsdk:mapbox-java-geojson:2.2.9
     |    +--- com.mapbox.mapboxsdk:mapbox-java-core:2.2.9
     |    \--- com.google.code.gson:gson:2.8.0
     \--- com.mapbox.mapboxsdk:mapbox-android-telemetry:2.2.9
          \--- com.squareup.okhttp3:okhttp:3.6.0 -> 3.9.1 (*)
```

**6.0.0-SNAPSHOT - MAS 3.0.0 - Telem 2.2.9**

```
\--- com.mapbox.mapboxsdk:mapbox-android-sdk:6.0.0-20180209.125103-139
     +--- com.mapbox.mapboxsdk:mapbox-android-telemetry:2.2.9
     |    +--- com.android.support:appcompat-v7:25.4.0 (*)
     |    \--- com.squareup.okhttp3:okhttp:3.6.0 -> 3.9.1
     |         \--- com.squareup.okio:okio:1.13.0
     +--- com.mapbox.mapboxsdk:mapbox-sdk-geojson:3.0.0-beta.2
     |    +--- com.mapbox.mapboxsdk:mapbox-sdk-core:3.0.0-beta.2
     |    |    +--- com.google.code.gson:gson:2.8.2
     |    |    +--- com.squareup.retrofit2:retrofit:2.3.0
     |    |    |    \--- com.squareup.okhttp3:okhttp:3.8.0 -> 3.9.1 (*)
     |    |    +--- com.squareup.retrofit2:converter-gson:2.3.0
     |    |    |    +--- com.squareup.retrofit2:retrofit:2.3.0 (*)
     |    |    |    \--- com.google.code.gson:gson:2.7 -> 2.8.2
     |    |    \--- com.squareup.okhttp3:logging-interceptor:3.9.1
     |    |         \--- com.squareup.okhttp3:okhttp:3.9.1 (*)
     |    \--- com.google.code.gson:gson:2.8.2
     +--- com.android.support:support-annotations:25.4.0
     +--- com.android.support:support-fragment:25.4.0 (*)
     +--- com.jakewharton.timber:timber:4.5.1
     \--- com.squareup.okhttp3:okhttp:3.9.1 (*)
```

We should take advantage of the new Gradle keywords `api`/`implementation` to only expose the libraries strictly necessary. Right now all the dependencies are declared using the deprecated `compile` keyword which translates to `api` and makes that dependency to leak (e.g. https://github.com/mapbox/mapbox-java/blob/master/services-core/build.gradle#L27).
As you can see when using MAS 3.0 we're exposing `com.squareup.retrofit2:retrofit`, `com.squareup.retrofit2:converter-gson` and `com.squareup.okhttp3:logging-interceptor` dependencies which wasn't the case in MAS 2.x

cc @cammace @osana @tobrun @LukasPaczos

osana:
closes #717

### DIFF
--- a/samples/build.gradle
+++ b/samples/build.gradle
@@ -5,13 +5,11 @@ sourceCompatibility = JavaVersion.VERSION_1_8
 targetCompatibility = JavaVersion.VERSION_1_8
 
 dependencies {
-  compile project(":services-core")
-  compile project(":services-directions")
-  compile project(":services-geocoding")
-  compile project(":services-optimization")
-  compile project(":services-geojson")
-  compile project(":services-matching")
-  compile project(":services-staticmap")
+  implementation project(":services-geocoding")
+  implementation project(":services-optimization")
+  implementation project(":services-geojson")
+  implementation project(":services-matching")
+  implementation project(":services-staticmap")
 }
 
 buildConfig {
--- a/services-core/build.gradle
+++ b/services-core/build.gradle
@@ -21,17 +21,17 @@ static def getGitRevision() {
 dependencies {
 
   // Gson
-  compile dependenciesList.gson
+  api dependenciesList.gson
 
   // Retrofit
-  compile dependenciesList.retrofit
-  compile dependenciesList.retrofit2Gson
+  api dependenciesList.retrofit
+  api dependenciesList.retrofit2Gson
 
   // Annotations
   compileOnly dependenciesList.supportAnnotation
 
   // OkHttp
-  compile dependenciesList.okhttp3Logging
+  api dependenciesList.okhttp3Logging
 
   // AutoValue
   compileOnly dependenciesList.autoValue
--- a/services-directions/build.gradle
+++ b/services-directions/build.gradle
@@ -1,27 +1,19 @@
 apply plugin: 'java-library'
 
 dependencies {
-
-  compile project(":services-core")
-  compile project(":services-geojson")
-
-  // Retrofit
-  compile dependenciesList.retrofit
-  compile dependenciesList.retrofit2Gson
+  api project(":services-core")
+  api project(":services-geojson")
 
   // Annotations
   compileOnly dependenciesList.supportAnnotation
-
-  // OkHttp
-  compile dependenciesList.okhttp3Logging
 
   // AutoValue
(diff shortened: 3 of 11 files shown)


tobrun:
A little bit of context around this. With the newer gradle version the old `compile` was deprecated and replaced with: 
 - implementation
 - api

The former doesn't expose dependencies to the consumer apps while the api does. For most cases we actually want to expose the dependencies and we should rely on api for that. For dependencies as support android versions we actually rather not expose them if not needed. More context in https://developer.android.com/studio/build/gradle-plugin-3-0-0-migration.html#new_configurations.

samples/build.gradle
```diff
@@ -5,13 +5,13 @@ sourceCompatibility = JavaVersion.VERSION_1_8
...
-  compile project(":services-geojson")
-  compile project(":services-matching")
-  compile project(":services-staticmap")
+  implementation project(":services-geocoding")
+  implementation project(":services-optimization")
+  implementation project(":services-geojson")
+  implementation project(":services-matching")
+  implementation project(":services-staticmap")
+
+  implementation dependenciesList.retrofit
```
cammace: Why is this needed for the sample project, doesn't one of our service libraries carry retrofit over?

osana: @tobrun @cammace I ran into this this as well when I was doing initial changes and had to include `implementation dependenciesList.retrofit` but then eventually was able to build without it. I will give a try.

PR #738, opened 2018-03-06
Turf methods name changes

### LINKED ISSUE #723: Synchronize with renamed Turf.js functions
[This document](https://github.com/mapbox/mapbox-java/blob/master/docs/turf-port.md) listing Turf.js function names was last updated in October. Since then, the JavaScript version of Turf has renamed some modules. Here’s a non-exhaustive list off the top of my head:

* [ ] `turf-point-on-line` → `turf-nearest-point-on-line`
* [ ] `turf-line-distance` → `turf-length`

This document should be updated. Beyond that, should the corresponding functions in this Java port be renamed as well?

/cc @cammace @zugaldia

osana:
closes #723

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfClassification.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfClassification.java
@@ -28,7 +28,7 @@ public class TurfClassification {
    * @since 3.0.0
    */
   @NonNull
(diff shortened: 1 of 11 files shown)


osana:
Updated https://github.com/mapbox/mapbox-java/blob/master/docs/turf-port.md to reflect the name changes

cammace:
cc @danesfeder who will have a few method name changes to make in the nav SDK

PR #739, opened 2018-03-07
MapMatchingMatching to DirectionRoute

### LINKED ISSUE #705: Util to convert MapMatchingMatching object to DirectionsRoute
if a user wants to convert a MapMatchingMatching object to a DirectionsRoute, they have to manually construct the directionsRoute using it's builder. It would be nice if MapMatchingMatching offered a `toDirectionsRoute()` api.

osana:
closes #705


services-matching/src/main/java/com/mapbox/api/matching/v5/models/MapMatchingMatching.java
```diff
@@ -101,6 +102,40 @@ public static Builder builder() {
     return new AutoValue_MapMatchingMatching.GsonTypeAdapter(gson);
   }
 
+
+  public DirectionsRoute toDirectionRoute() {
+
+//    RouteOptions routeOptions = RouteOptions.builder()
```
cammace: lets remove these commented out lines of code?

PR #751, opened 2018-03-14
release 3.0.0-beta.4

osana:


circle.yml
```diff
@@ -73,6 +73,6 @@ jobs:
       - deploy:
           name: Publish Java Services To Maven Central
           command: |
-            if [ "${CIRCLE_BRANCH}" == "master" ]; then
+            if [ "${CIRCLE_BRANCH}" == "release3.0-beta.4" ]; then
```
cammace: need to change back to master

services-core/gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=3.0.0-SNAPSHOT
+VERSION_NAME=3.0.0-beta.4
```
cammace: all of these need to be changed back to `3.0.0-SNAPSHOT`

PR #762, opened 2018-03-21
Add Abbreviations and type to BannerComponent

danesfeder:
Added recently for banner instruction work to be consumed by the navigation SDK.

cc @devotaaabel @Guardiola31337


danesfeder:
💯

PR #767, opened 2018-03-25
Issue 653:Missing java docs

mghildiy:
Missing elements added to javadocs.


mghildiy:
Why is this build going on for so long?

services-geojson/src/main/java/com/mapbox/geojson/utils/PolylineUtils.java
```diff
@@ -210,6 +214,10 @@ private static double getSqDist(Point p1, Point p2) {
 
   /**
    * Square distance from a point to a segment.
+   *
+   * @param point {@link Point} whose distance from segment needs to be determined
+   * @param p1,p2 points defining the segment
```
cammace: Looking through this PR again, can we split up these javadoc params into two different tags? Each for p1 and p2. In fact, following our checkstyle rules, these params should be renamed to `pointOne` and `pointTwo` at the very least (or even a more descriptive name if possible).

osana:
I am pulling this change into #786

osana:
Thank you for contributing to mapbox-java.
In the future please make sure you run `checkstyle`
you can do it on command line:
```./gradlew checkstyleMain```

PR #768, opened 2018-03-25
Implemented Comparable in BannerComponents

devotaaabel:
Implemented Comparable in BannerComponents so that they can be sorted/compared by abbreviation priority.

### DIFF
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
@@ -19,7 +19,7 @@ import java.io.Serializable;
  * @since 3.0.0
  */
 @AutoValue
-public abstract class BannerComponents implements Serializable {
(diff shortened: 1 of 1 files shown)


services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
```diff
@@ -112,6 +112,19 @@ public static Builder builder() {
...
+   * Allows ability to sort/compare by abbreviation priority.
+   *
+   * @param bannerComponents to compare to
+   * @return a negative integer, zero, or a positive integer as this object is less than, equal to,
+   * or greater than the specified object.
+   * @since 3.0.0
+   */
+  @Override
+  public int compareTo(BannerComponents bannerComponents) {
+    return this.abbreviationPriority().compareTo(bannerComponents.abbreviationPriority());
```
osana: Looks like abbreviationPriority is Nullable, so null check should be there

devotaaabel:
@osana ready to re-review

PR #791, opened 2018-04-13
adding dependency graph

osana:


osana:
dependency-graph-mapbox-libraries

osana:
I am not sure why services-*  appear twice on the graph above.

Makefile
```diff
@@ -41,6 +41,9 @@ publish-local:
     export IS_LOCAL_DEVELOPMENT=true; ./gradlew :services:uploadArchives ; \
     export IS_LOCAL_DEVELOPMENT=true; ./gradlew :services-turf:uploadArchives ; \
 
+graphs:
+	./gradlew generateDependencyGraphMapboxLibraries
```
Guardiola31337: I'd include explicitly each module in which we're generating/including the dependency-graph plugin. It's clearer IMO.

```
graphs:
	./gradlew :services-core:generateDependencyGraphMapboxLibraries
	./gradlew :services-geojson:generateDependencyGraphMapboxLibraries
	./gradlew :services:generateDependencyGraphMapboxLibraries
	./gradlew :services-turf:generateDependencyGraphMapboxLibraries
```

gradle/dependencies.gradle
```diff
@@ -14,14 +14,15 @@ ext {
   ]
 
   pluginVersion = [
-      checkstyle : '7.8.2',
+      checkstyle : '8.2',
```
Guardiola31337: These bumps 👇 are unrelated to OP.

build.gradle
```diff
@@ -68,3 +68,4 @@ task clean(type: Delete) {
 }
 
 apply from: "${rootDir}/gradle/sonarqube.gradle"
+apply from: "${rootDir}/gradle/dependencies-graph.gradle"
```
Guardiola31337: This is not needed. This is why we were getting this https://github.com/mapbox/mapbox-java/pull/791#issuecomment-381215309 weird graph.

Makefile
osana: @Guardiola31337 I removed 
```apply from: "${rootDir}/gradle/dependencies-graph.gradle"```
from the top level `build.gradle` and now I do not have top level graph. 
Probably I should add one for samples, otherwise the graphs are not very meaningful.

Comment (Guardiola31337) [reply Makefile]: @osana 
> I removed
apply from: `"${rootDir}/gradle/dependencies-graph.gradle"`
from the top level `build.gradle` and now I do not have top level graph.

Yeah, that's correct. We're only interested in the library graphs (for now). The idea is to know what Mapbox dependencies include each library module. Samples wouldn't give us any extra value. Does that make sense?

In any case, we were getting an incorrect graph as commented in https://github.com/mapbox/mapbox-java/pull/791#discussion_r182170428 so another good reason to not include it.

**Comment by osana** [comment]
To follow up on @Guardiola31337 suggestion 
I opened a ticket that graph does not include sourceSet dependencies:
https://github.com/vanniktech/gradle-dependency-graph-generator-plugin/issues/41

File: Makefile [hunk 4]
```diff
@@ -131,3 +137,8 @@ optimization-fixtures:
 
 	curl "https://api.mapbox.com/optimized-trips/v1/mapbox/driving/13.388860,52.517037;13.397634,52.529407;13.428555,52.523219;13.418555,52.523215?roundtrip=true&distributions=3,1&access_token=$(MAPBOX_ACCESS_TOKEN)" \
 		-o services-optimization/src/test/resources/optimized_trip_distributions.json
+
+clean:
+	./gradlew clean
+	rm -f */dependency-graph-mapbox-libraries.png
```
Guardiola31337: This is not necessary if we bump the version to `0.4.0` because the graphs are included under the `build` folder 👀 https://github.com/vanniktech/gradle-dependency-graph-generator-plugin/issues/41#issuecomment-382651675
> the graph is located under `build/reports/dependency-graph/`

gradle/dependencies.gradle
```diff
@@ -21,7 +21,9 @@ ext {
       coveralls  : '2.8.1',
       spotbugs   : '1.3',
       sonarqube  : '2.5',
-      buildConfig: '1.1.8'
+      buildConfig: '1.1.8',
+      dependencyGraph: '0.3.0'
```
Guardiola31337: There's a new version available 👀 `0.4.0` https://github.com/vanniktech/gradle-dependency-graph-generator-plugin/releases/tag/0.4.0 We should bump it and use the most up to date.

gradle/dependencies-graph.gradle
```diff
@@ -0,0 +1,26 @@
...
+
+import com.vanniktech.dependency.graph.generator.DependencyGraphGeneratorPlugin
+import com.vanniktech.dependency.graph.generator.DependencyGraphGeneratorExtension.Generator
+
+
+plugins.apply(DependencyGraphGeneratorPlugin)
+
+def mapboxGenerator = new Generator(
+        "mapboxLibraries", // Suffix for our Gradle task.
+        "", // Root suffix that we don't want in this case.
```
Guardiola31337: As @vanniktech suggested here https://github.com/vanniktech/gradle-dependency-graph-generator-plugin/issues/41#issuecomment-382651675 it's not necessary to include the root suffix if we're not going to use it.

Guardiola31337:
Re:
> In any case, it seems that the plugin is not taken into account when the dependencies are included via `sourceSets` 👀 https://github.com/mapbox/mapbox-java/blob/49a07650f81c5b8982501b9bedb553c58493b44e/services/build.gradle#L3-L8

Could be implemented using directly Graphviz somehow? Do you know @zugaldia?

Makefile
vanniktech: Yup true.

gradle/dependencies-graph.gradle
vanniktech: Yes and it's also no longer present in 0.4.0 which means you have to remove it anyways.

Guardiola31337: Awesome @vanniktech thanks for confirming!

.gitignore
```diff
@@ -23,3 +23,6 @@ obj
 
 #MacOS
 .DS_Store
+
+dependency-graph-mapbox-libraries.png
```
vanniktech: This is no longer needed now since it's generated into the build folder.

gradle/dependencies-graph.gradle
```diff
@@ -0,0 +1,25 @@
...
+import com.vanniktech.dependency.graph.generator.DependencyGraphGeneratorPlugin
+import com.vanniktech.dependency.graph.generator.DependencyGraphGeneratorExtension.Generator
+
+
+plugins.apply(DependencyGraphGeneratorPlugin)
+
+def mapboxGenerator = new Generator(
+        "mapboxLibraries", // Suffix for our Gradle task.
+        { dependency -> dependency.getModuleGroup().startsWith("com.mapbox") }, // Only want Mapbox libs.
+        { dependency -> true }, // Include transitive dependencies.
```
vanniktech: Technically that's the default and does not need to be passed. I've done so in the example since I need the next parameter to configure and in Groovy I don't think you can use named parameters with default values.

gradle/dependencies-graph.gradle
```diff
@@ -0,0 +1,25 @@
...
+
+    dependencies {
+        classpath pluginDependencies.dependencyGraph
+    }
+}
+
+import com.vanniktech.dependency.graph.generator.DependencyGraphGeneratorPlugin
+import com.vanniktech.dependency.graph.generator.DependencyGraphGeneratorExtension.Generator
+
+
```
vanniktech: nit: new line is not necessary

Issue #752, opened 2018-03-15
Method LineString.fromPolyline divides coordinates by 10


vicobz:
#### Expected Behavior

When I call `fromPolyline(geometry, PRECISION_6)` method from `LineString` class in order to parse a geometry String into a list of features to update a GeoJson source, I'm expecting the resulting `LineString` object to render accurately the coordinates held by this geometry.

#### Current Behavior

All coordinates values held by the resulting `LineString` object are divided by 10.
For example, `[2.2862036, 48.8267868]` is resulting in `[0.22862036, 4.88267868]`.

#### Steps to Reproduce

Call the above method `fromPolyline`.

#### Context (Environment)
I'm using `com.mapbox.mapboxsdk:mapbox-android-sdk:5.5.0` version on Android O with OnePlus 5 concurrently with several Mapbox plugins (last releases).

This could be related to #353.

mghildiy:
Can you provide your geometry string?

vicobz:
@mghildiy Not exactly the same as when I wrote the issue although quite equivalent: `qn_iHgp}LzCy@xCsAsC}PoEeD_@{A@uD_@Sg@Je@a@I_@FcAoFyGcCqFgQ{L{CmD`

cammace:
We use this static factory [pretty heavily](https://github.com/mapbox/mapbox-navigation-android/search?utf8=%E2%9C%93&q=linestring.from&type=) in the Mapbox Navigation and haven't noticed any issues with the geometric Points being off. If you are using the latest beta of this library and still having issues, maybe something to do with your Locale?

vicobz:
@cammace I'm using the latest beta, will try to investigate on this when I get some time. About my `Locale`, to what extent can this affect the conversion of polylines ?

osana:
Added tests did not show any problems. Closing

PR #796, opened 2018-04-17
added more tests for PolyLineUtils

osana:
Tests for issue #752 
cc @vicobz


services-geojson/src/test/java/com/mapbox/geojson/utils/PolylineUtilsTest.java
```diff
@@ -32,12 +37,67 @@ public void testDecodePath() {
...
+    List<Point> originalPath = Arrays.asList(
+      Point.fromLngLat(2.2862036, 48.8267868),
+      Point.fromLngLat(2.4, 48.9)
+    );
+    String encoded = encode(originalPath, Constants.PRECISION_6);
+    List<Point> path = LineString.fromPolyline(encoded, Constants.PRECISION_6).coordinates();
+
+    assertEquals(originalPath.size(), path.size());
+    for (int i = 0; i < originalPath.size(); i++) {
+      assertEquals(originalPath.get(i).latitude(), path.get(i).latitude(), 0.000001);
```
cammace: Let's make this Delta a constant since we use it in so many places for these test.

PR #804, opened 2018-04-24
Add voiceLocale to DirectionsRoute and update test fixtures

danesfeder:
Needed for https://github.com/mapbox/mapbox-navigation-android/issues/882


services-directions/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java
```diff
@@ -123,6 +123,17 @@ public static DirectionsRoute fromJson(String json) {
...
+
+  /**
+   * String of the locale used for voice instructions.  Defaults to en, and
+   * can be any accepted instruction language.
+   *
+   * @return String compatible with voice instructions, null otherwise
+   * @since 3.1.0
+   */
+  @Nullable
+  public abstract String voiceLocale();
```
1ec5: I would suggest making this a `Locale` for consistency, but then there’s the whole `iw` = Hebrew thing to contend with… _*hides in bushes*_

danesfeder: But the API returns a `String`?  I think we'd have to convert it after we get the response if we want it to be a `Locale`.  Also, shouldn't the directions API only return `String` language tags that are compatible with API voice?  So this should actually solve the deprecated tag issue.

danesfeder: Better question - is the `language` we set for https://www.mapbox.com/api-documentation/#instructions-languages eventually used for this `voiceLocale`?

1ec5: > But the API returns a `String`?  I think we'd have to convert it after we get the response if we want it to be a `Locale`.

The API of course returns a `String`, since `Locale` has no JSON representation. I was just supposing that having the return value be `Locale` would balance out the use of `Locale` for the language option in the request.

> Also, shouldn't the directions API only return `String` language tags that are compatible with API voice?  So this should actually solve the deprecated tag issue. 

Wouldn’t `Locale`’s choice of ISO 639 codes be an implementation detail to hide from developers who work directly with this library?

> Better question - is the `language` we set for https://www.mapbox.com/api-documentation/#instructions-languages eventually used for this `voiceLocale`?

It influences the `voiceLocale` value, but it may not be identical, because it reflects the actual language used in the instructions. Currently, it may differ from the requested locale for the following reasons:

* The locale code specifies an ISO&nbsp;15924 script code or an ISO&nbps;3166 or UN&nbsp;M.49 region code. In some cases, such as Chinese or Portuguese, the API will respect this code and preserve it in the response. Otherwise, it’ll ignore and strip off the code.
* The locale is unavailable because OSRM Text Instructions lacks translations in that language, in which case the API falls back to `en`.
* The locale is unavailable because the Voice API lacks a voice in that language, in which case the return value will be `null`.

In the future, there may be more sophisticated fallbacks, such as three-letter ISO&nbsp;639-2 codes falling back to two-letter ISO&nbsp;639-1 codes.

These may be great reasons to stick to a string, to ensure high fidelity as the value is relayed from the Directions API to the Voice API. On the other hand, if the Android TTS engine or a third-party TTS engine expects a `Locale`, then it may be more convenient to proactively convert to a `Locale`.

The Directions and Voice APIs are expected to behave similarly with respect to locale codes, subject to availability. The main concern I’d have is whether round-tripping the string through `Locale` and back would lose any information, particularly in the case of Chinese.

PR #818, opened 2018-05-16
Move TurfAssertions.getCoord() to TurfMeta.getCoord() to follow Turf.js

### LINKED ISSUE #816: Move Turf getCoord to correct class
Following the same name scheme, we've been using since the start, we should deprecate the `TurfAssertion.getCoord()` and have it call into a new method in `TurfMeta` called getCoord(). no arithmetic functionality needs changing just needs name updating to keep consistent with Turf.js

`TurfAssertion.getCoord()` -> `TurfMeta.getCoord()`

cc: @osana

osana:
closes #816 

Also updated https://github.com/mapbox/mapbox-java/blob/master/docs/turf-port.md


services-turf/src/main/java/com/mapbox/turf/TurfAssertions.java
```diff
@@ -19,21 +18,6 @@ private TurfAssertions() {
...
 
-  /**
-   * Unwrap a coordinate {@link Point} from a Feature with a Point geometry.
-   *
-   * @param obj any value
-   * @return a coordinate
-   * @see <a href="http://turfjs.org/docs/#getcoord">Turf getCoord documentation</a>
-   * @since 1.2.0
-   */
-  public static Point getCoord(Feature obj) {
```
cammace: This is a semver change, you need to deprecate and point to using the other method instead of deleting it here.

osana: @cammace this exactly what I was not sure about.
This  issue indeed said: deprecate but then 
https://github.com/mapbox/mapbox-java/pull/738
contained name changes and you reviewed it.
Some of those name changes were for API's since 1.2

Should I deprecate those as well instead of changename

cammace: Yes, those need to be fixed also. Sorry for missing that in #738

PR #833, opened 2018-06-07
downgrade to java 7

### LINKED ISSUE #832: Revert support for java 1.7 or figure a way to use java 8 features while maintaining compatibility for Android before N
WHat kind of issue is this?

- [ ] **Questions** are better to ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/mapbox+android). We are actively monitoring the questions there and oftentimes, others have already asked a similar question.
- [ ] **Feature request** should start off by describing the problem the feature will resolve. Please open a GitHub ticket beforehand with information about the feature to spark some discussion about the topic and once you have our support.  
- [ x ] **Bug Report** can be reported using the provided template below:

This issue was already brought up here : https://github.com/mapbox/mapbox-gl-native/issues/11392 and solution was going back to 1.7, this is most likely the PR that caused this issue: https://github.com/mapbox/mapbox-java/pull/770
#### Expected Behavior

Everything compiles while supporting lowest android sdk supported by mapbox: 14.
#### Current Behavior

App won't compile unless i target Android N
```
AGPBI: {"kind":"error","text":"Static interface methods are only supported starting with Android N (--min-api 24): com.mapbox.geojson.Geometry com.mapbox.geojson.Geometry.fromJson(java.lang.String)","sources":[{}],"tool":"D8"}
```
I am using newest build tools com.android.tools.build:gradle:3.1.2
#### Possible Solution

- Revert support for Java8
- figure a way to maintain 1.8 support using
https://developer.android.com/studio/write/java8-support
it might be that unless library is provided as AAR those do not apply - so maybe it would make sense to provide this part of mapbox also as aar dependency for android. I am not sure if that would solve the problem though - someone would need to confirm.

#### Context (Environment)

Build Tools: com.android.tools.build:gradle:3.1.2
MinSdk: 21
mapBoxServicesVersion = '3.2.0'

### LINKED ISSUE #868: Remove Java 8 features from geojson module
Request to remove java 8 from geojson module and as result remove the static method decleration of toJson on the Geometry interface, see [here](https://github.com/mapbox/mapbox-java/blob/master/services-geojson/src/main/java/com/mapbox/geojson/Geometry.java#L26). Since the method isn't public can we patch this up without breaking semver?

cc @osana @zugaldia

osana:
closes #868 
related to #832

java8 was introduced by  #770 #749

### DIFF
--- a/services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteLegTest.java
+++ b/services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteLegTest.java
@@ -71,6 +71,8 @@ public class RouteLegTest extends TestUtils {
       "heavy",
       "heavy");
 
+    List<LegStep> legSteps = new ArrayList<>();
+
     LegAnnotation annotation = LegAnnotation.builder()
       .congestion(new ArrayList<String>())
       .distance(distanceList)
@@ -84,7 +86,7 @@ public class RouteLegTest extends TestUtils {
       .distance(53.4)
       //.weight(14.3)
       .duration(14.3)
-      .steps(new ArrayList<>())
+      .steps(legSteps)
       .summary("")
       .build();
 
--- a/services-geojson/src/main/java/com/mapbox/geojson/CoordinateContainer.java
+++ b/services-geojson/src/main/java/com/mapbox/geojson/CoordinateContainer.java
@@ -8,7 +8,7 @@ package com.mapbox.geojson;
  * @param <T> a generic allowing varying dimensions for each GeoJson geometry
  * @since 3.0.0
  */
-public interface CoordinateContainer<T> extends Geometry {
+public abstract class CoordinateContainer<T> extends Geometry {
 
   /**
    * the coordinates which define the geometry. Typically a list of points but for some geometry
@@ -17,5 +17,5 @@ public interface CoordinateContainer<T> extends Geometry {
    * @return the {@link Point}s which make up the coordinates defining the geometry
    * @since 3.0.0
    */
-  T coordinates();
+  public abstract T coordinates();
 }
--- a/services-geojson/src/main/java/com/mapbox/geojson/GeoJson.java
+++ b/services-geojson/src/main/java/com/mapbox/geojson/GeoJson.java
@@ -7,7 +7,7 @@ package com.mapbox.geojson;
  *
  * @since 1.0.0
  */
-public interface GeoJson {
+public abstract class GeoJson {
 
   /**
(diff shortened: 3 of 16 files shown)


services-geojson/src/main/java/com/mapbox/geojson/CoordinateContainer.java
```diff
@@ -17,5 +17,5 @@
    * @return the {@link Point}s which make up the coordinates defining the geometry
    * @since 3.0.0
    */
-  T coordinates();
+  public abstract T coordinates();
```
tobrun: the previous method was package private, can we retain that with this setup?

services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteLegTest.java
```diff
@@ -84,7 +86,7 @@ public void testToFromJson1() {
       .distance(53.4)
       //.weight(14.3)
       .duration(14.3)
-      .steps(new ArrayList<>())
+      .steps(legSteps)
```
tobrun: these changes seem unrelated but no biggy to keep them around

services-geojson/src/main/java/com/mapbox/geojson/GeoJson.java
```diff
@@ -7,7 +7,7 @@
  *
  * @since 1.0.0
  */
-public interface GeoJson {
+public abstract class GeoJson {
```
tobrun: geojson itself is public and changing it from an interface to an abstract class would be a breaking change as users using it will need to change `implements` to `extends` + if they are already extending an object, they won't be able to make this change due to java only supporting single inheritance.

services-geojson/src/main/java/com/mapbox/geojson/CoordinateContainer.java
osana: This was part of an interface so when implemented it would have to be public

services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteLegTest.java
osana: For some reason it was giving me compilation error under java 7.
Casting was not possible for some reason.

osana:
I did not have to remove 
https://github.com/mapbox/mapbox-java/blob/master/services-geojson/src/main/java/com/mapbox/geojson/Geometry.java#L26 
as it used to be a default method in the `interface` but now it is a method in an abstract class.

PR #836, opened 2018-06-08
Static image marker fixes

freszu:
New version of Mapbox static map utils 3.x caused a few issues for us that should solve it.
Old working link from 2.x:
```
https://api.mapbox.com/styles/v1/username/styleId/static/pin-l+ff0000(-115.255990,36.263328),pin-l+ff0000(-115.255990,36.263328)/auto/800x300?access_token=token&logo=false
```
New failing link from 3.x:
```
https://api.mapbox.com/styles/v1/username/styleId/static/pin-l-ff0000(-115.255990,36.263328),null/auto/800x300?access_token=token&logo=false
```
PR contains:
- Fix string format for color
- remove useless null check - TextUtils already does that
- Fix wrong link generated when 2 pins are in the same location have same size and color - old behaviour was creating proper link, new one fails silently generating illegal url by placing null as second pin. This was the result of indexOf call which was also removed because its a waste of cpu cycles.

### DIFF
--- a/services-staticmap/src/main/java/com/mapbox/api/staticmap/v1/MapboxStaticMap.java
+++ b/services-staticmap/src/main/java/com/mapbox/api/staticmap/v1/MapboxStaticMap.java
@@ -103,11 +103,11 @@ public abstract class MapboxStaticMap {
 
     List<String> annotations = new ArrayList<>();
     if (staticMarkerAnnotations() != null) {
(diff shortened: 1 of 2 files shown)


osana:
Thank you for your contribution. I will port your PR to be merged

osana:
merged by #839

PR #839, opened 2018-06-14
Static image marker fixes 

osana:
porting external PR #836

cc @freszu


freszu:
What's up with the idea of copying my changes around?

osana:
@freszu  This so that CI test could be automatically run. As you can see one test is failing. I suspect that the test is wrong - will look into this tomorrow.

PR #860, opened 2018-08-03
Support for Tilequery API

zugaldia:
This PR adds support for the Tilequery API. The [Mapbox Tilequery API](https://www.mapbox.com/api-documentation/#tilequery) allows you to retrieve data about specific features from a vector tileset, based on a given latitude and longitude.

The basic wiring is already set up. Before merging this PR, the following components need to be added:
- [x] Tests (fixtures are already present)
- [x] An example of usage
- [x] Javadoc

/cc: @osana


zugaldia:
@osana I've now added tests and Javadoc for all new public methods, however, I'm having issues running the code locally so I'm flying blind. I think there's something not properly set up / needs documentation for how this project has the AutoValue generated classes be recognized by the IDE. I've tried with both Android Studio and IntelliJ IDEA with no luck.

Otherwise, this PR is ready for review. It's light on tests because of ^^^^ - feel free to make any changes directly to this branch. Thanks!

Issue #861, opened 2018-08-06
Release 3.4.0


danesfeder:
### Includes:
- Added toJson and fromJson to Directions models (#854)
- Upgraded dependencies (#853)
- Missing space (#846)
- Add voiceLanguage to MapMatchingMatching (#847)

### Release checklist @danesfeder
- [x] If there's a new SNAPSHOT version, then update this repo's `README.md` file
- [x] Change version name to `3.4.0` in `gradle.properties`
- [x] Publish `3.4.0` artifact to Maven Central
- [x] Change version name back to `3.5.0-SNAPSHOT`
- [x] Update `CHANGELOG.md`
- [x] Tag `v3.4.0` in GitHub
- [x] Update Maps SDK dependency 

### Post-release @langsmith

- [x] Add the Mapbox Java packages' API reference docs to the their respective folders in `/android-docs/api/mapbox-java`
- [x] Update the `constants.js` file in `android-docs` repo so that `compile 'com.mapbox.mapboxsdk:mapbox-sdk-services:x.x.x'` line is updated in [the Gradle section of the MAS overview page](https://www.mapbox.com/android-docs/java-sdk/overview/)
- [x] Update [the Android demo app](https://github.com/mapbox/mapbox-android-demo)
- [x] Update version # in config files in /help @langsmith
- [x] Update version # in Studio Preview
- [x] Update version # in cnplgn

/cc: @mapbox/android

langsmith:
All done on my end @danesfeder :shipit:

danesfeder:
This release is good to go - closing 💯

PR #862, opened 2018-08-06
Release v3.4.0

danesfeder:
Refs #861


osana:
@danesfeder Thank you for running with this. Could you also add:
https://github.com/mapbox/mapbox-java/pull/855 in release notes

PR #872, opened 2018-08-20
No need to have TypeAdapterFactory

osana:
GeoJsonAdapterFactory could be used instead


services-tilequery/src/test/java/com/mapbox/api/tilequery/TilequeryTest.java
```diff
@@ -0,0 +1,82 @@
+package com.mapbox.api.tilequery;
```
zugaldia: The original PR already had a test class in `services-tilequery/src/test/java/com/mapbox/api/tilequery/MapboxTilequeryTest.java`. Could you add your tests to it, instead of creating a new one?

PR #876, opened 2018-08-27
 java 7 downgrade

### LINKED ISSUE #868: Remove Java 8 features from geojson module

osana:
closes #868 

related to #832

java8 was introduced by #770 #749

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -49,8 +49,8 @@ subprojects {
     }
   }
 
-  sourceCompatibility = "1.8"
-  targetCompatibility = "1.8"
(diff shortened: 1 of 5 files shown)


tobrun:
We should only merge this if we are planning to do a major release of geojson

PR #886, opened 2018-09-10
The first and last positions should be equivalent in polygon ring

### LINKED ISSUE #884: TurfTransformation.circle produces invalid GeoJSON
- [x] **Bug Report** can be reported using the provided template below:

#### Expected Behavior
As per [GeoJSON specification (Section 3.1.6)](https://tools.ietf.org/html/rfc7946#section-3.1.6) a valid polygon must have the same first and last point. 
> The first and last positions are equivalent, and they MUST contain
      identical values; their representation SHOULD also be identical.

#### Current Behavior
Current implementation generates circle where first and last point are not the same. This is reflected in tests where [expected circle](https://github.com/mapbox/mapbox-java/blob/master/services-turf/src/test/resources/turf-transformation/circle_out.json) polygon is incorrect.

```sh
$ geojsonhint circle_out.json
  1:1  error  the first and last positions in a LinearRing of coordinates must be the same

✖ 1 error
```

#### Possible Solution/Implementation
Add first point to coordinates list in `circle` method
https://github.com/mapbox/mapbox-java/blob/master/services-turf/src/main/java/com/mapbox/turf/TurfTransformation.java#L65

#### Steps to Reproduce
1. Run [TurfTransformationTest](https://github.com/mapbox/mapbox-java/blob/master/services-turf/src/test/java/com/mapbox/turf/TurfTransformationTest.java)

osana:
ported ouside PR https://github.com/mapbox/mapbox-java/pull/885
closes #884


langsmith:
👍  :shipit:

PR #895, opened 2018-09-26
Route options should be serialized/deserialized using backend names

### LINKED ISSUE #894: RouteOptions#fromJson is not working correctly
[`RouteOptions#fromJson`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java#L293) is not working correctly. I guess because how we're serializing / deserializing some of the fields i.e. when executing the request we're sending the fields correctly (snake case) 👀 https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java#L54 but we're not taking that into account when converting a `RouteOptions` object back from a JSON 👀 

```java
@Test
public void incorrectFromJson() {
  String routeOptionsJson = "{\n" +
          "\t\t\"profile\": \"auto\",\n" +
          "\t\t\"user\": \"mapbox\",\n" +
          "\t\t\"baseUrl\": \"https://api.mapbox.com\",\n" +
          "\t\t\"coordinates\": [{\n" +
          "\t\t\t\"type\": \"Point\",\n" +
          "\t\t\t\"coordinates\": [-3.707788, 40.395039]\n" +
          "\t\t}, {\n" +
          "\t\t\t\"type\": \"Point\",\n" +
          "\t\t\t\"coordinates\": [-3.712179, 40.401819]\n" +
          "\t\t}],\n" +
          "\t\t\"access_token\": \"TOP SECRET!!!\",\n" +
          "\t\t\"geometries\": \"polyline6\",\n" +
          "\t\t\"overview\": \"full\",\n" +
          "\t\t\"steps\": true,\n" +
          "\t\t\"bearings\": \";\",\n" +
          "\t\t\"continue_straight\": true,\n" +
          "\t\t\"annotations\": \"congestion,distance\",\n" +
          "\t\t\"language\": \"en\",\n" +
          "\t\t\"roundabout_exits\": true,\n" +
          "\t\t\"voice_instructions\": true,\n" +
          "\t\t\"banner_instructions\": true,\n" +
          "\t\t\"voice_units\": \"imperial\"\n" +
          "\t}";
  RouteOptions incorrect = RouteOptions.fromJson(routeOptionsJson);
  assertEquals("TOP SECRET!!!", incorrect.accessToken());
}
```

I'm running into a NPE because the access token doesn't get serialized 👀 

```
java.lang.NullPointerException: Null accessToken
```

Wondering if the same wrong behavior is happening for other objects across the different services. We should double check.

cc @osana

osana:
closes #894 

RouteOptions.test_toJson test is still failing com.google.gson.stream.MalformedJsonException

 @Guardiola31337  There must be some typo... that I cannot see. Could you 👀 .

### DIFF
--- a/services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
+++ b/services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
@@ -14,6 +14,7 @@ import org.junit.Test;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Locale;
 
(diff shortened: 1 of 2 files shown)


services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -214,4 +215,106 @@ public void toJson_fromJson() throws Exception {
...
+      .annotations("congestion,distance")
+      .language("en")
+      .roundaboutExits(true)
+      .voiceInstructions(true)
+      .bannerInstructions(true)
+      .voiceUnits("imperial")
+      .requestUuid("uuid1")
+      .build();
+
+    String jsonString = routeOptions.toString();
```
Guardiola31337: @osana 
Found it 🎉 

Should be `routeOptions.toJson()` instead of `toString()`.

In any case, we should be careful because we can't compare json strings as we're doing here (`compareJson` below), because the order in which the json fields are represented is taken into account in the assertion which doesn't mean that the JSON is incorrect.

I've also noticed a discrepancy in the `coordinates` which we should fix 👀 

Expected
```
"coordinates":[-3.707788,40.395039]},{"type":"Point","coordinates":[-3.712179,40.401819]}]
```

Actual
```
"coordinates":[[-3.707788,40.395039],[-3.712179,40.401819]]
```

services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -214,4 +215,94 @@ public void toJson_fromJson() throws Exception {
 
     assertEquals(routeOptions, routeOptionsFromJson);
   }
+
+  @Test
+  public void test_fromJson() {
```
Guardiola31337: NIT `test` from `test_fromJson` is redundant IMO

services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -214,4 +215,94 @@ public void toJson_fromJson() throws Exception {
...
+      "\"language\": \"en\"," +
+      "\"roundabout_exits\": true," +
+      "\"voice_instructions\": true," +
+      "\"banner_instructions\": true," +
+      "\"voice_units\": \"imperial\"," +
+      "\"uuid\": \"uuid1\"" +
+      "}";
+
+    RouteOptions routeOptions = RouteOptions.fromJson(jsonString);
+    assertEquals("auto", routeOptions.profile());
```
Guardiola31337: Let's add a new line here to remark the [triple A](https://speakerdeck.com/guardiola31337/elegant-unit-testing-mobilization-2016?slide=40) "rule".

services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -214,4 +215,94 @@ public void toJson_fromJson() throws Exception {
...
+    assertEquals(true, routeOptions.roundaboutExits());
+    assertEquals(true, routeOptions.voiceInstructions());
+    assertEquals(true, routeOptions.bannerInstructions());
+    assertEquals("imperial", routeOptions.voiceUnits());
+    assertEquals("uuid1", routeOptions.requestUuid());
+  }
+
+  @Test
+  public void test_toJson() {
+
```
Guardiola31337: NIT unnecessary empty line

services-directions/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -214,4 +215,94 @@ public void toJson_fromJson() throws Exception {
...
+      "\"bearings\": \";\"," +
+      "\"continue_straight\": true," +
+      "\"annotations\": \"congestion,distance\"," +
+      "\"language\": \"en\"," +
+      "\"roundabout_exits\": true," +
+      "\"voice_instructions\": true," +
+      "\"banner_instructions\": true," +
+      "\"voice_units\": \"imperial\"," +
+      "\"uuid\": \"uuid1\"}";
+
```
Guardiola31337: NIT unnecessary empty line

osana:
@Guardiola31337 I believe all NITs are fixe now.

PR #898, opened 2018-10-01
Added StepManeuver.Type for ease of use

### LINKED ISSUE #897: StepManeuver#type() enum with @StringDef
Suggestion to improve `StepManeuver#type()` typing by using `@StringDef`. 
Example of a StringDef can be found [here](https://github.com/mapbox/mapbox-gl-native/blob/master/platform/android/MapboxGLAndroidSDK/src/main/java/com/mapbox/mapboxsdk/constants/Style.java).

```
  /**
   * <li>turn - a basic turn into direction of the modifier</li>
   * <li>new name - the road name changes (after a mandatory turn)</li>
   * <li>depart - indicates departure from a leg</li>
   * <li>arrive - indicates arrival to a destination of a leg</li>
   * <li>merge - merge onto a street</li>
   * <li>on ramp - take a ramp to enter a highway</li>
   * <li>off ramp - take a ramp to exit a highway</li>
   * <li>fork - take the left/right side of a fork</li>
   * <li>end of road - road ends in a T intersection</li>
   * <li>continue - continue on a street after a turn</li>
   * <li>roundabout - traverse roundabout, has additional property {@link #exit()} in RouteStep
   * containing the exit number. The modifier specifies the direction of entering the roundabout.
   * </li>
   * <li>rotary - a traffic circle. While very similar to a larger version of a roundabout, it does
   * not necessarily follow roundabout rules for right of way. It can offer
   * {@link LegStep#rotaryName()} and/or {@link LegStep#rotaryPronunciation()} parameters in
   * addition to the {@link #exit()} property.</li>
   * <li>roundabout turn - small roundabout that is treated as an intersection</li>
   * <li>notification - change of driving conditions, e.g. change of mode from driving to ferry</li>
   */
```

osana:
closes #897

### DIFF
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerText.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerText.java
@@ -52,34 +52,14 @@ public abstract class BannerText extends DirectionsJsonObject {
   public abstract List<BannerComponents> components();
 
   /**
-   * This indicates the type of maneuver. It can be any of these listed:
-   * <br>
-   * <ul>
-   * <li>turn - a basic turn into direction of the modifier</li>
-   * <li>new name - the road name changes (after a mandatory turn)</li>
-   * <li>depart - indicates departure from a leg</li>
-   * <li>arrive - indicates arrival to a destination of a leg</li>
-   * <li>merge - merge onto a street</li>
-   * <li>on ramp - take a ramp to enter a highway</li>
-   * <li>off ramp - take a ramp to exit a highway</li>
-   * <li>fork - take the left/right side of a fork</li>
-   * <li>end of road - road ends in a T intersection</li>
-   * <li>continue - continue on a street after a turn</li>
-   * <li>roundabout - traverse roundabout, has additional property exit in RouteStep
-   * containing the exit number. The modifier specifies the direction of entering the roundabout.
-   * </li>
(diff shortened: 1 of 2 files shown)


osana:
@tobrun thanks
@danesfeder @Guardiola31337 could one of you have a glance at this?

Guardiola31337:
@osana 
Sounds good to me. This way we will be able to remove the same `@StringDef` implemented in `NavigationConstants` 👀 https://github.com/mapbox/mapbox-navigation-android/blob/392969b63b7c84dcf7f110b2ed4dc17093d189a9/libandroid-navigation/src/main/java/com/mapbox/services/android/navigation/v5/navigation/NavigationConstants.java#L245-L279 and use this one - noticed that we were missing `EXIT_ROUNDABOUT` from our side.

I believe we can do the same for `BannerText`'s type 👀 https://github.com/mapbox/mapbox-java/blob/402d63e2bee5196ac72be82b8521dc7919bdf305/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerText.java#L54-L83 
Also if I'm not mistaken these are the same options as in `StepManeuver`. If so, what about extracting the `@StringDef` somewhere else and use it across both types? Does that make sense?

osana:
@Guardiola31337 Thank you for looking into this. Thank you for pointing out that those are used in `BannerText` as well. I agree that it is best to move them into separate class, for example : StepManeuverType?

PR #903, opened 2018-10-11
trying out shifting coordinates

osana:


zugaldia:
I like how this approach helps decouple the logic downstream. Capturing chat, let's include performance tests to understand any hits coming from the approach in this PR (introducing a default operation in all cases) vs only applying the shifting conditionally vs no changes.

services-geojson/src/main/java/com/mapbox/geojson/shifter/CoordinateShifter.java
```diff
@@ -0,0 +1,9 @@
+package com.mapbox.geojson.shifter;
+
+public interface CoordinateShifter {
+
+  double shiftLat(double lat);
```
lloydsheng: I'd suggest the API can be `Coordinate shift(Coordinate lon)`,  case the latitude and longitude used to calculate together. https://github.com/mapbox/mapbox-plugin-ios-china/blob/afbaf3fac65e7ebc243e4a6c2e4a8b85fe858cca/ChinaPlugin/China/Algorithm/gcj.c#L234-L235

osana: @lloydsheng Thank you for pointing this out. This will be a pain because we are trying to do this in the constructor.

osana:
@tobrun @LukasPaczos  Could you have a look
I still need to work on toJson/fromJson adjustments.

langsmith:
Update after chatting with @tobrun about this:

- Getting access to @osana’s work in China plugin

  - Merge @osana’s changes to `master`
  - In the China Plugin project, I will use `6.6.0` of the Maps SDK but and use `exclude group` as seen in the code example in https://stackoverflow.com/a/36103326/6358488 . I will use `exclude group` to block usage of the Maps SDK’s regular Java SDK version. I will also add a new dependency line that adds the Java SDK `SNAPSHOT`, which should include @osana’s shifting work by then.

- Regarding testing:

  - Timestamp comparison via simple logcat logging (@tobrun mentioned there might be some sort of framework for this)
  - Profiling memory through Android Studio. GC for use in Android Studio. Run tests and see how much memory consumed. Compare to baseline non-shifting.
  - Try to see if there's a way to test CPU usage
  - Seeing the performance of C++  JNI bridge for shifting will be interesting

services-geojson/src/main/java/com/mapbox/geojson/shifter/CoordinateShifter.java
```diff
@@ -0,0 +1,49 @@
...
+/**
+ * ShifterManager allows to move all points according to some pluggable algorithm.
+ * Once set it will be applied to all Point object created through this method.
+ *
+ * @since 4.1.0
+ */
+public interface CoordinateShifter {
+
+  /**
+   * Shift coordinagte values according to its algorithm.
```
langsmith:
```suggestion
   * Shift coordinate values according to its algorithm.
```

services-geojson/src/main/java/com/mapbox/geojson/shifter/CoordinateShifter.java
```diff
@@ -0,0 +1,49 @@
...
+   *
+   * @param lon unshifted longitude
+   * @param lat unshifted latitude
+   * @return shifted longitude, shifted latitude in the form of List of Double
+   * @since 4.1.0
+   */
+  List<Double> shiftLonLat(double lon, double lat);
+
+  /**
+   * Shift coordinagte values according to its algorithm.
```
langsmith:
```suggestion
   * Shift coordinate values according to its algorithm.
```

services-geojson/src/main/java/com/mapbox/geojson/shifter/CoordinateShifter.java
```diff
@@ -0,0 +1,49 @@
...
+   * @param lon unshifted longitude
+   * @param lat unshifted latitude
+   * @param altitude  unshifted altitude
+   * @return shifted longitude, shifted latitude, shifted altitude in the form of List of Double
+   * @since 4.1.0
+   */
+  List<Double> shiftLonLatAlt(double lon, double lat, double altitude);
+
+  /**
+   * Unshift coordinagte values according to its algorithm.
```
langsmith:
```suggestion
   * Unshift coordinate values according to its algorithm.
```

services-geojson/src/main/java/com/mapbox/geojson/gson/GeometryDeserializer.java
```diff
@@ -51,7 +51,7 @@ public Geometry deserialize(JsonElement json, Type typeOfT, JsonDeserializationC
 
     try {
       // Use the current context to deserialize it
-      Type classType = Class.forName("com.mapbox.geojson." + geometryType);
+      Type classType = Class.forName("com.mapbox.geojson.AutoValue_" + geometryType);
```
tobrun: is this change required for this PR specifically?

services-geojson/src/main/java/com/mapbox/geojson/gson/PointDeserializer.java
```diff
@@ -43,11 +43,7 @@ public PointDeserializer() {
   public Point deserialize(JsonElement json, Type typeOfT, JsonDeserializationContext context) {
     JsonArray rawCoordinates;
 
-    if (json.isJsonObject()) {
-      rawCoordinates = json.getAsJsonObject().getAsJsonArray("coordinates");
-    } else {
-      rawCoordinates = json.getAsJsonArray();
-    }
+    rawCoordinates = json.getAsJsonArray();
```
tobrun: is this change required for this PR specifically?

langsmith:
@osana has expressed that she's comfortable with the state of this pr and that I can go ahead and merge to `master` so that next steps can happen

langsmith:
@osana has expressed that she's comfortable with the state of this pr and that I can go ahead and merge to `master` so that next steps can happen.

langsmith:
@osana expressed her comfort with my merging this, so I'm merging this pr. This will allow for next steps as outlined in my comment above.

langsmith:
@osana expressed her comfort with my merging this, so I'm merging this pr. This will allow for next steps as outlined in my comment above.

PR #907, opened 2018-10-22
BoundingBoxSerializer should unshift its points before serialization similar to PointSerializer

osana:


services-geojson/src/main/java/com/mapbox/geojson/gson/BoundingBoxSerializer.java
```diff
@@ -41,17 +44,24 @@ public JsonElement serialize(BoundingBox src, Type typeOfSrc, JsonSerializationC
...
 
     // Southwest
-    bbox.add(new JsonPrimitive(src.southwest().longitude()));
-    bbox.add(new JsonPrimitive(src.southwest().latitude()));
-    if (src.southwest().hasAltitude()) {
-      bbox.add(new JsonPrimitive(src.southwest().altitude()));
+    Point point = src.southwest();
+    List<Double> unshiftedCoordinates =
+            CoordinateShifterManager.getCoordinateShifter().unshiftPoint(point);
+
```
Chaoba: Maybe we should check list's size in case something is wrong in `CoordinateShifterManager.getCoordinateShifter().unshiftPoint(point);`

langsmith:
Looks good to me. In line with the larger changes being made in `/mapbox-java` for shifting 👍

Issue #908, opened 2018-10-29
NoClassDefFoundError for BindingContextFactory in MapboxDirections.java


danesfeder:
Seeing this in crash logs, it seems there's some sort of issue resolving the `BindingContextFactory`.  

```
Fatal Exception: java.lang.NoClassDefFoundError
Failed resolution of: Lcom/sun/xml/internal/ws/spi/db/BindingContextFactory;
    com.mapbox.api.directions.v5.MapboxDirections.errorDidOccur (MapboxDirections.java:194)
    com.mapbox.api.directions.v5.MapboxDirections.access$000 (MapboxDirections.java:60)
    com.mapbox.api.directions.v5.MapboxDirections$1.onResponse (MapboxDirections.java:146)
    retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run (ExecutorCallAdapterFactory.java:70)
```

When navigating to `MapboxDirections.java` in the Navigation SDK codebase I see:
![screen shot 2018-10-29 at 9 20 15 am](https://user-images.githubusercontent.com/8434572/47652952-b7fc3200-db5d-11e8-8ffa-d820fee80c4d.png)

@osana @zugaldia could this be some sort of dependency conflict in the navigation repo?

zugaldia:
@danesfeder I think this is related to a ProGuard setting that we cannot include in this project (because it's a pure Java project) but that @tobrun recently added with https://github.com/mapbox/mapbox-gl-native/pull/13103. Could you try with a nav build including that ProGuard rule, or not using ProGuard at all?

danesfeder:
@zugaldia ah okay thank you - I'll give that a shot!

danesfeder:
Hmm it looks like we are including this already https://github.com/mapbox/mapbox-navigation-android/blob/master/libandroid-navigation/proguard-consumer.pro#L13

cc @Guardiola31337

Guardiola31337:
@zugaldia @danesfeder the `dontwarn` Proguard rule doesn't fix this - actually, it hides it until it crashes at runtime. This was introduced by https://github.com/mapbox/mapbox-java/pull/758 and https://github.com/mapbox/mapbox-java/pull/765 

The problem here is that we're using a class (`BindingContextFactory` - I guess it's included in some JDKs) which is not available on Android. So when the exception is thrown at runtime Android is not able to find that class and 💥 

[...]

PR #909, opened 2018-10-29
Remove dependency on BindingContextFactory

### LINKED ISSUE #852: Problem with BindingContextFactory dependency and use
mapbox services version: 3.3.0
device: lots of devices but I below stacktrace is for Pixel 1

Hi
I have a problem with BindingContextFactory logger in my app. I'm using my navigation sdk which is using my mapbox android services (I get other mapbox java services module from original one) with some changes in API models. the problem is even when I run app(my app or mapbox nav test app) in debug mode I get the below error:

```

FATAL EXCEPTION: main
     Process: com.mapbox.services.android.navigation.testapp, PID: 26431
     java.lang.NoClassDefFoundError: Failed resolution of: Lcom/sun/xml/internal/ws/spi/db/BindingContextFactory;
     at com.mapbox.api.directions.v5.MapboxDirections.errorDidOccur(MapboxDirections.java:197)
     at com.mapbox.api.directions.v5.MapboxDirections.access$000(MapboxDirections.java:62)
     at com.mapbox.api.directions.v5.MapboxDirections$1.onResponse(MapboxDirections.java:149)
     at retrofit2.ExecutorCallAdapterFactory$ExecutorCallbackCall$1$1.run(ExecutorCallAdapterFactory.java:70)
     at android.os.Handler.handleCallback(Handler.java:790)
     at android.os.Handler.dispatchMessage(Handler.java:99)
     at android.os.Looper.loop(Looper.java:164)
     at android.app.ActivityThread.main(ActivityThread.java:6494)
      at java.lang.reflect.Method.invoke(Native Method)
      at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run(RuntimeInit.java:438)
      at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:807)
       Caused by: java.lang.ClassNotFoundException: Didn't find class "com.sun.xml.internal.ws.spi.db.BindingContextFactory" on path: DexPathList[[zip file "/data/app/com.mapbox.services.android.navigation.testapp-ibn2OQUrbMDZayAFOVWYUg==/base.apk"],nativeLibraryDirectories=[/data/app/com.mapbox.services.android.navigation

```

when in Navigation SDK I open MapboxDirections the BindingContextFactory is red in both code and import. I even sync and try the original mapbox services but it's still red! and at last I clone the original version of Mapbox Navigation SDK and it's still red! I have many crashes because of this logger I even tried to find this dependency and add it but I can't find it. is it JDK related lib? I'm using java version 8. 
I don't know what is the problem exactly but seems it leads to lots of problem! if you know how we can solve it please let me know because I didn't find any real related thing to this logger on google. is it related to how I import dependency? I'm using implementation keyword but I didn't use BidningContextFactory in my app or Nav SDK
I even tried to add dependency directly and I find this but it doesn't include the class as well:
`implementation (group: 'com.sun.xml.ws', name: 'jaxws-rt', version: '2.1.4')`
is there any specific reason that you use BindigContextFactory? I actually want to suggest, delete those line and use something else for logging.
P.S: I intentionally force server to give me error instead of 200 for testing this

osana:
closes #852

### DIFF
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
@@ -24,7 +24,6 @@ import com.mapbox.core.utils.ApiCallHelper;
 import com.mapbox.core.utils.MapboxUtils;
 import com.mapbox.core.utils.TextUtils;
 import com.mapbox.geojson.Point;
-import com.sun.xml.internal.ws.spi.db.BindingContextFactory;
 
 import java.io.IOException;
 import java.lang.annotation.Annotation;
@@ -32,6 +31,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Locale;
 import java.util.logging.Level;
+import java.util.logging.Logger;
 
 import okhttp3.ResponseBody;
 import retrofit2.Call;
@@ -184,14 +184,14 @@ public abstract class MapboxDirections extends
     Converter<ResponseBody, DirectionsError> errorConverter =
       getRetrofit().responseBodyConverter(DirectionsError.class, new Annotation[0]);
     if (callback == null) {
-      BindingContextFactory.LOGGER.log(
+      Logger.getLogger(MapboxDirections.class.getName()).log(
         Level.WARNING, "Failed to complete your request and callback is null");
     } else {
       try {
         callback.onFailure(getCall(),
           new Throwable(errorConverter.convert(response.errorBody()).message()));
       } catch (IOException ioException) {
-        BindingContextFactory.LOGGER.log(
+        Logger.getLogger(MapboxDirections.class.getName()).log(
           Level.WARNING, "Failed to complete your request. ", ioException);
       }
     }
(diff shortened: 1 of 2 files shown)


Guardiola31337:
As commented in https://github.com/mapbox/mapbox-java/issues/908#issuecomment-433998887

> After fixing this I believe that Proguard rule won't be necessary anymore so we'll need to remove it from the different projects in which is included (I'll remark this in the PR).

Let's make sure that we remove the Proguard rule from the Maps and Navigation SDKs (I believe these are the only ones including it) when OP gets released @danesfeder @tobrun

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -184,14 +184,14 @@ private void errorDidOccur(@Nullable Callback<DirectionsResponse> callback,
     Converter<ResponseBody, DirectionsError> errorConverter =
       getRetrofit().responseBodyConverter(DirectionsError.class, new Annotation[0]);
     if (callback == null) {
-      BindingContextFactory.LOGGER.log(
+      Logger.getLogger(MapboxDirections.class.getName()).log(
```
zugaldia: Instead of duplicating `Logger.getLogger(MapboxDirections.class.getName())` several times, could you extract it as a local private class-level variable instead?

zugaldia:
> Let's make sure that we remove the Proguard rule from the Maps and Navigation SDKs (I believe these are the only ones including it) when OP gets released @danesfeder @tobrun

Could you ticket that out on https://github.com/mapbox/mapbox-gl-native?

PR #938, opened 2018-12-12
to improve performance: do not use BigDecimal for rounding

### LINKED ISSUE #932: Geometry toJson() is very slow
Converting a feature's geometry to geojson by calling toJson() can be extremely slow for complex geometries. 

#### Expected Behavior
Convert a complex geometry to geojson in < 100ms

#### Current Behavior
A 3700 point polygon takes ~646ms, which totals to nearly 2 seconds for us to convert all the geometries of queryRenderedFeatures() to json.

#### Possible Solution
I've narrowed the issue to the gson serializer, PositionSerializer. It appears creating and tweaking a BigDecimal for each the latitude and longitude is very expensive.

Instead of using PositionSerializer which uses this, 

```
public JsonElement serialize(Position src, Type typeOfSrc, JsonSerializationContext context) {
    JsonArray rawCoordinates = new JsonArray();

    BigDecimal lat = new BigDecimal(src.getLatitude());
    lat = lat.setScale(7, RoundingMode.HALF_UP).stripTrailingZeros();

    BigDecimal lon = new BigDecimal(src.getLongitude());
    lon = lon.setScale(7, RoundingMode.HALF_UP).stripTrailingZeros();

    rawCoordinates.add(new JsonPrimitive(lon));
    rawCoordinates.add(new JsonPrimitive(lat));

    // Includes altitude
    if (src.hasAltitude()) {
      rawCoordinates.add(new JsonPrimitive(src.getAltitude()));
    }

    return rawCoordinates;
  }
```

I overrode the serialize method to do this:

 ```
public JsonElement serialize(Position src, Type typeOfSrc, JsonSerializationContext context) {
            JsonArray rawCoordinates = new JsonArray();
            rawCoordinates.add(trim(src.getLongitude()));
            rawCoordinates.add(trim(src.getLatitude()));
            return rawCoordinates;
        }

        private double trim(double input) {
            return Math.round(input * 100000000D) / 100000000D;
        }
```

and saw a 10x improvement. My call to get json from queryRenderedFeatures went from 2 seconds to 200ms. 

I'm sure there's a more elegant/precise solution that what I've created. Just wanted to point it out for anyone else who's having issues with large/complex geometries being very slow to convert.

### PR DESCRIPTION
**Pull request by osana**
closes #932

### DIFF
--- a/services-geojson/src/main/java/com/mapbox/geojson/utils/GeoJsonUtils.java
+++ b/services-geojson/src/main/java/com/mapbox/geojson/utils/GeoJsonUtils.java
@@ -1,0 +1,25 @@
+package com.mapbox.geojson.utils;
+
+/**
+ * GeoJson utils class contains method that can be used throughout geojson package.
+ *
+ * @since 4.3.0
+ */
+public class GeoJsonUtils {
+
+  private static double ROUND_PRECISION = 10000000.0;
+  private static long MAX_DOUBLE_TO_ROUND = (long) (Long.MAX_VALUE / ROUND_PRECISION);
+
+  /**
+   * Trims a double value to have only 7 digits after period.
+   *
+   * @param value to be trimed
+   * @return trimmed value
+   */
+  public static double trim(double value) {
+    if (value > MAX_DOUBLE_TO_ROUND || value < -MAX_DOUBLE_TO_ROUND) {
+      return value;
+    }
+    return Math.round(value * ROUND_PRECISION) / ROUND_PRECISION;
+  }
+}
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: services-geojson/src/main/java/com/mapbox/geojson/utils/GeoJsonUtils.java [hunk 1]
```diff
@@ -0,0 +1,19 @@
...
+public class GeoJsonUtils {
+
+  /**
+   * Trims a double value to have only 7 digits after period.
+   *
+   * @param value to be trimed
+   * @return trimmed value
+   */
+  public static double trim(double value) {
+    return Math.round(value * 10000000.0) / 10000000.0;
```
LukasPaczos: Can we do a check for infinity overflow here after multiplication and throw an exception if it occurs?

PR #942, opened 2018-12-17
Added waypoint targets to MapboxDirections request

### LINKED ISSUE #939: Add waypoint targets to MapboxDirections
See last `query parameter` in https://www.mapbox.com/api-documentation/#retrieve-directions

> If this parameter is provided, the Directions API will compute the side of the street, left or right, for each target based on the  waypoint_targets and the driving direction. 

Added for iOS in https://github.com/mapbox/MapboxDirections.swift/pull/326

cc @mapbox/navigation-android

osana:
closes #939


Guardiola31337:
I went ahead and fix above comments myself so we can 🚀

I hope that was ok with you @osana 🙏

PR #946, opened 2018-12-20
[WIP] Added MapboxDirectionsRefresh 

devotaaabel:
I'm currently exploring options for sharing repeated code between the `directions-refresh` endpoint and `map-matching`. They share the same responses but the requests are slightly different:

- `coordinates` are a path param in `map-matching`, but they are a query param in `directions-refresh`
- Ideally we will omit `radiuses` from `directions-refresh`, because the field is being overwritten on the back end anyway 
- Obviously the path is different, and `directions-refresh` is a `POST` request while `map-matching` is a `GET`

I've explored some inheritance options, but AutoValue doesn't play well with inheritance. The implementation in this PR just duplicates the code, but I'm currently exploring composition options (while trying to not introduce breaking changes to `MapboxMapMatching`). If anyone has any thoughts/ideas, please let me know!

Addresses https://github.com/mapbox/navigation-sdks/issues/87 
Blocked by https://github.com/mapbox/api-directions/pull/1692

cc. @akitchen

### DIFF
--- a/samples/src/main/java/com/mapbox/samples/BasicDirectionsRefresh.java
+++ b/samples/src/main/java/com/mapbox/samples/BasicDirectionsRefresh.java
@@ -1,0 +1,39 @@
+package com.mapbox.samples;
+
+import com.mapbox.api.directionsrefresh.v1.MapboxDirectionsRefresh;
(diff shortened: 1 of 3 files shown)


devotaaabel:
Directions refresh is going in a different direction so going to close this.

PR #948, opened 2019-01-08
add getting mapmatching through POST method

### LINKED ISSUE #947: Add Map Matching API HTTP POST support
Details here https://www.mapbox.com/api-documentation/#using-http-post

> The Map Matching API supports access via the HTTP `POST` method. To submit a request using HTTP POST, you must make the following changes to the request:
>
> 1. The HTTP method must be `POST`
> 2. The `Content-Type` of the request must be `application/x-www-form-urlencoded`
> 3. The coordinate list must be present in the request body as the `coordinates=` parameter, and there must be no coordinates in the URL
> 4. The `access_token` parameter must be part of the `POST` URL, not the body
> 5. All other parameters must be part of the request body
>
> An example `POST` request looks like this:
>
> ```
>POST /matching/v5/mapbox/driving?access_token=your-access-token HTTP/1.0
>Content-Type: application/x-www-form-urlencoded
>
>coordinates=2.344003915786743,48.85805170891599;2.346750497817993,48.85727523615161;2.348681688308716,48.85936462637049;2.349550724029541,48.86084691113991;2.349550724029541,48.8608892614883;2.349625825881958,48.86102337068847;2.34982967376709,48.86125629633996&steps=true&tidy=true&waypoints=0;6
>```

cc @kbauhaus @osana

osana:
closes #947


services-matching/src/main/java/com/mapbox/api/matching/v5/MapMatchingService.java
```diff
@@ -84,4 +87,76 @@
     @Query("waypoints") String waypoints,
     @Query("waypoint_names") String waypointNames,
     @Query("approaches") String approaches);
+
+
+  /**
+   * Constructs the POSt call using the information passed in through the
```
Guardiola31337: Minor thing - Typo *`POSt` instead of `POST`

services-matching/src/main/java/com/mapbox/api/matching/v5/MapMatchingService.java
```diff
@@ -84,4 +87,76 @@
...
+   * @param voiceUnits         voice units
+   * @param waypoints          Which input coordinates should be treated as waypoints.
+   * @param waypointNames     wustom names for waypoints used for the arrival instruction.
+   * @param approaches         which side of the road to approach a waypoint.
+   * @return the MapMatchingResponse in a Call wrapper
+   * @since 4.4.0
+   */
+  @FormUrlEncoded
+  @POST("matching/v5/{user}/{profile}")
+  Call<MapMatchingResponse> getPostCall(
```
Guardiola31337: Isn't the `get` in https://github.com/mapbox/mapbox-java/blob/d90e3edf1e1a1c9a7fe325c7992914b16cff2267/services-matching/src/main/java/com/mapbox/api/matching/v5/MapMatchingService.java#L69 noting that it is a `GET` request? If so, should we name this method `postCall` instead?

services-matching/src/main/java/com/mapbox/api/matching/v5/MapboxMapMatching.java
```diff
@@ -338,6 +364,28 @@ public static Builder builder() {
     private String[] waypointNames;
     private String[] approaches;
 
+    /**
+     * Use GET method to request data (default).
```
Guardiola31337: Minor comment - Shouldn't this be `POST` instead of `GET`? It seems it got reversed with the `get` method below 👇

PR #954, opened 2019-01-18
Update gradle tools version, update CI image

tobrun:


osana:
@tobrun The problem in CI seems similar to the one I had in gl-native: to https://github.com/mapbox/mapbox-gl-native/pull/13384

tobrun:
@osana forgot about that fix but in the meantime I just addressed the encoding issues. You can replace them by searching for the html entity equivalent online.

PR #959, opened 2019-02-08
Add via waypoints support to MapboxDirections

### LINKED ISSUE #956: Add waypoint indices option for Directions API
The Directions API will soon expose a waypoints parameter similar to the Map Matching API’s: Project-OSRM/osrm-backend#5345. 

We should add / account for this parameter - iOS equivalent in https://github.com/mapbox/MapboxDirections.swift/issues/334

Guardiola31337:
Fixes #956 

- Adds via waypoints support to `MapboxDirections`


osana:
@Guardiola31337 Two questions:
- could you add a test to make sure that `waypoints` are being added to the request string
- CI changes ? I am guessing that will be changed.

Guardiola31337:
@osana 
> - could you add a test to make sure that `waypoints` are being added to the request string

Of course! This is updated and ready for another round of 👀 

> - CI changes ? I am guessing that will be changed.

Yeah, I've released momentarily a `SNAPSHOT` to test OP's changes in the Navigation side. It's already reverted.

PR #961, opened 2019-02-11
Fix via waypoints APIs naming

Guardiola31337:
- Follow up from https://github.com/mapbox/mapbox-java/pull/959

We found in https://github.com/mapbox/mapbox-navigation-android/pull/1733#discussion_r255541722 that naming could be confusing and should be more explicit in order to make clients' life easier.

### DIFF
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
@@ -92,7 +92,7 @@ public abstract class MapboxDirections extends
       voiceUnits(),
       exclude(),
       approaches(),
-      waypoints(),
+      viaWayPoints(),
       waypointNames(),
       waypointTargets());
   }
@@ -362,7 +362,7 @@ public abstract class MapboxDirections extends
   abstract String approaches();
 
   @Nullable
-  abstract String waypoints();
+  abstract String viaWayPoints();
 
   @Nullable
   abstract String waypointNames();
@@ -424,7 +424,7 @@ public abstract class MapboxDirections extends
     private Point destination;
     private Point origin;
     private String[] approaches;
-    private Integer[] waypoints;
+    private Integer[] viaWayPoints;
     private String[] waypointNames;
     private Point[] waypointTargets;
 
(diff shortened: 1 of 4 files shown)


osana:
@Guardiola31337 
If we are doing this name change, we should do a similar name change in MapboxMapMatching
I guess we can deprecate the use of `waypoints` and add `viawaypoints`  (for now)

Also please `viawaypoints` to RouteOptions

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -424,7 +424,7 @@ public static Builder builder() {
     private Point destination;
     private Point origin;
     private String[] approaches;
-    private Integer[] waypoints;
```
akitchen: This was correct as-was. Should be `waypoints` as one word, no capital P

Guardiola31337: Yeah, you're correct. Will update in a follow-up PR. Thanks for flagging 🙇

Guardiola31337:
@osana 
> If we are doing this name change, we should do a similar name change in MapboxMapMatching
I guess we can deprecate the use of `waypoints` and add `viawaypoints`  (for now)

Yeah, I agree. We should be consistent across the different services. Will implement ☝️ if we're not cool breaking SemVer.

> Also please add `viawaypoints` to RouteOptions

You mean, adding "waypoints" option to Map Matching route options, right? I've already added it to Directions 👀 https://github.com/mapbox/mapbox-java/blob/57e8710c6f1628df04277ac1978000cfad459141/services-directions/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java#L284 https://github.com/mapbox/mapbox-java/blob/57e8710c6f1628df04277ac1978000cfad459141/services-directions/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java#L569

osana:
@Guardiola31337 Please add a call to viaWaypoints()  when RouteOptions are being saved/generated: https://github.com/mapbox/mapbox-java/blob/6e6110a303f7704524734022e066ef5e8d90982a/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java#L235

I believe tests might be updated as well.

PR #962, opened 2019-02-11
Fix waypoints APIs naming for map matching and directions services

Guardiola31337:
Follow up PR from latest feedback in https://github.com/mapbox/mapbox-java/pull/961

Edited:
- Fix waypoints APIs naming for map matching and directions services - from `waypoints` to ~`separatesLegs`~ waypoint indices 👀 https://github.com/mapbox/mapbox-java/pull/962#issuecomment-462809409

Noting that for the Map Matching service we're deprecating https://github.com/mapbox/mapbox-java/blob/c834c07361e8b6283b93b098930413a06e933c64/services-matching/src/main/java/com/mapbox/api/matching/v5/MapboxMapMatching.java#L488 in favor of https://github.com/mapbox/mapbox-java/blob/aba088c98a2bab20c155667eab8d447275cb03e2/services-matching/src/main/java/com/mapbox/api/matching/v5/MapboxMapMatching.java#L507 to avoid breaking SemVer.

cc @1ec5

### DIFF
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
@@ -92,7 +92,7 @@ public abstract class MapboxDirections extends
       voiceUnits(),
       exclude(),
       approaches(),
-      viaWayPoints(),
+      waypointIndices(),
       waypointNames(),
       waypointTargets());
   }
@@ -235,6 +235,7 @@ public abstract class MapboxDirections extends
         RouteOptions.builder()
           .profile(profile())
           .coordinates(coordinates())
+          .waypointIndices(waypointIndices())
           .waypointNames(waypointNames())
           .waypointTargets(waypointTargets())
           .continueStraight(continueStraight())
@@ -362,7 +363,7 @@ public abstract class MapboxDirections extends
   abstract String approaches();
 
   @Nullable
-  abstract String viaWayPoints();
+  abstract String waypointIndices();
 
   @Nullable
   abstract String waypointNames();
@@ -424,7 +425,7 @@ public abstract class MapboxDirections extends
     private Point destination;
     private Point origin;
     private String[] approaches;
-    private Integer[] viaWayPoints;
+    private Integer[] waypointIndices;
     private String[] waypointNames;
     private Point[] waypointTargets;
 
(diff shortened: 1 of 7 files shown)


services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -794,24 +795,24 @@ public Builder addApproaches(String... approaches) {
     abstract Builder approaches(@Nullable String approaches);
 
     /**
-     * Optionally, set which input coordinates should be treated as via way points.
+     * Optionally, set which input coordinates should be treated as separate legs.
      * <p>
      * Most useful in combination with  steps=true and requests based on traces
```
danesfeder: Does `traces` make sense here?  Reading this javadoc, I'm still not entirely sure what `separateLegs` means. 

> set which input coordinates should be treated as separate legs.

By adding coordinates to a Directions API request, you are creating new legs.  I think it may make sense here to say that `separateLegs` is directly influencing the guidance behavior when set.

Guardiola31337:
After some discussion, we agreed on fixing the naming to "waypoint indices":

- In the Map Matching service the indices added mean which coordinates should act as "actual waypoints".
- In the Directions service the indices added mean which waypoints should be "silent" (from a guidance point of view). Note the explicit `silent` naming used here.

Hope this is clear now.

cc @osana @1ec5 @mcwhittemore

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
Guardiola31337: https://github.com/mapbox/mapbox-java/pull/962#issuecomment-462809409

Guardiola31337:
> - In the Directions service the indices added mean which waypoints should be "silent" (from a guidance point of view). Note the explicit `silent` naming used here.

Actually, this is not technically correct because coordinate indices not added act as silent waypoints, so not adding them is what makes them silent.

I'm updating (last time, promise!) the naming to `waypointIndices` and adding a note to Javadoc so it makes total sense to developers using the services. FWIW both Map Matching and Directions services will use the same nomenclature which was the initial goal of this PR.

Sorry for the back and forth.

langsmith:
@danesfeder & @Guardiola31337 , if needed, please don't forget about updating mentions of waypoints in https://docs.mapbox.com/android/java/overview/directions/  and/or https://docs.mapbox.com/android/java/overview/map-matching/

thanks

cc @colleenmcginnis

PR #966, opened 2019-02-13
Fixed typo in services build.gradle resulting in directions refresh not being included in services artifact

devotaaabel:


services-directions/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java
```diff
@@ -39,7 +39,7 @@ public static Builder builder() {
    * @since 4.4.0
    */
   @Nullable
-  public abstract String routeIndex();
+  public abstract Integer routeIndex();
```
Guardiola31337: This is breaking SemVer and it seems out of scope of this PR.

osana: Good point!

Issue #967, opened 2019-02-13
Release 4.4.1


osana:
### Includes 
- Directions refresh was not being included in services artifact [#966](https://github.com/mapbox/mapbox-java/pull/966)

  ### Release checklist @osana 
    ~~[ ] If there's a new SNAPSHOT version, then update this repo's `README.md` file~~
    - [x] Change version name to `4.4.1` in `gradle.properties` (removing `-SNAPSHOT`)
    - [ ] Publish `4.4.1` artifact to Maven Central
    - [ ] Change version name back to having `-SNAPSHOT` at the end
    - [ ] Update `CHANGELOG.md`
    - [ ] Tag `v new release version` in GitHub
    - [ ] Update dependencies in the Maps SDK
    
   ### Post-release @langsmith   
    - [ ] Update Javadoc for the Mapbox Java docs section in `/android-docs` repo
    - [ ] Update the demo app
    - [ ] Update version # in config files in /help @langsmith
    - [ ] **After updating the demo app**: Update `android-docs`
    - [ ] Update the `constants.js` file in `android-docs` repo so that `implementation 'com.mapbox.mapboxsdk:mapbox-sdk-services:x.x.x'` line is updated in [the Gradle section of the MAS overview page](https://www.mapbox.com/android-docs/java-sdk/overview/)
    - [ ] run `git submodule update --remote examples/maps` to pull in updated example code
    - [ ] Update version # in Studio Preview
    
    /cc: @mapbox/android

osana:
Closing this ticket as the release should be 4.5.0 instead of 4.4.1 as Directions Refresh was not included in 4.4.0.

It was a typo fix but from api stand point : new APIs were introduced

PR #968, opened 2019-02-13
Release 4.4.1

osana:
have to do a patch release. Directions Refresh was not included int the artifact

addresses #967

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
 # Project-wide Gradle settings.
-VERSION_NAME=4.5.0-SNAPSHOT
+VERSION_NAME=4.4.1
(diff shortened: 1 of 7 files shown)


gradle.properties
```diff
@@ -1,5 +1,5 @@
 # Project-wide Gradle settings.
-VERSION_NAME=4.5.0-SNAPSHOT
+VERSION_NAME=4.4.1
```
Guardiola31337: This should be `4.5.0` instead of `4.4.1`. We're adding new APIs and make them available. Some APIs weren't included in `4.4.0`. Talked to @osana and we can fix it because `4.4.1` wasn't released yet 😌

PR #973, opened 2019-02-18
Add explicit US locale to String.Format of bouding box

### LINKED ISSUE #972: Incorrect BoundingBox String formatting
A client has reached out that the bounding box in his request url to route-tiles is being formatted as decimals with comma (instead of a point). The source of this is that String.format uses the device locale. 

To fix this issue we need to explitcly need to state a Locale in 
https://github.com/mapbox/mapbox-java/blob/master/services-route-tiles/src/main/java/com/mapbox/api/routetiles/v1/MapboxRouteTiles.java#L42

cc @osana, fyi @devotaaabel

tobrun:
Closes #972


services-route-tiles/local.properties
```diff
@@ -0,0 +1,9 @@
+## This file must *NOT* be checked into Version Control Systems,
```
LukasPaczos: Missing `.gitignore` entry? :)

PR #988, opened 2019-04-02
Adding Isochrone API support

### LINKED ISSUE #905: Add support for the isochrone API
The isochrone API was last week announced in [this](https://blog.mapbox.com/announcing-the-isochrone-api-3a72704cd046) blogpost. The API documentation can be found [here](https://www.mapbox.com/api-documentation/#isochrone). This issue is a feature request for adding a new module dedicated to this API. 

cc @osana

langsmith:
This pr resolves #905 by adding support for [the Mapbox Isochrone API](https://docs.mapbox.com/api/navigation/#isochrone).

As @zugaldia said in https://github.com/mapbox/mapbox-java/issues/905#issuecomment-431387345, this isochrone support work is basically the same that was done to add support for the Tilequery API.

[The Mapbox Android demo app has an Isochrone API example](https://github.com/mapbox/mapbox-android-demo/blob/master/MapboxAndroidDemo/src/main/java/com/mapbox/mapboxandroiddemo/examples/javaservices/IsochroneActivity.java) ([pr is here](https://github.com/mapbox/mapbox-android-demo/pull/884)), but the API is interacted with in that example in a more raw way than using a formal Mapbox Java SDK wrapper.


langsmith:
Would love 👁👁 on this when you get a chance @osana . No rush on my end. 

- I get the following error when I run `MapboxIsochroneTest` on my local machine. It's the same crash that is in CircleCI. 

Screen Shot 2019-04-02 at 3 34 18 PM


Screen Shot 2019-04-02 at 3 26 05 PM

Any idea what I'm doing wrong? I'm not very familiar with `AutoValue`.

- [The code within `build()`](https://github.com/mapbox/mapbox-java/pull/988/files#diff-3b322564527a1e662e970782b2f3f1a2R244) is probably the area that will need most adjusting. `Tilequery` immediately does `autoBuild()` inside of its `build()` method, so that's what I mirrored with isochrone. I don't think `MapboxIsochrone` would require me to use other methods inside of `build()` like some of the other Java wrappers do (e.g. https://github.com/mapbox/mapbox-java/blob/master/services-optimization/src/main/java/com/mapbox/api/optimization/v1/MapboxOptimization.java#L489-L493)

osana:
@langsmith You are missing ```@AutoValue``` annotation at the very top of the class 

```
 * @see <a href="https://docs.mapbox.com/api/navigation/#isochrone">API documentation</a>
  * @since 4.5.0
  */
+@AutoValue
 public abstract class MapboxIsochrone extends MapboxService<FeatureCollection, IsochroneService> {
 ```

Also ```isochrone``` should be added to the TESTABLE_MODULES list in project's build.gradle:

```
def TESTABLE_MODULES = ["services",
                         "services-core",
                         "services-directions",
                         "services-geocoding",
                         "services-geojson",
+                        "services-isochrone",
                         "services-matching",
```

**Comment by langsmith** [comment]
D'oh. Thanks @osana . Just pushed refactoring to address `build.gradle` and the `@AutoValue` annotation.

**Comment by langsmith** [comment]
Running tests on my local machine seems to be working, so now it's a matter of fixing the failing ones 💪 . Progress...

Screen Shot 2019-04-03 at 10 43 15 AM

**Comment by osana** [comment]
💪  Make sure ```@since``` tags are properly updated as well

**Comment by osana** [comment]
@langsmith tags should be 4.6.0  (I believe) as this is a new feature and we are currently at 4.5.0

**Comment by langsmith** [comment]
>@langsmith tags should be 4.6.0 (I believe) as this is a new feature and we are currently at 4.5.0

🤦‍♂️ , right, right.

**Comment by langsmith** [comment]
fyi, the `@since` additions in non-isochrone-related classes was because of checkstyle

File: services-isochrone/src/test/java/com/mapbox/api/isochrone/IsochroneTestUtils.java [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+import org.hamcrest.junit.ExpectedException;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Rule;
+
+import java.io.IOException;
+
+public class IsochroneTestUtils extends TestUtils {
+    protected static final String ISOCHRONE_WITH_POLYGONS_VALID = "isochroneWithPolygons.json";
+    protected static final String ISOCHRONE_NO_POLYGONS_VALID = "isochroneNoPolygons.json";
```
osana: This one does not seem to be used.
It looks like all the tests will be using : ISOCHRONE_WITH_POLYGONS_VALID

services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
```diff
@@ -199,4 +199,20 @@ public static String formatWaypointNames(String[] waypointNames) {
...
   }
+
+  /**
+   * Converts String array with waypoint_names values
+   * to a string ready for API consumption.
+   *
+   * @param stringList a single String which has a list of comma-separated values
+   *                   such as HEX color values or minutes.
+   * @return the number of elements in between commas in the raw stringList.
+   * @since 3.3.0
```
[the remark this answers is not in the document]
langsmith: fixed and rebased

langsmith:
>Another thought that we usually allow a List of objects to be given to a Builder, that in turn creates a String out of a list that is request friendly.

👍 Ok @osana , I just pushed a commit which enables passing through `List<Integer>`. 

- https://github.com/mapbox/mapbox-java/pull/988/files#diff-3b322564527a1e662e970782b2f3f1a2R198
- https://github.com/mapbox/mapbox-java/pull/988/files#diff-3b322564527a1e662e970782b2f3f1a2R291

services-isochrone/src/main/java/com/mapbox/api/isochrone/MapboxIsochrone.java
```diff
@@ -0,0 +1,334 @@
...
+     * up to four contours. Times must be in increasing order. For example,
+     * "5,20,40,50" and not "20,10,40,55". The maximum time that can be
+     * specified is 60 minutes.
+     *
+     * @param listOfMinuteValues an integer list with at least one number
+     *                           for the minutes which represent each contour
+     * @return this builder for chaining options together
+     * @since 4.6.0
+     */
+    public Builder contoursMinutes(@NonNull List<Integer> listOfMinuteValues) {
```
osana: Here is an example how it is done in other parts of mapbox-java. I think it would be good to follow the same pattern. https://github.com/mapbox/mapbox-java/blob/4afbd732b758010a70c742dbe3a3f15a9b7d13a7/services-matching/src/main/java/com/mapbox/api/matching/v5/MapboxMapMatching.java#L639

langsmith:
Ok @osana . I've adjusted the contour minutes and colors methods to take `String ...` instead of `List<String>`.   Ready for another round of review when you get a chance 🙇

services-isochrone/src/main/java/com/mapbox/api/isochrone/MapboxIsochrone.java
langsmith: Fixed 👍

services-isochrone/src/test/java/com/mapbox/api/isochrone/MapboxIsochroneTest.java
```diff
@@ -0,0 +1,354 @@
...
+      .build();
+  }
+
+  @Test
+  public void build_optionalParameters() throws Exception {
+    MapboxIsochrone client = MapboxIsochrone.builder()
+      .accessToken(ACCESS_TOKEN)
+      .coordinates(testPoint)
+      .profile(testProfile)
+      .addContoursMinutes(5,25,40)
```
osana: Given that you are using the fixtures, Request should match the one that is in the fixture request 
contours_minutes=5,30,55&contours_colors=6706ce,04e813,4286f4

- This is just for the ease of debugging if problems occur in the future

langsmith: Ok, just pushed a change to fix this.

langsmith: and then rebased

PR #990, opened 2019-04-09
Fixing banner MAS models to match what actually comes from Directions API

### LINKED ISSUE #951: Fix banner MAS models to match what actually comes from Directions API
As part of https://github.com/mapbox/mapbox-navigation-android/pull/1543 work we've noticed that some banner Directions API / MAS models don’t match each other completely matches and some annotations need to be fixed in MAS so they match what actually comes from the API. These are the changes needed:

[`BannerInstructions`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerInstructions.java):
[`primary`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerInstructions.java#L48-L49) 👉 should be `@NonNull`

[`BannerText`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerText.java):
[`text`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerText.java#L42-L43) 👉 should be `@NonNull`
**UPDATED**
~[`type`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerText.java#L61-L63) 👉 should be `@NonNull`~
~[`modifier`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerText.java#L73-L74) 👉 should be `@NonNull`~

[`BannerComponents`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java):
[`text`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java#L44) 👉 not specified and should be `@NonNull`
[`type`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java#L64) 👉 not specified and should be `@NonNull`

cc @kevinkreiser @osana

langsmith:
Resolves #951 by adjusting the several annotations which @Guardiola31337 listed in #951.

@osana , I don't believe there are any other downstream changes that need to happen based on these annotation updates, but if so, please let me know.

### DIFF
--- a/services-directions/src/test/java/com/mapbox/api/directions/v5/models/BannerInstructionsTest.java
+++ b/services-directions/src/test/java/com/mapbox/api/directions/v5/models/BannerInstructionsTest.java
@@ -29,8 +29,8 @@ public class BannerInstructionsTest extends TestUtils {
   public void testSerializable() throws Exception {
     BannerInstructions bannerInstructions = BannerInstructions.builder()
       .distanceAlongGeometry(100d)
-      .primary(BannerText.builder().build())
-      .secondary(BannerText.builder().build())
+      .primary(BannerText.builder().text("Banner primary sample text").build())
+      .secondary(BannerText.builder().text("Banner secondary sample text").build())
       .build();
     byte[] serialized = TestUtils.serialize(bannerInstructions);
     assertEquals(bannerInstructions, deserialize(serialized, BannerInstructions.class));
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
@@ -1,5 +1,6 @@
 package com.mapbox.api.directions.v5.models;
 
+import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import com.google.auto.value.AutoValue;
@@ -41,6 +42,7 @@ public abstract class BannerComponents extends DirectionsJsonObject
    * @return a single snippet of the full text instruction
    * @since 3.0.0
    */
+  @NonNull
   public abstract String text();
 
   /**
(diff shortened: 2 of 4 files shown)


osana:
👍  I am guessing once tests are fixed that would be good to go

langsmith:
>👍 I am guessing once tests are fixed that would be good to go

Yea, had to fix one test

services-directions/src/test/java/com/mapbox/api/directions/v5/models/BannerInstructionsTest.java
```diff
@@ -29,8 +29,8 @@ public void sanity() throws Exception {
   public void testSerializable() throws Exception {
     BannerInstructions bannerInstructions = BannerInstructions.builder()
       .distanceAlongGeometry(100d)
-      .primary(BannerText.builder().build())
-      .secondary(BannerText.builder().build())
+      .primary(BannerText.builder().text("Banner primary sample text").build())
+      .secondary(BannerText.builder().text("Banner secondary sample text").build())
```
Guardiola31337: If `secondary` is `@Nullable`, this is not needed right? I mean, passing a `null` value (as before) should work too.

langsmith: Thanks for checking in @Guardiola31337 . Looks like `null` is fine as a parameter.

Screen Shot 2019-04-10 at 2 01 08 PM

Guardiola31337: Yeah and we'd be implicitly "testing" that `secondary` could be `@Nullable`, thanks for confirming @langsmith

PR #991, opened 2019-04-10
Added walking options

devotaaabel:
This is currently blocked until testing on staging is complete.

### DIFF
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
@@ -137,14 +137,6 @@ public interface DirectionsService {
    * @param waypointNames       custom names for waypoints used for the arrival instruction
    * @param waypointTargets     list of coordinate pairs for drop-off locations
    * @param enableRefresh       whether the routes should be refreshable
-   * @param walkingSpeed        walking speed in kilometers per hour
-   * @param walkwayBias         a factor that modifies the cost when encountering roads or paths
-   *                            that do not allow vehicles and are set aside for pedestrian use
-   * @param alleyBias           a factor that modifies the cost when alleys are encountered
-   * @param ferryBias           a factor that modifies the cost when ferries are encountered
-   * @param stepPenalty         a penalty in seconds added to each transition onto a path with
-   *                            steps or stairs
-   * @param maxHikingDifficulty the maximum difficulty of hiking trails that is allowed
    * @return the {@link DirectionsResponse} in a Call wrapper
    * @since 4.6.0
    */
@@ -174,12 +166,6 @@ public interface DirectionsService {
     @Field("waypoints") String waypointIndices,
     @Field("waypoint_names") String waypointNames,
     @Field("waypoint_targets") String waypointTargets,
-    @Field("enable_refresh") Boolean enableRefresh,
-    @Field("walking_speed") Double walkingSpeed,
-    @Field("walkway_bias") Double walkwayBias,
-    @Field("alley_bias") Double alleyBias,
-    @Field("ferry_bias") Double ferryBias,
-    @Field("step_penalty") Integer stepPenalty,
-    @Field("maxHikingDifficulty") Integer maxHikingDifficulty
+    @Field("enable_refresh") Boolean enableRefresh
   );
 }
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
@@ -175,11 +175,11 @@ public interface DirectionsService {
     @Field("waypoint_names") String waypointNames,
     @Field("waypoint_targets") String waypointTargets,
     @Field("enable_refresh") Boolean enableRefresh,
-    @Query("walking_speed") Double walkingSpeed,
-    @Query("walkway_bias") Double walkwayBias,
-    @Query("alley_bias") Double alleyBias,
-    @Query("ferry_bias") Double ferryBias,
-    @Query("step_penalty") Integer stepPenalty,
-    @Query("maxHikingDifficulty") Integer maxHikingDifficulty
+    @Field("walking_speed") Double walkingSpeed,
+    @Field("walkway_bias") Double walkwayBias,
+    @Field("alley_bias") Double alleyBias,
+    @Field("ferry_bias") Double ferryBias,
+    @Field("step_penalty") Integer stepPenalty,
+    @Field("maxHikingDifficulty") Integer maxHikingDifficulty
   );
 }
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
@@ -22,39 +22,47 @@ public interface DirectionsService {
    * Constructs the html get call using the information passed in through the
    * {@link MapboxDirections.Builder}.
    *
-   * @param userAgent          the user agent
-   * @param user               the user
-   * @param profile            the profile directions should use
-   * @param coordinates        the coordinates the route should follow
-   * @param accessToken        Mapbox access token
-   * @param alternatives       define whether you want to receive more then one route
-   * @param geometries         route geometry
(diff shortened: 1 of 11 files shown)


services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -902,6 +932,12 @@ public MapboxDirections build() {
         approaches(formattedApproaches);
       }
 
+      if (walkingOptions() == null) {
+        walkingOptions(WalkingOptions.builder().build());
```
Guardiola31337: Is `WalkingOptions` `@Nullable`? Do we actually need to pass in a non-null object then?

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -902,6 +932,12 @@ public MapboxDirections build() {
...
+      } else if (!DirectionsCriteria.PROFILE_WALKING.equals(profile())) {
+        throw new ServicesException("Walking options are for use with the walking profile");
```
Guardiola31337: Does the backend service return any error in this situation? If so, we should bubble up that error and not thrown an exception at this point.

services-directions/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
```diff
@@ -0,0 +1,132 @@
...
+import android.support.annotation.Nullable;
+
+import com.google.auto.value.AutoValue;
+
+/**
+ * Class for specifying options for use with the walking profile.
+ * @since 4.6.0
+ */
+@AutoValue
+public abstract class WalkingOptions {
```
Guardiola31337: Should we include these options as part of `RouteOptions`? If so, we should make sure that `toJson` / `fromJson` methods are implemented and serialization / deserializations work 👌

services-core/gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=4.5.0-SNAPSHOT
+VERSION_NAME=4.6.0-SNAPSHOT
```
osana: I noticed this when I was just releasing 4.6.0.
If you rebase this PR to master - you will not need this commit.

services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
```diff
@@ -152,6 +174,12 @@
     @Field("waypoints") String waypointIndices,
     @Field("waypoint_names") String waypointNames,
     @Field("waypoint_targets") String waypointTargets,
-    @Field("enable_refresh") Boolean enableRefresh
+    @Field("enable_refresh") Boolean enableRefresh,
+    @Query("walking_speed") Double walkingSpeed,
```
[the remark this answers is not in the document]
devotaaabel: @Guardiola31337 My understanding is that `@Field` is only used for `POST` because it's `@FormUrlEncoded`, but if you think this shouldn't be the case let me know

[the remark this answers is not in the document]
devotaaabel: @Guardiola31337 nevermind I just understood what you were saying. It's interesting because I thought they were supposed to be `@Field` but the example test passed with `@Query`... I'll update them anyway

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
devotaaabel: @Guardiola31337 this is so I don't have to do a null check [here](https://github.com/mapbox/mapbox-java/pull/991/files/6ab38103da15a761fcccc0acb36c9b88be878841#diff-7ee26788285a9a937ff59dfdcaff423aR111)

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -179,6 +179,18 @@ public void build_originDestinationGetAddedToListCorrectly() throws Exception {
...
+  @Test
+  public void build_walkingOptions() {
+    MapboxDirections directions = MapboxDirections.builder()
+      .destination(Point.fromLngLat(13.4930, 9.958))
+      .origin(Point.fromLngLat(1.234, 2.345))
+      .accessToken(ACCESS_TOKEN)
+      .profile(DirectionsCriteria.PROFILE_WALKING)
+      .walkingOptions(WalkingOptions.builder().alleyBias(2d).build())
+      .build();
+    assertTrue(directions.cloneCall().request().url().toString().contains("alley_bias=2.0"));
```
kevinkreiser: is it ok that this test is using a value that is out of bounds? valid values are from `-1` to `1`

services-directions/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
```diff
@@ -0,0 +1,150 @@
...
+     * @return this builder
+     * @since 4.8.0
+     */
+    public abstract Builder walkingSpeed(
+      @Nullable @FloatRange(from = 0.14, to = 6.94) Double walkingSpeed);
+
+    /**
+     * A factor that modifies the cost when encountering roads or paths that do not allow
+     * vehicles and are set aside for pedestrian use. Pedestrian routes generally attempt to
+     * favor using these walkways and sidewalks. The default walkway_factor is 0.
```
danpaz: May want to indicate what the range of `bias` values mean and their effect. Suggest adding something like: `The allowed range of values is from -1 to 1, where -1 indicates indicates preference to _avoid_ pedestrian walkways, 1 indicates preference to _favor_ pedestrian walkways, and 0 indicates no preference`. Same comment for alleyBias below.

osana:
@devotaaabel Make sure the  `@since` tags are updated to `4.8.0`

services-directions/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Walking speed in meters per second. Must be between 0.14 and 6.94 meters per second.
+   * Defaults to 1.42 meters per second
+   *
+   * @return walkingSpeed in meters per second
+   * @since 4.8.0
+   */
+  @Nullable
+  public abstract Double walkingSpeed();
```
Guardiola31337: Shouldn't options include the `@SerializedName`? That’s what the backend expects / generates

services-directions/src/test/java/com/mapbox/api/directions/v5/models/WalkingOptionsTest.java
```diff
@@ -0,0 +1,46 @@
...
+
+  import com.mapbox.api.directions.v5.WalkingOptions;
+  import com.mapbox.core.TestUtils;
+
+  import org.junit.Test;
+
+  import static org.junit.Assert.assertEquals;
+
+public class WalkingOptionsTest extends TestUtils {
+ private static final String JSON = "{\"walkingSpeed\":1.0,\"walkwayBias\":2.0,\"alleyBias\":3.0}";
```
Guardiola31337: This JSON isn’t something the backend would return they would return snake_case except for `maxHikingDifficulty` - wondering if we could ask to services team to make it snake_case too for consistency 🤔

services-directions/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
Guardiola31337: 👀 https://github.com/mapbox/mapbox-java/pull/991#discussion_r282667172

services-directions/src/test/java/com/mapbox/api/directions/v5/models/WalkingOptionsTest.java
Guardiola31337: I've just 👀 that `maxHikingDifficulty` is not supported anymore and edited ☝️

osana:
@devotaaabel 
Thank you for working on this PR.
Can I ask you to add some more description on the functionality this PR adds? 
I do not think we have an issue open so there is nothing to refer to either.

PR #992, opened 2019-04-11
Exclude *.java from being added to library jars

### LINKED ISSUE #989: Exclude AutoValue java files from jar
Ticketing this out as we didn't have a public ticket tracking this: currently AutoValue enabled dependencies from this repo will include generate java files as part of the jar generated. 

image

Capturing from upstream that this is unneeded and these files can safely be removed:
 - https://github.com/google/auto/issues/706

Considering we aren't able to bump to gradle 5.0 due to java 8 requirement, we should find a way to exclude this files. Either through a [gradle configuration](https://stackoverflow.com/questions/19575474/gradle-how-to-exclude-a-particular-package-from-a-jar) or [pom.xml configuration](https://stackoverflow.com/questions/26083326/how-do-i-exclude-java-files-from-a-jar-in-maven-jar-plugin) or ..

cc @osana

osana:
Mapbox java sdk  uses [AutoValue library](https://github.com/google/auto/blob/master/value/userguide/index.md) to generate boilerplate code for  value setters / getters and to generate `equals`, `hashCode` and `toString` methods.

It is a useful tool. However, it was reported that generated `.java` files end up in the library jars. This is undersirable. This PR excludes `.java` files from being added to the library jars.

closes #989

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -85,6 +85,11 @@ def RELEASE_MODULES = ["services",
                        "services-turf"]
 
 subprojects { subproject ->
+
+  tasks.withType(Jar) {
+    exclude "**/*.java"
+  }
+
   if (TESTABLE_MODULES.contains(subproject.name)) {
     subproject.apply plugin: "com.vanniktech.android.junit.jacoco"
     subproject.apply from: "${rootDir}/gradle/jacoco.gradle"


osana:
An easy way to check that there are  no java classes is to list jar content. 

```jar tf services-directions/build/libs/services-directions-4.6.0-SNAPSHOT.jar```

PR #997, opened 2019-04-11
Adding userAgent parameter to Isochrone request retrofit call URL

### LINKED ISSUE #996: Isochrone URL incorrectly built
What kind of issue is this?

- [ ] **Questions** are better to ask on [Stack Overflow](https://stackoverflow.com/questions/tagged/mapbox+android). We are actively monitoring the questions there and oftentimes, others have already asked a similar question.
- [ ] **Feature request** should start off by describing the problem the feature will resolve. Please open a GitHub ticket beforehand with information about the feature to spark some discussion about the topic and once you have our support.  
- [x] **Bug Report** can be reported using the provided template below:

#### Expected Behavior

Isochrone API support has landed and I started to refactor the Mapbox Android demo app's Isochrone example to use the Java SDK's wrapper. But I discovered that the URL isn't being created correctly 😭 

Isochrone API request URL should be correctly built with `mapbox` put in front of the chosen [IsochroneCriteria directions profile (cycling, walking, or driving)](https://github.com/mapbox/mapbox-java/blob/master/services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneCriteria.java).  

#### Current Behavior

The URL is missing the `/mapbox/` section that is required, which leads to a `Not found` error because the URL is incorrect.

Screen Shot 2019-04-11 at 3 52 58 PM

#### Possible Solution

If you look at the `profile` section of the chart in the Isochrone docs at https://docs.mapbox.com/api/navigation/#retrieve-isochrones-around-a-location, you'll see that the profiles are actually prefaced by `mapbox/`. For example, the driving one is actually `mapbox/driving` instead of `driving`. 

I think the best option would be to hardcode `/mapbox/` into the Retrofit request URL at https://github.com/mapbox/mapbox-java/blob/master/services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneService.java#L53

I guess the other option would be to adjust the constant Strings in [the IsochroneCriteria class](https://github.com/mapbox/mapbox-java/blob/master/services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneCriteria.java) by adding `mapbox/` in front of each constant String. But I think we'll run into some funky UTF-8 related coding issues. Probably easier to just harcode the Retrofit URL.

I tried `.profile(String.format("mapbox/%s", IsochroneCriteria.PROFILE_DRIVING))` in the demo app example, but the request URL ends up being `https://api.mapbox.com/isochrone/v1/mapbox%2Fdriving/12.490007,41.88729?......` I don't think the `%2F` would appear if we just hardcoded the Retrofit request URL. I might be wrong though.

#### Steps to Reproduce

1.  Open the demo app in Android Studio on the `ls-isochrone-refactor-to-use-Java-wrapper` branch. [Related pr is here](https://github.com/mapbox/mapbox-android-demo/pull/1029).
2. Run the app on a virtual/real device.
3. Tap on `Java Services` in the sidebar menu and then scroll down to the Isochrone API example
4. Tap once anywhere on the map and view the request URL in the Android Studio log cat.
5. Nothing is drawn on the map because `response.body()` is `null`. It's `null` because the request URL was invalid. For example, here's a sample URL that the demo app example is currently delivering: https://api.mapbox.com/isochrone/v1/driving/12.492492,41.88174?contours_minutes=14%2C35%2C53&access_token=PASTE-YOUR-ACCESS-TOKEN-HERE&contours_colors=80f442%2C403bd3%2Cbc404c&polygons=false&denoise=0.4&generalize=2.0

In the URL above, `https://api.mapbox.com/isochrone/v1/driving/` is missing the `/mapbox/` slug. If you add the slug between `/v1/` and `/driving/` (plus a valid mapbox token), you'll find that the Isochrone API will provide a valid response.

#### Context (Environment)

Only affect has been my refactor work in the demo app. This bug renders the wrapper unusable for myself and others.

langsmith:
Resolves #996 by adding a required `/mapbox/` slug to the Isochrone API service's Retrofit request URL. Based on my demo app refactoring work in https://github.com/mapbox/mapbox-android-demo/pull/1029, the #996 bug renders the wrapper unusable. 

The API's docs

Screen Shot 2019-04-11 at 3 57 45 PM

Totally fine if this is the wrong way to fix the bug. Figured I'd open something to get the conversation started.

### DIFF
--- a/services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneCriteria.java
+++ b/services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneCriteria.java
@@ -14,6 +14,13 @@ public class IsochroneCriteria {
 
 
   /**
+   * Mapbox default username.
+   *
+   * @since 4.7.0
+   */
+  public static final String PROFILE_DEFAULT_USER = "mapbox";
+
+  /**
    * For walking routing. This profile shows routes that are short and safe for cyclist, avoiding
    * highways and preferring streets with bike lanes.
    *
--- a/services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneService.java
+++ b/services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneService.java
@@ -18,46 +18,102 @@ public interface IsochroneService {
   /**
    * Constructs the HTTP request for the specified parameters.
    *
-   * @param profile A Mapbox Directions routing profile ID. Options are
-   *                {@link IsochroneCriteria#PROFILE_DRIVING} for travel times by car,
-   *                {@link IsochroneCriteria#PROFILE_WALKING} for pedestrian and hiking travel
-   *                times,and {@link IsochroneCriteria#PROFILE_CYCLING} for travel times by bicycle.
-   * @param coordinates    A {@link Point} object which represents a {longitude,latitude} coordinate
-   *                       pair around which to center the isochrone lines.
+   * @param profile         A Mapbox Directions routing profile ID. Options are
+   *                        {@link IsochroneCriteria#PROFILE_DRIVING} for travel
+   *                        times by car, {@link IsochroneCriteria#PROFILE_WALKING}
+   *                        for pedestrian and hiking travel times,and
+   *                        {@link IsochroneCriteria#PROFILE_CYCLING} for travel times
+   *                        by bicycle.
+   * @param coordinates     A {@link Point} object which represents a
+   *                        {longitude,latitude} coordinate pair around which to
+   *                        center the isochrone lines.
    * @param contoursMinutes A single String which has a comma-separated time in minutes to use for
    *                        each isochrone contour.
-   * @param accessToken    A valid Mapbox access token.
-   * @param contoursColors The colors to use for each isochrone contour, specified as hex values
-   *                       without a leading # (for example, ff0000 for red). If this parameter is
-   *                       used, there must be the same number of colors as there are entries in
-   *                       contours_minutes. If no colors are specified, the Isochrone API will
-   *                       assign a default rainbow color scheme to the output.
-   * @param polygons       Specify whether to return the contours as GeoJSON polygons (true) or
-   *                       linestrings (false, default). When polygons=true, any contour that
-   *                       forms a ring is returned as a polygon.
-   * @param denoise        A floating point value from 0.0 to 1.0 that can be used to remove
-   *                       smaller contours. The default is 1.0. A value of 1.0 will only
-   *                       return the largest contour for a given time value. A value of 0.5
-   *                       drops any contours that are less than half the area of the largest
-   *                       contour in the set of contours for that same time value.
-   * @param generalize     A positive floating point value in meters used as the tolerance for
-   *                       Douglas-Peucker generalization. There is no upper bound. If no value is
-   *                       specified in the request, the Isochrone API will choose the most
-   *                       optimized generalization to use for the request. Note that the
-   *                       generalization of contours can lead to self-intersections, as well
-   *                       as intersections of adjacent contours.
-   *
+   * @param accessToken     A valid Mapbox access token.
+   * @param contoursColors  The colors to use for each isochrone contour, specified as hex values
+   *                        without a leading # (for example, ff0000 for red). If this parameter is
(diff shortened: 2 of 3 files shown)


langsmith:
Actually, I'm now seeing how other navigation API wrappers in the repo, handle the user agent. Will adjust the isochrone support accordingly

osana:
@langsmith Technically speaking this becomes a 5.0.0 release because the API changed and it is not backward compatible. How come we did not catch this on the test level? I guess because a response was mocked. Did your sample code work?

langsmith:
> I am surprised the tests did not catch this? I guess because a response was mocked. Did your sample code work?

Yea, sample code worked. I think we didn't catch this because I put the `/mapbox/` slug in the `Makefile` isochrone fixture URLs (https://github.com/mapbox/mapbox-java/blob/master/Makefile#L199).

>Also, I think the way to fix this without making a major release would be to add the new method (not modify the existing one). One old one can be deprecated and we will remove it when 5.0.0 is done.

Sounds 💯 . I've pushed and rebased this fix. Added a new `getCall()` method which has the `user` parameter. Also added needed `deprecated` annotations to the previous `getCall()` method.

langsmith:
Going to test this work by using a Java services SNAPSHOT in the demo app isochrone refactor pr https://github.com/mapbox/mapbox-android-demo/pull/1029

services-isochrone/src/main/java/com/mapbox/api/isochrone/MapboxIsochrone.java
```diff
@@ -145,6 +150,16 @@ public static Builder builder() {
...
     public abstract Builder accessToken(@NonNull String accessToken);
 
+    /**
+     * The username for the account that the Isochrone engine runs on. In most cases, this should
+     * always remain the default value of {@link IsochroneCriteria#PROFILE_DEFAULT_USER}.
+     *
+     * @param user a non-null string which will replace the default user used in the Isochrone
+     *             request
+     * @return this builder for chaining options together
+     * @since 2.1.0
```
osana: `@since` tag should be 4.7.0

services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneCriteria.java
```diff
@@ -13,6 +13,13 @@
 public class IsochroneCriteria {
 
 
+  /**
+   * Mapbox default username.
+   *
+   * @since 4.6.1
```
osana: should be 4.7.0

services-isochrone/src/main/java/com/mapbox/api/isochrone/MapboxIsochrone.java
langsmith: So weird. How'd that happen :brainBoom:

services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneCriteria.java
langsmith: Fixed

services-isochrone/src/main/java/com/mapbox/api/isochrone/MapboxIsochrone.java
langsmith: Fixed

services-isochrone/src/main/java/com/mapbox/api/isochrone/IsochroneService.java
```diff
@@ -18,46 +18,102 @@
...
+   *                        specified in the request, the Isochrone API will choose the most
+   *                        optimized generalization to use for the request. Note that the
+   *                        generalization of contours can lead to self-intersections, as well
+   *                        as intersections of adjacent contours.
+   * @return a {@link FeatureCollection} in a Call wrapper
+   * @since 4.7.0
+   */
+  @GET("/isochrone/v1/{user}/{profile}/{coordinates}")
+  Call<FeatureCollection> getCall(
+    @Path("profile") String profile,
```
osana: NIT: could we have the order as in the path: user, profile, coordinate

langsmith: Fixed this, pushed, and rebased

PR #1006, opened 2019-04-17
Adding BoundingBoxPolygon turf method

langsmith:
This pr ports the bboxPolygon method to Turf for Java.

http://turfjs.org/docs/#bboxPolygon

https://github.com/Turfjs/turf/blob/master/packages/turf-bbox-polygon/index.ts 

`bboxPolygon(BoundingBox boundingBox)` is now a new method in the `TurfMeasurement` class

Used the geojson.io website during my work to test that the `Polygon` object returned by `bboxPolygon(BoundingBox boundingBox)`, is valid and correct:

Screen Shot 2019-04-17 at 9 45 54 AM
Screen Shot 2019-04-17 at 9 46 02 AM

Screen Shot 2019-04-17 at 10 08 10 AM
Screen Shot 2019-04-17 at 10 07 53 AM


services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -395,4 +396,46 @@ public static Point along(@NonNull LineString line, @FloatRange(from = 0) double
...
+    double west = boundingBox.west();
+    double south = boundingBox.south();
+    double east = boundingBox.east();
+    double north = boundingBox.north();
+
+    double[] lowLeft = new double[]{south,west};
+    double[] topLeft = new double[]{north,west};
+    double[] topRight = new double[]{north,east};
+    double[] lowRight = new double[]{south,east};
+
```
osana: do you really need these double[] arrays? I think you can just use `west, south, east, north` .

langsmith: Don't need them. Originally had them to match the turf.js version for continuity. Switched to your suggestion of `boundingBox.north()`, `boundingBox.east()`, etc.

Just pushed and rebased

services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -395,4 +396,47 @@ public static Point along(@NonNull LineString line, @FloatRange(from = 0) double
...
+  /**
+   * Takes a {@link BoundingBox} and uses its coordinates to create a {@link Polygon}
+   * geometry.
+   *
+   * @param boundingBox a {@link BoundingBox} object to calculate with
+   * @return a {@link Polygon} object
+   * @see <a href="http://turfjs.org/docs/#bboxPolygon">Turf BoundingBox Polygon documentation</a>
+   * @since 4.7.0
+   */
+  public static Polygon bboxPolygon(BoundingBox boundingBox) {
```
osana: Here are a couple of other observations sorry that I did not notice during the last review:

[ArrayList() Constructs an empty list with an initial capacity of ten](https://docs.oracle.com/javase/8/docs/api/java/util/ArrayList.html#ArrayList--
).
So when we are doing a library implementation it is best to take these things into account.  I think using Arrays.asList() could be one of the solutions:
    return Polygon.fromLngLats(
                   Arrays.asList(
                     Arrays.asList( Point.fromLngLat(boundingBox.south(), boundingBox.west()),
                                             Point.fromLngLat(boundingBox.south(), boundingBox.east()),
                                             Point.fromLngLat(boundingBox.north(), boundingBox.east()),
                                            Point.fromLngLat(boundingBox.north(), boundingBox.west()),
                                            Point.fromLngLat(boundingBox.south(), boundingBox.west()))));

Now ... as I typed this. I noticed that latitudes are passed into Points before longitudes but it should be the other way around.  Strange that the tests did not catch that..

langsmith: Just pushed and re-based.

This is working for me:
```java
public static Polygon bboxPolygon(BoundingBox boundingBox) {
    return Polygon.fromLngLats(
        Arrays.asList(
            Arrays.asList(
                Point.fromLngLat(boundingBox.south(), boundingBox.west()),
                Point.fromLngLat(boundingBox.south(), boundingBox.east()),
                Point.fromLngLat(boundingBox.north(), boundingBox.east()),
                Point.fromLngLat(boundingBox.north(), boundingBox.west()),
                Point.fromLngLat(boundingBox.south(), boundingBox.west()))));
  }
```

langsmith: I see what you mean now @osana . I switched the coordinates.

```java
public static Polygon bboxPolygon(BoundingBox boundingBox) {
    return Polygon.fromLngLats(
      Arrays.asList(
        Arrays.asList(
          Point.fromLngLat(boundingBox.west(), boundingBox.south()),
          Point.fromLngLat(boundingBox.east(), boundingBox.south()),
          Point.fromLngLat(boundingBox.east(), boundingBox.north()),
          Point.fromLngLat(boundingBox.west(), boundingBox.north()),
          Point.fromLngLat(boundingBox.west(), boundingBox.south()))));
  }
```

I also adjusted the way that I build a `BoundingBox` object in my tests. Tests are passing.

Pushed and rebased.

PR #1007, opened 2019-04-19
Adding explode turf method

langsmith:
This pr ports the explode method to Turf for Java. This method takes a `Feature` or `FeatureCollection` and returns a `FeatureCollection` of `Point` objects which represent all of the positions in that original `Feature`/`FeatureCollection` parameter.

http://turfjs.org/docs/#explode

https://github.com/Turfjs/turf/blob/master/packages/turf-explode/index.js

`explode(FeatureCollection featureCollection)` & `explode(Feature feature)` are the new methods added in the `TurfConversion` class.

Screen Shot 2019-04-19 at 11 05 16 AM


langsmith:
@osana , for some reason, the test wouldn't accept my fixture for `Point` and `Polygon`. The GeoJSON is valid if you check it with geojson.io. 

That's why I'm building those object in a different way for now:
- https://github.com/mapbox/mapbox-java/pull/1007/files#diff-c40c369d4646a12fa9714fcb02910865R85
- https://github.com/mapbox/mapbox-java/pull/1007/files#diff-c40c369d4646a12fa9714fcb02910865R103

Is there something I'm doing wrong, which is preventing me from doing `Polygon polygon = Polygon.fromJson(loadJsonFixture(TURF_EXPLODE_POLYGON));` ?

Also, https://github.com/mapbox/mapbox-java/pull/1007/files#diff-b6a32aa24e73a2791a14203a75659fce is in this pr because it was invalid GeoJSON. It needed `properties`. Not sure how the `bbox` tests ✅ with this error before adding this `properties` field 🤔

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
```diff
@@ -163,4 +176,83 @@ public static double convertLength(@FloatRange(from = 0) double distance,
...
+  /**
+   * Takes a {@link Geometry} object, checks what type of geometry it is,
+   * and then returns all positions as {@link Point} objects.
+   *
+   * @param geometry a {@link Geometry} object
+   * @return a new {@link FeatureCollection} object with {@link Point} objects
+   * @since 4.7.0
+   */
+  private static FeatureCollection explode(@NonNull Geometry geometry) {
+    List<Feature> featuresForFinalFeatureCollection = new ArrayList<>();
```
osana: Looks like there is a method: 
https://turfjs.org/docs/#coordAll  
that is supposed to work for FeatureCollection/Feature and Geometry.
We ported only the Geometry parameter. 
What do you think if we port coordAll(FeatureCollection) and coordAll(Feature) and then take advantage of it here.

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
```diff
@@ -163,4 +176,83 @@ public static double convertLength(@FloatRange(from = 0) double distance,
...
+   * returns all positions as {@link Point} objects.
+   *
+   * @param featureCollection a {@link FeatureCollection} object
+   * @return a new {@link FeatureCollection} object with {@link Point} objects
+   * @since 4.7.0
+   */
+  public static FeatureCollection explode(@NonNull FeatureCollection featureCollection) {
+    List<Feature> featuresForFinalFeatureCollection = new ArrayList<>();
+    for (Feature singleFeature : featureCollection.features()) {
+      for (int i = 0; i < explode(singleFeature.geometry()).features().size(); i++) {
```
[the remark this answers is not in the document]
langsmith: Ok, I refactored and was able to condense it down to:

```java
  public static FeatureCollection explode(@NonNull FeatureCollection featureCollection) {
    List<Feature> featuresForFinalFeatureCollection = new ArrayList<>();
    for (Feature singleFeature : featureCollection.features()) {
      featuresForFinalFeatureCollection.addAll(explode(singleFeature.geometry()).features());
    }
    return FeatureCollection.fromFeatures(featuresForFinalFeatureCollection);
  }
```

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
langsmith: >What do you think if we port coordAll(FeatureCollection) and coordAll(Feature) and then take advantage of it here.

Yea, I think we should do this. I've needed it a bunch during my work on porting more methods to Turf for Java.

langsmith: >Yea, I think we should do this. 

https://github.com/mapbox/mapbox-java/pull/1009

langsmith: @osana , just refactored and rebased this pr. Added usage of the new `coordAll()` methods, which were added in https://github.com/mapbox/mapbox-java/pull/1009.

PR #1008, opened 2019-04-20
Adding combine turf method

langsmith:
This pr ports the combine method to Turf for Java. This method takes a `FeatureCollection` of `Point`, `LineString`, or `Polygon` features and returns a `Geometry`. The returned `Geometry` can be used to create a `MultiPoint`, `MultiLineString`, or `MultiPolygon` feature.

http://turfjs.org/docs/#combine

https://github.com/Turfjs/turf/blob/master/packages/turf-combine/index.ts

`combine(FeatureCollection featureCollection)` is the new method added to the `TurfConversion` class.

Screen Shot 2019-04-19 at 9 05 10 PM

Screen Shot 2019-04-20 at 6 53 10 AM

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
@@ -10,6 +10,7 @@ import com.mapbox.geojson.FeatureCollection;
 import com.mapbox.geojson.Geometry;
 import com.mapbox.geojson.LineString;
 import com.mapbox.geojson.MultiLineString;
+import com.mapbox.geojson.MultiPoint;
 import com.mapbox.geojson.MultiPolygon;
 import com.mapbox.geojson.Point;
 import com.mapbox.geojson.Polygon;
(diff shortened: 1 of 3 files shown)


langsmith:
>The returned `Geometry` can be used to create a `MultiPoint`, `MultiLineString`, or `MultiPolygon` feature.

For example,

```java
 MultiLineString newMultiLineStringObject = (MultiLineString) TurfConversion.combine(lineStringFeatureCollection);
```

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
```diff
@@ -163,4 +175,58 @@ public static double convertLength(@FloatRange(from = 0) double distance,
...
+   * or {@link MultiPolygon}.
+   *
+   * @param featureCollection a {@link FeatureCollection} that has {@link Feature}
+   *                          objects which are all {@link Point}, {@link LineString},
+   *                          or {@link Polygon} geometries.
+   * @return a Geometry which can be used to cast {@link MultiPoint},
+   * {@link MultiLineString}, or {@link MultiPolygon}
+   * @since 4.7.0
+   */
+  public static Geometry combine(@NonNull FeatureCollection featureCollection) {
```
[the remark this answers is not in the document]
langsmith: I think the code above won't work because `LineString` can't be cast to a `Point`          (`pointList.add((G)singleFeature.geometry());`) and other similar situations.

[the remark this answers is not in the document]
tobrun: http://turfjs.org/docs/#combine

> Returns FeatureCollection <(MultiPoint|MultiLineString|MultiPolygon)> - a FeatureCollection of corresponding type to input

Return type should be a FeatureCollection

langsmith:
>Return type should be a FeatureCollection

Fixed

langsmith:
Ok, `FeatureCollection` is now returned in all scenarios and I pushed some refactoring which checks that all geometries are of the same type. If not, https://github.com/mapbox/mapbox-java/blob/c3524eef9a58c6e5be81b14dd0fe5fc8c6b48085/services-turf/src/test/java/com/mapbox/turf/TurfConversionTest.java#L185 is thrown.

langsmith:
👉👉@tobrun

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
```diff
@@ -200,5 +206,63 @@ public static FeatureCollection explode(@NonNull Feature feature) {
...
+        }
+      }
+      return FeatureCollection.fromFeature(Feature.fromGeometry(
+        MultiLineString.fromLineStrings(lineStringList)
+      ));
+    }
+    throw new TurfException("Your FeatureCollection did not include "
+      + "a Point, LineString, or Polygon geometry type. Please pass in "
+      + "a FeatureCollection with two or more Features which are all of "
+      + "the same geometry type.");
```
tobrun: Above block doesn't match the js one in https://github.com/Turfjs/turf/blob/master/packages/turf-combine/index.ts. 

> Support multiple geometry types

We need to be able to support multiple types of geometry in one resulting feature collection. For example if you would combine FeatureCollection of Points and FeatureCollection of LineString in your input. You would get a FeatureCollection with of 2 features (1 multipoint and 1 multilinestring) as output. 


> Support all input

We are currently taking the 0 index of the input, we need to use all input.  

```
 featureCollection.features().get(0).geometry();
```

services-turf/src/test/java/com/mapbox/turf/TurfConversionTest.java
```diff
@@ -78,6 +80,129 @@ public void convertDistance() throws TurfException {
...
+              Point.fromLngLat(7.91015625, -9.102096738726443),
+              Point.fromLngLat(31.113281249999996, 17.644022027872726),
+              Point.fromLngLat(11.42578125, 16.636191878397664)
+            ))))
+        });
+
+    // Building a geometry with this FeatureCollection should through an error
+    TurfConversion.combine(pointAndPolygonFeatureCollection);
+  }
+
```
tobrun: tests look good but we also want to port the js tests as closely as possible:
https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js
Porting these is helpful as they act as integration tests that validate our implementation.

langsmith: Latest commit adds and refactors tests to match https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js#L154

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
langsmith: > We need to be able to support multiple types of geometry in one resulting feature collection

Addressed in my latest commit. 

> For example if you would combine FeatureCollection of Points and FeatureCollection of LineString in your input. You would get a FeatureCollection with of 2 features (1 multipoint and 1 multilinestring) as output

My understanding of this method is a little different than ⬆️ ; based on http://turfjs.org/docs/#combine and the test lines such as https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js#L24 and https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js#L142.

A single `FeatureCollection` is the parameter. The `FeatureCollection` has to have `Point` and/or `MultiPoint`, `LineString` and/or `MultiLineString`, or `Polygon` and/or `MultiPolygon`. No mixing `Point`-based geometries with `Polygon`-based, etc.

The returned `FeatureCollection` only has _one_ `Feature` in it and the single `Feature` is always of a "Multi-" geometry.

I basically explain this at https://github.com/mapbox/mapbox-java/blob/b3b20132ec24b223c12aaac42e96e88332f72d2e/services-turf/src/main/java/com/mapbox/turf/TurfConversion.java#L213-L226

What's a little weird is that http://turfjs.org/docs/#combine shows that non-Multi geometries are the only types that can be in the parameter `FeatureCollection`. Whereas, the tests allow for "Multi-" geometries (https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js#L142 and https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js#L69). My latest commit follows the tests so that "Multi-" geometries are supported (e.g. `Point` and `MultiPoint` are in the parameter `FeatureCollection`).


> We are currently taking the 0 index of the input, we need to use all input.

Addressed in my latest commit.

langsmith:
Ok @tobrun . Pushed some changes and responded to comments above. Would love another round of eyes when you get a chance 🙇

langsmith:
👉👉 @tobrun 😬

tobrun:
> A single FeatureCollection is the parameter. The FeatureCollection has to have Point and/or MultiPoint, LineString and/or MultiLineString, or Polygon and/or MultiPolygon. No mixing Point-based geometries with Polygon-based, etc.


The [`heterogeneous`](https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js#L154) test showcase that is should be possible to combine both Point with Polygon. To get the proper implementation on the java side, I should try fully porting the testsuite on the js implementation. This helps to verify assumptions you are making about the code.

langsmith:
> The heterogeneous test showcase that is should be possible to combine both Point with Polygon. To get the proper implementation on the java side, I should try fully porting the testsuite on the js implementation. This helps to verify assumptions you are making about the code.

I'd missed the inclusion of the `Point` at https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js#L188 🔬 . Confirmed by getting more clarification at https://github.com/Turfjs/turf/issues/1756.

Refactored the method implementation and tests so that `heterogenous` combinations are allowed. Rebased. The order of my tests in this pr now match the test order in https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js. While the coordinates differ, my tests check the same situations as in https://github.com/Turfjs/turf/blob/master/packages/turf-combine/test.js (e.g. `Point` & `Point`, `LineString` & `MultiLineString`, etc). I also added `combinePointAndMultiPolygonAndLineStringGeometry`, `combinePointAndLineStringGeometry`, and `combine_featureCollectionSizeCheck`. 

Would love another round of 👀 when ya'll get a chance.

PR #1009, opened 2019-04-22
Adding coordAll() versions for Feature, FeatureCollection, and Geometry parameters

langsmith:
This pr adds methods for getting a `List<Point>` from a `Feature`, `FeatureCollection`, or `Geometry`. These parameters were missing from our Turf for Java library, but are part of https://turfjs.org/docs/#coordAll .

Having these methods will be incredibly helpful for adding a number of new Turf methods to our Turf for Java library. Also unblocks https://github.com/mapbox/mapbox-java/pull/1007.

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
@@ -3,12 +3,15 @@ package com.mapbox.turf;
 import android.support.annotation.NonNull;
 
 import com.mapbox.geojson.Feature;
+import com.mapbox.geojson.FeatureCollection;
+import com.mapbox.geojson.Geometry;
+import com.mapbox.geojson.GeometryCollection;
 import com.mapbox.geojson.LineString;
 import com.mapbox.geojson.MultiLineString;
 import com.mapbox.geojson.MultiPoint;
+import com.mapbox.geojson.MultiPolygon;
 import com.mapbox.geojson.Point;
 import com.mapbox.geojson.Polygon;
-import com.mapbox.geojson.MultiPolygon;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -37,24 +40,20 @@ public final class TurfMeta {
    */
   @NonNull
   public static List<Point> coordAll(@NonNull Point point) {
-    List<Point> coords = new ArrayList<>();
-    coords.add(point);
-    return coords;
+    return coordAll(new ArrayList<Point>(), point);
   }
 
   /**
-   * Get all coordinates from a {@link LineString} object, returning a {@code List} of Point
-   * objects. If you have a geometry collection, you need to break it down to individual geometry
-   * objects before using {@link #coordAll}.
+   * Private helper method to go with {@link TurfMeta#coordAll(Point)}.
    *
-   * @param lineString any {@link LineString} object
+   * @param coords the {@code List} of {@link Point}s.
+   * @param point  any {@link Point} object
    * @return a {@code List} made up of {@link Point}s
-   * @since 2.0.0
+   * @since 4.8.0
    */
   @NonNull
-  public static List<Point> coordAll(@NonNull LineString lineString) {
-    List<Point> coords = new ArrayList<>();
-    coords.addAll(lineString.coordinates());
+  private static List<Point> coordAll(@NonNull List<Point> coords, @NonNull Point point) {
+    coords.add(point);
     return coords;
   }
 
@@ -69,8 +68,48 @@ public final class TurfMeta {
    */
   @NonNull
   public static List<Point> coordAll(@NonNull MultiPoint multiPoint) {
-    List<Point> coords = new ArrayList<>();
+    return coordAll(new ArrayList<Point>(), multiPoint);
+  }
+
+  /**
+   * Private helper method to go with {@link TurfMeta#coordAll(MultiPoint)}.
+   *
+   * @param coords     the {@code List} of {@link Point}s.
+   * @param multiPoint any {@link MultiPoint} object
+   * @return a {@code List} made up of {@link Point}s
+   * @since 4.8.0
+   */
+  @NonNull
+  private static List<Point> coordAll(@NonNull List<Point> coords, @NonNull MultiPoint multiPoint) {
     coords.addAll(multiPoint.coordinates());
+    return coords;
+  }
+
+  /**
+   * Get all coordinates from a {@link LineString} object, returning a {@code List} of Point
+   * objects. If you have a geometry collection, you need to break it down to individual geometry
+   * objects before using {@link #coordAll}.
+   *
+   * @param lineString any {@link LineString} object
+   * @return a {@code List} made up of {@link Point}s
+   * @since 2.0.0
+   */
+  @NonNull
+  public static List<Point> coordAll(@NonNull LineString lineString) {
+    return coordAll(new ArrayList<Point>(), lineString);
+  }
+
+  /**
+   * Private helper method to go with {@link TurfMeta#coordAll(LineString)}.
+   *
+   * @param coords     the {@code List} of {@link Point}s.
+   * @param lineString any {@link LineString} object
+   * @return a {@code List} made up of {@link Point}s
+   * @since 4.8.0
+   */
+  @NonNull
+  private static List<Point> coordAll(@NonNull List<Point> coords, @NonNull LineString lineString) {
+    coords.addAll(lineString.coordinates());
     return coords;
   }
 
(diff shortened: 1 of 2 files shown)


services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
```diff
@@ -142,6 +145,76 @@ private TurfMeta() {
...
+  /**
+   * Get all coordinates from a {@link Feature} object, returning a {@link List} of {@link Point}
+   * objects.
+   *
+   * @param feature the {@link Feature} that you'd like to extract the Points from.
+   * @return a {@code List} made up of {@link Point}s
+   * @since 4.7.0
+   */
+  public static List<Point> coordAll(@NonNull Feature feature) {
+    return coordAll(FeatureCollection.fromFeature(feature));
```
osana: The library should not add overhead. No need to create FeatureCreation.

services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
```diff
@@ -142,6 +145,76 @@ private TurfMeta() {
...
+   *
+   * @param featureCollection the {@link FeatureCollection} that you'd like
+   *                          to extract the Points from.
+   * @return a {@code List} made up of {@link Point}s
+   * @since 4.7.0
+   */
+  public static List<Point> coordAll(@NonNull FeatureCollection featureCollection) {
+    List<Point> finalCoordsList = new ArrayList<>();
+    for (Feature singleFeature : featureCollection.features()) {
+      finalCoordsList.addAll(coordAll(singleFeature.geometry()));
```
osana: This will result in extra Lists creates (one per feature).
We could have a helper method that would add to existing list
```
private static List<Point>  addCoordAll(@NonNull List<Point> pointList, @NonNull Feature feature)
```
and
```
public static List<Point> coordAll(@NonNull Feature feature) {
    return addCoordAll(new ArrayList<Point>, @NonNull Feature feature);
}
```

services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
```diff
@@ -142,6 +145,81 @@ private TurfMeta() {
     return coords;
   }
 
+  /**
```
osana: Here is one way to approach the implementation and to avoid unneeded ArrayList() creattion.
We have public `coorAll` methods for all Geometries. We can add private versions that take List<Point> as an extra parameter.  See the example below:

```
public static List<Point> coordAll(@NonNull Polygon polygon, boolean excludeWrapCoord) {
     return coordAll(new ArrayList<Point>(), polygon, excludeWrapCoord);
  }

  private  static List<Point> coordAll(@NonNull List<Point> coords,
                                       @NonNull Polygon polygon, boolean excludeWrapCoord) {
    int wrapShrink = excludeWrapCoord ? 1 : 0;

    for (int i = 0; i < polygon.coordinates().size(); i++) {
      for (int j = 0; j < polygon.coordinates().get(i).size() - wrapShrink; j++) {
        coords.add(polygon.coordinates().get(i).get(j));
      }
    }
    return coords;
  }
```

services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
```diff
@@ -142,6 +145,81 @@ private TurfMeta() {
...
+   *
+   * @param featureCollection the {@link FeatureCollection} that you'd like
+   *                          to extract the Points from.
+   * @return a {@code List} made up of {@link Point}s
+   * @since 4.7.0
+   */
+  public static List<Point> coordAll(@NonNull FeatureCollection featureCollection) {
+    List<Point> finalCoordsList = new ArrayList<>();
+    for (Feature singleFeature : featureCollection.features()) {
+      finalCoordsList.addAll(addCoordAll(finalCoordsList, singleFeature));
```
osana: It should be just:
```      addCoordAll(finalCoordsList, singleFeature);```

File: services-turf/src/main/java/com/mapbox/turf/TurfMeta.java [hunk 5]
```diff
@@ -142,6 +145,81 @@ private TurfMeta() {
...
+
+  /**
+   * Get all coordinates from a {@link Geometry} object, returning a {@link List} of {@link Point}
+   * objects.
+   *
+   * @param geometry the {@link Geometry} object to extract the {@link Point}s from
+   * @return a {@code List} made up of {@link Point}s
+   * @since 4.7.0
+   */
+  private static List<Point> coordAllFromSingleGeometry(@NonNull Geometry geometry) {
```
osana: Add `List<Point> to the list of parameters, then it will look something like:
```
  private static List<Point> coordAll(@NonNull List<Point> pointList, @NonNull Geometry geometry) {
    if (geometry instanceof Point) {
      pointList.add((Point) geometry);

    } else if (geometry instanceof MultiPoint) {
      pointList.addAll(((MultiPoint)geometry).coordinates());

    } else if (geometry instanceof LineString) {
      pointList.addAll(((LineString)geometry).coordinates());

    } else if (geometry instanceof MultiLineString) {
      coordAll(pointList, (MultiLineString) geometry);

    } else if (geometry instanceof Polygon) {
      coordAll(pointList, (Polygon) geometry);

    } else if (geometry instanceof MultiPolygon) {
      coordAll(pointList, (MultiPolygon) geometry);

    } else if (geometry instanceof GeometryCollection) {
      // recursive
      for (Geometry singleGeometry : ((GeometryCollection) geometry).geometries()) {
        coordAll(pointList, singleGeometry);
      }
    }
    return pointList;
  }
```

services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
```diff
@@ -142,6 +145,81 @@ private TurfMeta() {
...
+  public static List<Point> coordAll(@NonNull FeatureCollection featureCollection) {
+    List<Point> finalCoordsList = new ArrayList<>();
+    for (Feature singleFeature : featureCollection.features()) {
+      finalCoordsList.addAll(addCoordAll(finalCoordsList, singleFeature));
+    }
+    return finalCoordsList;
+  }
+
+  private static List<Point> addCoordAll(@NonNull List<Point> pointList, @NonNull Feature feature) {
+    pointList.addAll(coordAllFromSingleGeometry(feature.geometry()));
```
osana: This should be :
```return coordAll(pointList, feature.geometry());```

File: services-turf/src/main/java/com/mapbox/turf/TurfMeta.java (hunk 3 above)
Comment (langsmith) [reply services-turf/src/main/java/com/mapbox/turf/TurfMeta.java]: Ahhhh, I see now. Got it. 

The changes above would break semver, correct? Or to preserve, could we just deprecate the current `public static List<Point> coordAll(@NonNull Polygon polygon, boolean excludeWrapCoord) {` and create a new `coordAll(@NonNull Polygon polygon, boolean excludeWrapCoord)` as you've done above?

**Comment by langsmith** [comment]
Ok @osana , just pushed two commits that address your feedback above. Mainly added the code that you suggested above. I had to add a `@NonNull boolean excludeWrapCoord` parameter to the `Feature` & `FeatureCollection` methods because the boolean is needed in 

```java
 else if (geometry instanceof Polygon) {
      coordAll(pointList, (Polygon) geometry, excludeWrapCoord);

    } else if (geometry instanceof MultiPolygon) {
      coordAll(pointList, (MultiPolygon) geometry, excludeWrapCoord);

    } else if (geometry instanceof GeometryCollection) {
      // recursive
      for (Geometry singleGeometry : ((GeometryCollection) geometry).geometries()) {
        coordAllFromSingleGeometry(pointList, singleGeometry, excludeWrapCoord);
      }
    }
```

Anyways. Let me know how things look now.

services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
osana: We are actually not breaking semver-major as only `private static` methods are added. It is a semver-minor change as new APIs are added.

langsmith:
Ok @osana . Just pushed and rebased. Added the `@NonNull` annotations where needed and also adjusted mentions of `@since 4.7.0` to `@since 4.8.0`.

services-turf/src/main/java/com/mapbox/turf/TurfMeta.java
```diff
@@ -143,7 +229,91 @@ private TurfMeta() {
...
+  public static List<Point> coordAll(@NonNull FeatureCollection featureCollection,
+                                     @NonNull boolean excludeWrapCoord) {
+    List<Point> finalCoordsList = new ArrayList<>();
+    for (Feature singleFeature : featureCollection.features()) {
+      addCoordAll(finalCoordsList, singleFeature, excludeWrapCoord);
+    }
+    return finalCoordsList;
+  }
+
+  private static List<Point> addCoordAll(@NonNull List<Point> pointList, @NonNull Feature feature,
```
osana: @NonNull

PR #1012, opened 2019-04-24
Fix possible IndexOutOfBoundsException exception

zhao-gang:
Check if list is empty before getting its items. Also add an
annotation to indicate that steps parameter should be greater than 0.


osana:
@zhao-gang Thank you for contributing.
This PR was merged via #1013

PR #1019, opened 2019-05-01
TurfMeasurement#square

tobrun:
Port of https://turfjs.org/docs/#square

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
@@ -501,4 +501,41 @@ public final class TurfMeasurement {
   public static Polygon envelope(GeoJson geoJson) {
     return bboxPolygon(bbox(geoJson));
   }
+
+
+  /**
+   * Takes a bounding box and calculates the minimum square bounding box
+   * that would contain the input.
+   *
+   * @param boundingBox extent in west, south, east, north order
+   * @return a square surrounding bbox
+   * @since 4.9.0
+   */
+  public static BoundingBox square(@NonNull BoundingBox boundingBox) {
+    double horizontalDistance = distance(boundingBox.southwest(),
+            Point.fromLngLat(boundingBox.east(), boundingBox.south())
+    );
+    double verticalDistance = distance(
+            Point.fromLngLat(boundingBox.west(), boundingBox.south()),
+            Point.fromLngLat(boundingBox.west(), boundingBox.north())
+    );
+
+    if (horizontalDistance >= verticalDistance) {
+      double verticalMidpoint = (boundingBox.south() + boundingBox.north()) / 2;
+      return BoundingBox.fromLngLats(
+              boundingBox.west(),
+              verticalMidpoint - ((boundingBox.east() - boundingBox.west()) / 2),
+              boundingBox.east(),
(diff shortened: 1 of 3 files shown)


services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -439,4 +439,40 @@ public static Polygon bboxPolygon(@NonNull BoundingBox boundingBox) {
...
                 Point.fromLngLat(boundingBox.west(), boundingBox.south()))));
   }
+
+
+  /**
+   * Takes a bounding box and calculates the minimum square bounding box that would contain the input.
+   *
+   * @param boundingBox extent in west, south, east, north order
+   * @return a square surrounding bbox
+   */
```
langsmith: missing `@since 4.8.0`

services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -439,4 +439,40 @@ public static Polygon bboxPolygon(@NonNull BoundingBox boundingBox) {
...
+  public static BoundingBox square(@NonNull BoundingBox boundingBox) {
+    double horizontalDistance = distance(
+            Point.fromLngLat(boundingBox.west(), boundingBox.south()),
```
osana: you can just use `boundingBox.southwest()` instead of creating a point

services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -439,4 +439,40 @@ public static Polygon bboxPolygon(@NonNull BoundingBox boundingBox) {
...
+            Point.fromLngLat(boundingBox.east(), boundingBox.south())
+    );
+    double verticalDistance = distance(
+            Point.fromLngLat(boundingBox.west(), boundingBox.south()),
+            Point.fromLngLat(boundingBox.west(), boundingBox.north())
+    );
+
+    if (horizontalDistance >= verticalDistance) {
+      double verticalMidpoint = (boundingBox.south() + boundingBox.north()) / 2;
+      return BoundingBox.fromCoordinates(
```
osana: Use `BoundingBox.fromLngLats` instead of `fromCoordinates`

services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -439,4 +439,40 @@ public static Polygon bboxPolygon(@NonNull BoundingBox boundingBox) {
...
+   * @return a square surrounding bbox
+   */
+  public static BoundingBox square(@NonNull BoundingBox boundingBox) {
+    double horizontalDistance = distance(
+            Point.fromLngLat(boundingBox.west(), boundingBox.south()),
+            Point.fromLngLat(boundingBox.east(), boundingBox.south())
+    );
+    double verticalDistance = distance(
+            Point.fromLngLat(boundingBox.west(), boundingBox.south()),
+            Point.fromLngLat(boundingBox.west(), boundingBox.north())
```
osana: you can just use boundingBox.northeast() instead of creating a point

tobrun: there is no northEast coordinate used

PR #1020, opened 2019-05-01
TurfMeasurement#envelope

tobrun:
This PR adds support for https://turfjs.org/docs/#envelope. It also fixes up bbox implementation by allowing any GeoJson as parameter and add support for feature and feature collection conversions. bboxPolygon has also been updated to allow for double[] which is the actual representation of bbox in turf vs the BoundingBox from geojson.

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
@@ -7,6 +7,9 @@ import android.support.annotation.FloatRange;
 import android.support.annotation.NonNull;
 
 import com.mapbox.geojson.BoundingBox;
+import com.mapbox.geojson.Feature;
+import com.mapbox.geojson.FeatureCollection;
+import com.mapbox.geojson.GeoJson;
 import com.mapbox.geojson.Geometry;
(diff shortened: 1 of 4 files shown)


services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -337,6 +340,57 @@ public static Point along(@NonNull LineString line, @FloatRange(from = 0) double
...
 
+  /**
+   * Takes a set of features, calculates the bbox of all input features, and returns a bounding box.
+   *
+   * @param geoJson a {@link GeoJson} object
+   * @return a double array defining the bounding box in this order {@code [minX, minY, maxX, maxY]}
+   * @since 4.8.0
+   */
+  public static double[] bbox(GeoJson geoJson) {
+    BoundingBox boundingBox = geoJson.bbox();
```
osana: One thing. I am not convinced that we can trust the bbox that came with the geoJson.

tobrun: capturing from the spec, this can be provided though seems optional. If bbox isn't implemented, it will try to base the bbox on the actual geometries.

PR #1025, opened 2019-05-08
support for multiple  languages in geocoding model objects

### LINKED ISSUE #446: Add geocoding API for setting multiple languages
Currently we only allow the setting of [one Language](https://github.com/mapbox/mapbox-java/blob/master/mapbox/libjava-services/src/main/java/com/mapbox/services/api/geocoding/v5/MapboxGeocoding.java#L553) for the geocoder. The Geocoding API will allow multiple Locales to be passed in. https://github.com/mapbox/MapboxGeocoder.swift/issues/104

Note this needs to be exposed in both `MapboxGeocoder` and Geocoding widget.

cc: @zugaldia @1ec5

osana:
This PR will resolve #446 

There were two parts to this PR:
- Geocoding module will not depend on AutoValue. This makes it easier to do json parsing customization. 
- Add custom parsing of fileds like: 
   text_{language} fields into a Map (lang -> text string in that language)

Added tests.


services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/GeocodingTypeAdapter.java
```diff
@@ -0,0 +1,120 @@
...
+  void writeString(@NonNull String name, String value,
+                           @NonNull JsonWriter jsonWriter) throws IOException {
+    jsonWriter.name(name);
+    if (value == null) {
+      jsonWriter.nullValue();
+    } else {
+      TypeAdapter<String> stringAdapter = this.stringAdapter;
+      if (stringAdapter == null) {
+        this.stringAdapter = stringAdapter;
+        stringAdapter = gson.getAdapter(String.class);
```
LukasPaczos: Looks like the adapter always uses `String.class`, does it have to be lazily initialized in each method instead once in the class field?

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/GeocodingTypeAdapter.java
```diff
@@ -0,0 +1,120 @@
...
+    TypeAdapter<String> stringAdapter = this.stringAdapter;
+    if (stringAdapter == null) {
+      this.stringAdapter = stringAdapter;
+      stringAdapter = gson.getAdapter(String.class);
+    }
+    return stringAdapter.read(jsonReader);
+  }
+
+  @Nullable
+  List<String> readLanguages(@Nullable List<String> languages,
```
LukasPaczos: The passed collections references seem to always be `null`s and are initialized only here, do we need to pass them in the first place then?

services-geojson/src/main/java/com/mapbox/geojson/Feature.java
```diff
@@ -48,7 +49,7 @@
  * @since 1.0.0
  */
 @Keep
-public final class Feature implements GeoJson {
+public final class Feature implements GeoJson, Serializable {
```
LukasPaczos: What's the advantage of making `Feature`/`FeatureCollection` serializable now? If we'd like to make them serializable in general, why not make `GeoJson` implement `Serializable`?

LukasPaczos:
Whoops, didn't mean to approve, only comment.

langsmith:
cc @kbauhaus as fyi that this is moving

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/GeocodingTypeAdapter.java
osana: I believe `readLanguages` will be called multiple times. In case of CarmenContext  json could look like:

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/GeocodingTypeAdapter.java
osana: This is the part that I was not sure about. 
The code is extracted from generated AutoValue.
There is an instance variable: stringAdapter which will be Lazily initialized.
I am not sure why there is a need to keep a local copy of `this. stringAdapter` 

Decided to keep it as is. I think I should address that part separately.

services-geojson/src/main/java/com/mapbox/geojson/Feature.java
osana: I noticed inconsistency that felt odd, so decided to add it there. I think you are right: we can just add Serializable to GeoJson. Will experiment

osana: I decide to remove these changes from this PR.
I think your suggestion is a valid one and should be addressed in a separate PR.

LukasPaczos:
I think this PR is functionally fine, but the style is off because it comes from generated code. We're going to need to maintain it manually now, so it should be cleaned up before it's merged to `master` in my opinion.

osana:
Thank you for 👀 @LukasPaczos 
Would love to hear your suggestions.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/GeocodingTypeAdapter.java
LukasPaczos: Will itt be called multiple times with the same list as an argument?

LukasPaczos:
I'm thinking about the left comments like unnecessary local object copies, lazy inits, or null checks.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/GeocodingTypeAdapter.java
osana: Yes. JsonReader is parsing each key/value as they come in. We cannot get all the `language_*`  pairs (and similarly for other pairs and JsonWriter). I believe TypeAdapters are more efficient then JsonSerializer/Deserializer as do not have an intermediate parsed state. Because of that though it is a bit more restrictive.

osana:
I lazy init for TypeAdapter<String> does not make sense - agree with you.
I do not understand why the generated code had a double assignment.
I can get rid of those and then see what happens.

I do think that some TypeAdapters should be lazily initialized as those fields are rarely used like foBoundingBox, etc.

Issue #1024, opened 2019-05-07
Release 4.8.0


osana:
### Pre-release checklist @osana 

- [x] update this repo's `README.md` file 
- [x] Change version name to `4.8.0-alpha.1` in `gradle.properties` (removing `-SNAPSHOT`)
- [x] Publish `4.8.0-alpha.1` artifact to Maven Central
- [x] Change version name back to having `-SNAPSHOT` at the end
- [x] Update `CHANGELOG.md`
- [x] Tag `v new release version` in GitHub

### Pre-release Testing
- [x]  Create  Maps SDK Android PR to update to `4.8.0-alpha.1` Maps Java SDK
- [ ] Create  Navigation Android PR to update to `4.8.0-alpha.1` Maps Java SDK
- [ ] Report any regressions on this ticket

### Pre-release checklist @osana 
- [x] Change version name to `4.8.0-alpha.2` in `gradle.properties` (removing `-SNAPSHOT`)
- [x] Publish `4.8.0-alpha.2` artifact to Maven Central
- [x] Change version name back to having `-SNAPSHOT` at the end
- [x] Update `CHANGELOG.md`
- [x] Tag `v new release version` in GitHub

### Pre-release Testing
- [x]  Create  Maps SDK Android PR to update to `4.8.0-alpha.2` Maps Java SDK
- [ ] Create  Navigation Android PR to update to `4.8.0-alpha.2` Maps Java SDK
- [ ] Report any regressions on this ticket

### Final 4.8.0 Release checklist @osana 
- [x] update this repo's `README.md` file 
- [x] Change version name to `4.8.0` in `gradle.properties` (removing `-SNAPSHOT`)
- [ ] Publish `4.8.0` artifact to Maven Central
- [x] Change version name back to having `-SNAPSHOT` at the end
- [ ] Update `CHANGELOG.md`
- [x] Tag `v new release version` in GitHub
- [x] Update dependencies in the Maps SDK to the final release (using above PR)
    
### Post Final release @langsmith   
- [x] Update the demo app
- [x] Update version # in config files in /help @langsmith
- [x] After updating the demo app: 
  - [x] Update Javadoc for the Mapbox Java docs section in `/android-docs` repo
[...]

PR #1030, opened 2019-05-10
Release 4.8.0

osana:
addresses #1024


services-core/gradle.properties
```diff
@@ -1,4 +1,4 @@
-VERSION_NAME=4.8.0-SNAPSHOT
+VERSION_NAME=4.9.0-SNAPSHOT
```
LukasPaczos: Just a note that since this version is always the same as the one in the root `gradle.properties`, it's probably safe to delete it.

PR #1033, opened 2019-05-17
Feature.id() can be String and Number

### LINKED ISSUE #914: Support feature id as a number
Capturing from [RFC](https://tools.ietf.org/html/rfc7946#page-6)

> If a Feature has a commonly used identifier, that identifier SHOULD be included as a member of the Feature object with the name "id", and the value of this member is **either a JSON string or number**.

On a public API level we only expose String.java:

```java
  /**
   * A feature may have a commonly used identifier which is either a unique String or number.
   *
   * @return a String containing this features unique identification or null if one wasn't given
   *   during creation.
   * @since 1.0.0
   */
  @Nullable
  public abstract String id();
```

osana:
According to GeoJson spec Feature.id can be String or Number.

This will be Semver-Major change as `id()` has to return `Object` to support both String and Number.

After #1025 lands, CarmenFeature.id() should be modified as well.

closes #914

### DIFF
--- a/services-geojson/src/main/java/com/mapbox/geojson/Feature.java
+++ b/services-geojson/src/main/java/com/mapbox/geojson/Feature.java
@@ -57,7 +57,7 @@ public final class Feature implements GeoJson {
   @JsonAdapter(BoundingBoxTypeAdapter.class)
   private final BoundingBox bbox;
 
-  private final String id;
+  private final Object id;
 
   private final Geometry geometry;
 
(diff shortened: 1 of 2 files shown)


services-geojson/src/main/java/com/mapbox/geojson/Feature.java
```diff
@@ -229,13 +264,14 @@ public BoundingBox bbox() {
    *
    * @return a String containing this features unique identification or null if one wasn't given
    *   during creation.
-   * @since 1.0.0
+   * @since 5.0.0
    */
   @Nullable
-  public String id() {
+  public Object id() {
```
tobrun: would be great to include as idAsString and idAsNumber facility methods  (if the type doesn't match throw an exception).

zugaldia:
Stale.

PR #1039, opened 2019-05-24
Add Interceptor support to Directions Refresh API and enable Debug logging

Guardiola31337:
- Adds `Interceptor` to set in the `OkHttp` client support to Directions Refresh API
- Enables Debug logging

### DIFF
--- a/services-directions-refresh/src/main/java/com/mapbox/api/directionsrefresh/v1/MapboxDirectionsRefresh.java
+++ b/services-directions-refresh/src/main/java/com/mapbox/api/directionsrefresh/v1/MapboxDirectionsRefresh.java
@@ -12,6 +12,9 @@ import com.mapbox.core.MapboxService;
 import com.mapbox.core.constants.Constants;
 import com.mapbox.core.utils.ApiCallHelper;
 
+import okhttp3.Interceptor;
+import okhttp3.OkHttpClient;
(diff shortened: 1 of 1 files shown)


services-directions-refresh/src/main/java/com/mapbox/api/directionsrefresh/v1/MapboxDirectionsRefresh.java
```diff
@@ -42,6 +45,25 @@ protected MapboxDirectionsRefresh() {
     );
   }
 
+  @Override
+  protected synchronized OkHttpClient getOkHttpClient() {
+    if (okHttpClient == null) {
```
devotaaabel: Would it be worth it to put this in `MapboxService` so we can use it from all services?

Guardiola31337: Yeah, that's correct. I also thought about this but would mean a bigger refactor because there are already some other services that accept `Interceptor`s.

Let's ticket this out and revisit later.

Guardiola31337: Follow up ticket 👉 https://github.com/mapbox/mapbox-java/issues/1041

PR #1056, opened 2019-06-24
Added toJson method to API service response classes (optimization, matrix, etc)

### LINKED ISSUE #814: Add toJson consistently to all API response objects
Some objects like the [geocoding response](https://github.com/mapbox/mapbox-java/blob/master/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/CarmenFeature.java#L286) include a `toJson` method which makes the response easily serializable and to log. However, other responses like [`DirectionsResponse`](https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/models/DirectionsResponse.java) do not include one.

Let's add a `toJson` method to all API response objects.

Related: https://github.com/mapbox/mapbox-java/issues/662.

langsmith:
This pr resolves #814 by adding `toJson()` to the API response classes that were missing the method.

### DIFF
--- a/services-directions-refresh/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshResponse.java
+++ b/services-directions-refresh/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshResponse.java
@@ -96,6 +96,22 @@ public abstract class DirectionsRefreshResponse extends DirectionsJsonObject {
   }
 
   /**
+   * Takes the currently defined values found inside this instance and converts it to a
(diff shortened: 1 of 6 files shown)


langsmith:
Thoughts on writing tests for this? Needed? Don't bother?

LukasPaczos:
> Don't bother?

`codecov` does not agree! :slightly_smiling_face: I'd love to see a test. You can manufacture a valid request in a JSON format, then prepare the same one using the APIs and match the `#toJson` output with the manually prepared JSON.

Issue #1053, opened 2019-06-19
sources.jar artifacts empty on Maven Central


gwimmel:
#### Expected Behavior

sources.jar on Maven Central (current version: [https://search.maven.org/remotecontent?filepath=com/mapbox/mapboxsdk/mapbox-sdk-turf/4.8.0/mapbox-sdk-turf-4.8.0-sources.jar](https://search.maven.org/remotecontent?filepath=com/mapbox/mapboxsdk/mapbox-sdk-turf/4.8.0/mapbox-sdk-turf-4.8.0-sources.jar) ) contains mapbox-sdk-turf sources.

#### Current Behavior
The above file seems to contain only empty folders. Therefore one cannot easily navigate to the mapbox-sdk-turf sources in one's IDE.

#### Possible Solution
If the current behaviour is not intentional, modify the release process for mapbox-sdk-turf such that a valid sources.jar file is generated and uploaded to Maven Central.

#### Steps to Reproduce

Download the sources.jar - file linked above and view the contents.

tobrun:
thank you for the report @gwimmel, confirmed that no java files are included.

LukasPaczos:
This is also the case for all artifacts, not only `turf`.

gwimmel:
I guess this is due to [https://github.com/mapbox/mapbox-java/commit/be2f75ce8506120d44b6f752a0a8650818e155c1#diff-c197962302397baf3a4cc36463dce5ea](https://github.com/mapbox/mapbox-java/commit/be2f75ce8506120d44b6f752a0a8650818e155c1#diff-c197962302397baf3a4cc36463dce5ea). The "**/*.java" - exclusion probably also applies to the sources jars, which it shouldn't.

gwimmel:
The pull reqest seems to be stale - should I open a new one with my suggestion?

tobrun:
@gwimmel I was trying out your proposed solution with a snapshot build but failed to get a build working (see [here](https://github.com/mapbox/mapbox-java/pull/1037#issuecomment-508393882) for context). Feel free to create a new pull request! You deserve the contributions on this one ;)

gwimmel:
Strange - you mean the deployment to jfrog does not work when my proposed solution is applied, but does work otherwise? The problem you cited seems unrelated to me. The sources.jar-Files are also generated/deployed in the current version, the only thing my fix should change is their content.

tobrun:
[...]

PR #1059, opened 2019-06-25
Removed exclusion of files that end in .java (fixing empty sources.jar)

langsmith:
Resolves https://github.com/mapbox/mapbox-java/issues/1053 and undos https://github.com/mapbox/mapbox-java/commit/be2f75ce8506120d44b6f752a0a8650818e155c1#diff-c197962302397baf3a4cc36463dce5ea

cc @gwimmel, who originally reported #1053

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,10 +86,6 @@ def RELEASE_MODULES = ["services",
 
 subprojects { subproject ->
 
-  tasks.withType(Jar) {
-    exclude "**/*.java"
-  }
-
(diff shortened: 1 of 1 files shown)


gwimmel:
maybe something like the following could be used instead?
```groovy
  tasks.withType(Jar) { jarTask ->
    if (!jarTask.name.endsWith("SourcesJar")) {
      jarTask.exclude("**/*.java")
    }
  }
```

tobrun:
fixed with https://github.com/mapbox/mapbox-java/pull/1064

PR #1062, opened 2019-07-02
Add an API to check if a custom shifter is set.

### LINKED ISSUE #1057: Add an option to check if a custom shifter is set
We should provide a `CoordinateShifterManager#isDefault` method to check if a custom shifter is set and skip shifting if not, to decrease the memory overhead associated with creating coordinate lists.

/cc @Chaoba

Chaoba:
Resolves #1057

### DIFF
--- a/services-geojson/src/main/java/com/mapbox/geojson/shifter/CoordinateShifterManager.java
+++ b/services-geojson/src/main/java/com/mapbox/geojson/shifter/CoordinateShifterManager.java
@@ -58,4 +58,12 @@ public final class CoordinateShifterManager {
     CoordinateShifterManager.coordinateShifter =
       coordinateShifter == null ? DEFAULT : coordinateShifter;
   }
+
+  /**
+   * Check whether the current shifter is the default one.
+   * @return true if using default shifter.
+   */
+  public static boolean isUsingDefaultShifter() {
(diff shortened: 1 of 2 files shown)


services-geojson/src/main/java/com/mapbox/geojson/shifter/CoordinateShifterManager.java
```diff
@@ -58,4 +58,12 @@ public static void setCoordinateShifter(CoordinateShifter coordinateShifter) {
     CoordinateShifterManager.coordinateShifter =
       coordinateShifter == null ? DEFAULT : coordinateShifter;
   }
+
+  /**
+   * Check whether the current shifter is the default one.
+   * @return true if using default shifter.
+   */
+  public static boolean isUsingDefaultShifter() {
+    return coordinateShifter == DEFAULT;
```
tobrun: this should be `equals` instead of `==`

Chaoba: It seems use `==` doesn't have any issue here as `coordinateShifter` will be the same object of  `DEFAULT` when using default shifter. Do you have any other reasons for using `equals` here?

PR #1074, opened 2019-09-17
[geocoding] Add intersection search support.

pengdev:
Resolves #1050 by adding `intersectionStreets(street1, street2)` method to the `MapboxGeocoding` builder.

PR #1075, opened 2019-09-17
[turf] Add polygonToLine method.

### LINKED ISSUE #1072: Turf: add polygonToLine method
This is a feature request to add http://turfjs.org/docs/#polygonToLine to this library. This method converts a Polygon to (Multi)LineString or MultiPolygon to a FeatureCollection of (Multi)LineString.

pengdev:
Resolves #1072 by adding the `polygonToLine` methods to `TurfConversion`.


pengdev:
The unit tests are failing because in the TurfJS, the `properties` value of a `Feature` can be empty, but in the `Feature` java class, an empty properties field is ignored: https://github.com/mapbox/mapbox-java/blob/2130ea708c72ef8b98aeb02102554d7dc32606a0/services-geojson/src/main/java/com/mapbox/geojson/Feature.java#L280

Shall we follow exactly TurfJS's implementation or shall we modify our test case to ignore the difference?

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
```diff
@@ -201,4 +209,143 @@ public static FeatureCollection explode(@NonNull Feature feature) {
     }
     return FeatureCollection.fromFeatures(finalFeatureList);
   }
+
+
```
langsmith: nit: remove blank line

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
```diff
@@ -201,4 +209,143 @@ public static FeatureCollection explode(@NonNull Feature feature) {
...
+  /**
+   * Takes a {@link Feature} that contains {@link Polygon} and
+   * covert it to a {@link Feature} that contains {@link LineString} or {@link MultiLineString}.
```
langsmith:
```suggestion
   * covert it to a {@link Feature} that is of a {@link LineString} or {@link MultiLineString} geometry.
```

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
```diff
@@ -201,4 +209,143 @@ public static FeatureCollection explode(@NonNull Feature feature) {
     }
     return FeatureCollection.fromFeatures(finalFeatureList);
   }
+
+
+  /**
+   * Takes a {@link Feature} that contains {@link Polygon} and
```
langsmith:
```suggestion
   * Takes a {@link Feature} that is of a {@link Polygon} geometry and
```

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
langsmith: actually, nevermind. What you had is fine.

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
langsmith: actually, nevermind. What you had is fine.

services-turf/src/main/java/com/mapbox/turf/TurfConversion.java
pengdev: sure, will do it.

tobrun:
> Shall we follow exactly TurfJS's implementation or shall we modify our test case to ignore the difference? @mapbox/maps-android

this is a implementation detail of geojson in java vs in js. Feel free to ignore this part.

pengdev:
> > Shall we follow exactly TurfJS's implementation or shall we modify our test case to ignore the difference? @mapbox/maps-android
> 
> this is a implementation detail of geojson in java vs in js. Feel free to ignore this part.

@tobrun Thanks! And I noticed during the conversion, the precision of coordinates has been reduced, i.e. from `102.227783203125` to `102.2277832`. Is this the intended behaviour?

tobrun:
> @tobrun Thanks! And I noticed during the conversion, the precision of coordinates has been reduced, i.e. from 102.227783203125 to 102.2277832. Is this the intended behaviour?

This is also an implementation detail difference between geojson js vs geojson java.. would you be able to ticket those two items out so we are tracking them? I'm have been planning a geojson rewrite for a while and we should address at that time. For now try matching the test as much as possible.

pengdev:
> > @tobrun Thanks! And I noticed during the conversion, the precision of coordinates has been reduced, i.e. from 102.227783203125 to 102.2277832. Is this the intended behaviour?
> 
> This is also an implementation detail difference between geojson js vs geojson java.. would you be able to ticket those two items out so we are tracking them? I'm have been planning a geojson rewrite for a while and we should address at that time. For now try matching the test as much as possible.

I found the precision reduction was introduced in https://github.com/mapbox/mapbox-java/pull/938 . It seems the precision is trimed due to performance consideration.

pengdev:
@tobrun I have fixed the unit test by parsing the output geojson to `Feature` object and then convert it to json string again. This way the output geojson is processed by our library as well.  I also updated change log and docs.

pengdev:
@tobrun I have used since `4.10.0` inside the java doc. And the same goes to PR #1074

langsmith:
I just created a `4.10.0` milestone and have switched this ticket to it

langsmith:
Gahh, too many various SDKs version in my mind. 4.9.0 will be the next stable version of the Java SDK. I've moved this pr and https://github.com/mapbox/mapbox-java/pull/1074 to the `4.9.0` milestone. Deleted the `4.10.0` milestone. @pengdev , I will make a new pr that adjusts your `@since` annotations in this pr and https://github.com/mapbox/mapbox-java/pull/1074 to `4.9.0` because I incorrectly advised you to use `4.10.0` 🤦‍♂ .

langsmith:
Fix pr at https://github.com/mapbox/mapbox-java/pull/1078

PR #1092, opened 2019-12-17
Replaced NotNull with NonNull annotation in TurfConversion class

langsmith:
This pr replaces `NotNull` and its import line with `NonNull`. This fixes a bug that has shown up downstream in many of Mapbox's own projects as well as our customers' work.


langsmith:
cc @Guardiola31337 , who has dealt with this `NotNull` import line in the past

langsmith:
Would this be considered a SEMVER major change?

tobrun:
> Would this be considered a SEMVER major change?

I think we can get away that using that NotNull annotation was a bug

langsmith:
@chloekraw , heads up that this is now merged

PR #1107, opened 2020-01-22
Extract refresh models into separate module

evabishchevich:
This PR introduces one more module with DTOs for Direction Refresh API


gradle.properties
```diff
@@ -1,5 +1,5 @@
 # Project-wide Gradle settings.
-VERSION_NAME=5.1.0-SNAPSHOT
+VERSION_NAME=5.2.0-SNAPSHOT
```
LukasPaczos: Are there any API changes?

gradle/checkstyle.gradle
```diff
@@ -3,6 +3,7 @@ apply plugin: 'checkstyle'
 checkstyle.toolVersion = '7.8.2'
 checkstyle.configFile = rootProject.file('config/checkstyle/checkstyle.xml')
 checkstyleMain.exclude '**/BuildConfig.java'
+checkstyleTest.enabled = false
```
LukasPaczos: We can still check the code style to keep ourselves in check :)

evabishchevich: It's only for test code. On the other hand it was failing even before this PR because tests don't have any javadoc on public methods and so on.

gradle.properties
evabishchevich: Nope, just don't want to mix it with previous version. 
Should we make it 5.1.1?

LukasPaczos: `5.1.0` has not been released yet, so we can keep using that snapshot version.

evabishchevich:
> Are there any duplicated auto value classes generated when building the `/services` module?
> See
> https://github.com/mapbox/mapbox-java/blob/1a932c14f94305c69d92f8efb6b302cbdd8e7b14/services/build.gradle#L25-L32

There is no such a problem. Checked in mavenLocal

gradle.properties
evabishchevich: yeah, but we may face some caching problems because of the same naming. I know that it's still 'SNAPSHOT'. 
Do you suggest leaving  5.1.0 as it is?

LukasPaczos: yes, we should leave it at `5.1.0`

services/build.gradle
```diff
@@ -12,7 +12,8 @@ sourceSets {
 dependencies {
     api project(":services-core")
     api project(":services-geojson")
-    api "com.mapbox.mapboxsdk:mapbox-sdk-directions-models:5.1.0-SNAPSHOT"
+    api project(":services-directions-models")
```
Guardiola31337: I believe we should keep adding the explicit dependency until it's released. Same for `services-directions-refresh-models` cc @lukaspakzos

gradle/checkstyle.gradle
Guardiola31337: Yeah, let's keep checkstyle as it was (if possible).

services/build.gradle
LukasPaczos: yea, I think that's correct

gradle/checkstyle.gradle
evabishchevich: It works just because we use '_checkstyleMain_' task on CI and it definitely won't work if use '_checkstyle_'

LukasPaczos:
Bintray repo is available here https://bintray.com/mapbox/mapbox/mapbox-sdk-directions-refresh-models.

services/build.gradle
evabishchevich: Do we still want to keep it as explicit dependency? 
@LukasPaczos  @Guardiola31337

PR #1110, opened 2020-01-23
Add toBuilder() method

korshaknn:
Only `RouteOptions`  and `WalkingOptions` don't have `toBuilder()` method.

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
@@ -1,6 +1,7 @@
 package com.mapbox.api.directions.v5;
 
 import androidx.annotation.FloatRange;
+import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
 
 import com.google.auto.value.AutoValue;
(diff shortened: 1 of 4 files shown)


services-directions-models/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
```diff
@@ -80,6 +80,16 @@ public final String toJson() {
...
 
+  /**
+   * Convert the current {@link WalkingOptions} to its builder holding the currently assigned
+   * values. This allows you to modify a single property and then rebuild the object resulting in
+   * an updated and modified {@link WalkingOptions}.
+   *
+   * @return a {@link WalkingOptions.Builder} with the same values set to match the ones defined
+   *   in this {@link WalkingOptions}
+   */
+  public abstract Builder toBuilder();
```
LukasPaczos: Can we annotate with `@NonNull`?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -349,6 +349,16 @@ public static RouteOptions fromJson(String json) {
...
 
+  /**
+   * Convert the current {@link RouteOptions} to its builder holding the currently assigned
+   * values. This allows you to modify a single property and then rebuild the object resulting in
+   * an updated and modified {@link RouteOptions}.
+   *
+   * @return a {@link RouteOptions.Builder} with the same values set to match the ones defined
+   *   in this {@link RouteOptions}
+   */
+  public abstract Builder toBuilder();
```
LukasPaczos: Same as above.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/WalkingOptions.java
korshaknn: done

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: done

PR #1115, opened 2020-01-27
add Map Matching params

korshaknn:


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -612,6 +621,39 @@ public abstract Builder overview(
...
     public abstract Builder walkingOptions(@NonNull WalkingOptions walkingOptions);
 
+    /**
+     * The origin_trace query parameter is a list of coordinates corresponding to locations leading
+     * up to the origin in the Directions API request.
+     *
+     * @param originTrace list of coordinates corresponding to locations leading up to the origin
+     *                    in the Directions API request
+     * @return this builder for chaining options together
+     * @since 4.3.0
```
LukasPaczos: Let's remove the `@since` annotations

korshaknn: done

RingerJK:
@korshaknn do we still need it?

korshaknn:
@RingerJK looks like it was a draft PR just to test some logic and it can be closed.
@LukasPaczos , @Guardiola31337  am I right?

LukasPaczos:
Yes, this was for internal delivery. Let's close for now.

PR #1118, opened 2020-01-30
RouteOptions and MapboxDirections refactoring

korshaknn:
Curent PR also has changes from https://github.com/mapbox/mapbox-java/pull/1115

### DIFF
--- a/services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
+++ b/services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
@@ -93,7 +93,9 @@ public final class TextUtils {
    * @param radiuses a double array which represents the radius values
    * @return a String ready for being passed into the Retrofit call
    * @since 3.0.0
+   * @deprecated use FormatUtils.formatRadiuses(List)
    */
+  @Deprecated
   public static String formatRadiuses(double[] radiuses) {
     if (radiuses == null || radiuses.length == 0) {
       return null;
@@ -118,7 +120,9 @@ public final class TextUtils {
    * @param bearings a List of doubles representing bearing values
    * @return a string with the bearing values
    * @since 3.0.0
+   * @deprecated use FormatUtils.formatBearing(List)
    */
+  @Deprecated
   public static String formatBearing(List<Double[]> bearings) {
     if (bearings.isEmpty()) {
       return null;
@@ -143,7 +147,9 @@ public final class TextUtils {
    * @param distributions the list of integer arrays representing the distribution
    * @return a string with the distribution values
    * @since 3.0.0
+   * @deprecated use FormatUtils.formatDistributions(List)
    */
+  @Deprecated
   public static String formatDistributions(List<Integer[]> distributions) {
     if (distributions.isEmpty()) {
       return null;
@@ -170,13 +176,15 @@ public final class TextUtils {
    * @param approaches a string representing approaches to each coordinate.
    * @return a formatted string.
    * @since 3.2.0
+   * @deprecated use FormatUtils.formatApproaches(List)
    */
+  @Deprecated
   public static String formatApproaches(String[] approaches) {
     for (int i = 0; i < approaches.length; i++) {
       if (approaches[i] == null) {
         approaches[i] = "";
       } else if (!approaches[i].equals("unrestricted")
-        && !approaches[i].equals("curb") && !approaches[i].isEmpty()) {
+          && !approaches[i].equals("curb") && !approaches[i].isEmpty()) {
         return null;
       }
     }
@@ -190,7 +198,9 @@ public final class TextUtils {
    * @param waypointNames a string representing approaches to each coordinate.
    * @return a formatted string.
    * @since 3.3.0
+   * @deprecated use FormatUtils.formatWaypointNames(List)
    */
+  @Deprecated
   public static String formatWaypointNames(String[] waypointNames) {
     for (int i = 0; i < waypointNames.length; i++) {
       if (waypointNames[i] == null) {
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
@@ -2,7 +2,6 @@ package com.mapbox.api.directions.v5.models;
 
 import androidx.annotation.NonNull;
 import androidx.annotation.Nullable;
-
 import com.google.auto.value.AutoValue;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
@@ -12,9 +11,10 @@ import com.mapbox.api.directions.v5.DirectionsAdapterFactory;
 import com.mapbox.api.directions.v5.DirectionsCriteria;
 import com.mapbox.api.directions.v5.WalkingOptions;
 import com.mapbox.api.directions.v5.WalkingOptionsAdapterFactory;
+import com.mapbox.api.directions.v5.utils.FormatUtils;
+import com.mapbox.api.directions.v5.utils.ParseUtils;
 import com.mapbox.geojson.Point;
 import com.mapbox.geojson.PointAsCoordinatesTypeAdapter;
-
 import java.util.List;
 
 /**
@@ -65,20 +65,26 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract String user();
 
   /**
+   * The routing profile to use. Possible values are
+   * {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC}, {@link DirectionsCriteria#PROFILE_DRIVING},
+   * {@link DirectionsCriteria#PROFILE_WALKING}, or {@link DirectionsCriteria#PROFILE_CYCLING}.
    * The same profile which was used during the request that resulted in this root directions
    * response. <tt>MapboxDirections.Builder</tt> ensures that a profile is always set even if the
    * <tt>MapboxDirections</tt> requesting object doesn't specifically set a profile.
    *
-   * @return string value representing the profile
+   * @return string value representing the profile defined in
+   *         {@link DirectionsCriteria.ProfileCriteria}
    * @since 3.0.0
    */
   @NonNull
   public abstract String profile();
 
   /**
-   * The coordinates used for the routes origin, destination, and optionally, waypoints. Note that
-   * these coordinates are different than the direction responses {@link DirectionsWaypoint}s in
-   * that these are the non-snapped coordinates.
+   * A list of Points to visit in order.
+   * There can be between two and 25 coordinates for most requests, or up to three coordinates for
+   * {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC} requests.
+   * Note that these coordinates are different than the direction responses
+   * {@link DirectionsWaypoint}s that these are the non-snapped coordinates.
    *
    * @return a list of {@link Point}s which represent the route origin, destination,
    *   and optionally, waypoints
@@ -88,8 +94,11 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract List<Point> coordinates();
 
   /**
-   * The same alternative setting which were used during the request that resulted in this root
-   * directions response.
+   * Whether to try to return alternative routes (true) or not (false, default). An alternative
+   * route is a route that is significantly different than the fastest route, but also still
+   * reasonably fast. Such a route does not exist in all circumstances. Up to two alternatives may
+   * be returned. This is available for {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC},
+   * {@link DirectionsCriteria#PROFILE_DRIVING}, {@link DirectionsCriteria#PROFILE_CYCLING}.
    *
    * @return boolean object representing the setting for alternatives
    * @since 3.0.0
@@ -98,8 +107,8 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract Boolean alternatives();
 
   /**
-   * The same language which was used during the request that resulted in this root directions
-   * response.
+   * The language of returned turn-by-turn text instructions. The default is en (English).
+   * Must be used in conjunction with {@link RouteOptions.Builder#steps(Boolean)}.
    *
    * @return the language as a string used during the request,
    *   if english, this will most likely be empty
@@ -109,28 +118,73 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract String language();
 
   /**
-   * The same radiuses were used during the request that resulted in this root directions response.
+   * The maximum distance a coordinate can be moved to snap to the road network in meters. There
+   * must be as many radiuses as there are coordinates in the request, each separated by ;.
+   * Values can be any number greater than 0, the string unlimited or empty string.
    *
-   * @return a string representing the radiuses
+   * @return a string representing the radiuses separated by ;.
    * @since 3.0.0
+   * @deprecated use {@link #radiusesList()} ()}
    */
   @Nullable
-  public abstract String radiuses();
+  @Deprecated
+  public String radiuses() {
+    return FormatUtils.formatRadiuses(radiusesList());
+  }
 
   /**
-   * The same bearings which were used during the request that resulted in this root directions
-   * response. Note that even though these are saved. it's a good idea to recalculate any bearings
-   * being used and use the newer values for the directions request.
+   * The maximum distance a coordinate can be moved to snap to the road network in meters. There
+   * must be as many radiuses as there are coordinates in the request.
+   * Values can be any number greater than 0, the string unlimited, or null.
    *
-   * @return a string representing the bearings used in the original request
-   * @since 3.0.0
+   * @return a list of radiuses
    */
   @Nullable
-  public abstract String bearings();
+  public abstract List<Double> radiusesList();
 
   /**
-   * The same continueStraight setting which was used during the request that resulted in this root
-   * directions response.
+   * Influences the direction in which a route starts from a waypoint. Used to filter the road
+   * segment the waypoint will be placed on by direction. This is useful for making sure the new
+   * routes of rerouted vehicles continue traveling in their current direction. A request that does
+   * this would provide bearing and radius values for the first waypoint and leave the remaining
+   * values empty. Returns two comma-separated values per waypoint: an angle clockwise from true
+   * north between 0 and 360, and the range of degrees by which the angle can deviate (recommended
+   * value is 45° or 90°), formatted as {angle, degrees}. If provided, the list of bearings must be
+   * the same length as the list of coordinates.
+   *
+   * @return a string representing the bearings with the ; separator. Angle and degrees for every
+   *   bearing value are comma-separated.
+   * @since 3.0.0
+   * @deprecated use {@link #bearingsList()}
+   */
+  @Nullable
+  @Deprecated
+  public String bearings() {
+    return FormatUtils.formatBearings(bearingsList());
+  }
+
+  /**
+   * Influences the direction in which a route starts from a waypoint. Used to filter the road
+   * segment the waypoint will be placed on by direction. This is useful for making sure the new
+   * routes of rerouted vehicles continue traveling in their current direction. A request that does
+   * this would provide bearing and radius values for the first waypoint and leave the remaining
+   * values empty. Returns a list of values, each value is a list of an angle clockwise from true
+   * north between 0 and 360, and the range of degrees by which the angle can deviate (recommended
+   * value is 45° or 90°).
+   * If provided, the list of bearings must be the same length as the list of coordinates.
+   *
+   * @return a List of list of doubles representing the bearings used in the original request.
+   *         The first value in the list is the angle, the second one is the degrees.
+   */
+  @SerializedName("bearings")
+  @Nullable
+  public abstract List<List<Double>> bearingsList();
+
+  /**
+   * The allowed direction of travel when departing intermediate waypoints. If true, the route
+   * will continue in the same direction of travel. If false, the route may continue in the opposite
+   * direction of travel. Defaults to true for {@link DirectionsCriteria#PROFILE_DRIVING} and false
+   * for {@link DirectionsCriteria#PROFILE_WALKING} and {@link DirectionsCriteria#PROFILE_CYCLING}.
    *
    * @return a boolean value representing whether or not continueStraight was enabled or
    *   not during the initial request
@@ -141,8 +195,11 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract Boolean continueStraight();
 
   /**
-   * This is set to true if you want to enable instructions while exiting roundabouts
-   * and rotaries.
+   * Whether to emit instructions at roundabout exits (true) or not (false, default). Without
+   * this parameter, roundabout maneuvers are given as a single instruction that includes both
+   * entering and exiting the roundabout. With roundabout_exits=true, this maneuver becomes two
+   * instructions, one for entering the roundabout and one for exiting it. Must be used in
+   * conjunction with {@link RouteOptions#steps()}=true.
    *
    * @return a boolean value representing whether or not roundaboutExits was enabled or disabled
    *   during the initial route request
@@ -153,26 +210,36 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract Boolean roundaboutExits();
 
   /**
-   * Geometry type used to make the initial directions request.
+   * The format of the returned geometry. Allowed values are:
+   * {@link DirectionsCriteria#GEOMETRY_POLYLINE} (default, a polyline with a precision of five
+   * decimal places), {@link DirectionsCriteria#GEOMETRY_POLYLINE6} (a polyline with a precision
+   * of six decimal places).
    *
-   * @return String geometry type used to make the initial directions request.
+   * @return String geometry type from {@link DirectionsCriteria.GeometriesCriteria}.
    * @since 3.1.0
    */
+  @Nullable
   public abstract String geometries();
 
   /**
-   * Type of returned overview geometry that was used to make the initial directions request.
+   * Displays the requested type of overview geometry. Can be
+   * {@link DirectionsCriteria#OVERVIEW_FULL} (the most detailed geometry
+   * available), {@link DirectionsCriteria#OVERVIEW_SIMPLIFIED} (default, a simplified version of
+   * the full geometry), or {@link DirectionsCriteria#OVERVIEW_FALSE} (no overview geometry).
    *
-   * @return null or one of the options found in
-   *   {@link DirectionsCriteria.OverviewCriteria}
+   * @return null or one of the options found in {@link DirectionsCriteria.OverviewCriteria}
    * @since 3.1.0
    */
   @Nullable
   public abstract String overview();
 
   /**
-   * Boolean value used to determine whether to return steps and turn-by-turn instructions in the
-   * initial directions request.
+   * Whether to return steps and turn-by-turn instructions (true) or not (false, default).
+   * If steps is set to true, the following guidance-related parameters will be available:
+   * {@link RouteOptions#bannerInstructions()}, {@link RouteOptions#language()},
+   * {@link RouteOptions#roundaboutExits()}, {@link RouteOptions#voiceInstructions()},
+   * {@link RouteOptions#voiceUnits()}, {@link RouteOptions#waypointNamesList()},
+   * {@link RouteOptions#waypointTargetsList()}, waypoints from {@link RouteOptions#coordinates()}
    *
    * @return true if you'd like step information, false otherwise
    * @since 3.1.0
@@ -181,26 +248,69 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract Boolean steps();
 
   /**
-   * The same annotations in String format which were used during the request that resulted in this
-   * root directions response.
+   * A comma-separated list of annotations. Defines whether to return additional metadata along the
+   * route. Possible values are:
+   * {@link DirectionsCriteria#ANNOTATION_DURATION}
+   * {@link DirectionsCriteria#ANNOTATION_DISTANCE}
+   * {@link DirectionsCriteria#ANNOTATION_SPEED}
+   * {@link DirectionsCriteria#ANNOTATION_CONGESTION}
+   * {@link DirectionsCriteria#ANNOTATION_MAXSPEED}
+   * See the {@link RouteLeg} object for more details on what is included with annotations.
+   * Must be used in conjunction with overview=full.
    *
    * @return a string containing any of the annotations that were used during the request
    * @since 3.0.0
+   *
+   * @deprecated use {@link #annotationsList()}
    */
+  @Deprecated
   @Nullable
-  public abstract String annotations();
+  public String annotations() {
+    return FormatUtils.join(";", annotationsList(), true);
+  }
 
   /**
-   * The same exclusions the user originally made when the request was made.
+   * A list of annotations. Defines whether to return additional metadata along the
+   * route. Possible values are:
+   * {@link DirectionsCriteria#ANNOTATION_DURATION}
+   * {@link DirectionsCriteria#ANNOTATION_DISTANCE}
+   * {@link DirectionsCriteria#ANNOTATION_SPEED}
+   * {@link DirectionsCriteria#ANNOTATION_CONGESTION}
+   * {@link DirectionsCriteria#ANNOTATION_MAXSPEED}
+   * See the {@link RouteLeg} object for more details on what is included with annotations.
+   * Must be used in conjunction with overview=full.
    *
-   * @return a string matching one of the {@link DirectionsCriteria} exclusions
+   * @return a list of annotations that were used during the request
+   */
+  @SerializedName("annotations")
+  @Nullable
+  public abstract List<String> annotationsList();
+
+  /**
+   * Exclude certain road types from routing. The default is to not exclude anything from the
+   * profile selected. The following exclude flags are available for each profile:
+   *
+   * {@link DirectionsCriteria#PROFILE_DRIVING}: One of {@link DirectionsCriteria#EXCLUDE_TOLL},
+   * {@link DirectionsCriteria#EXCLUDE_MOTORWAY}, or {@link DirectionsCriteria#EXCLUDE_FERRY}.
+   *
+   * {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC}: One of
+   * {@link DirectionsCriteria#EXCLUDE_TOLL}, {@link DirectionsCriteria#EXCLUDE_MOTORWAY}, or
+   * {@link DirectionsCriteria#EXCLUDE_FERRY}.
+   *
+   * {@link DirectionsCriteria#PROFILE_WALKING}: No excludes supported
+   *
+   * {@link DirectionsCriteria#PROFILE_CYCLING}: {@link DirectionsCriteria#EXCLUDE_FERRY}
+   *
+   * @return a string matching one of the {@link DirectionsCriteria.ExcludeCriteria} exclusions
    * @since 3.0.0
    */
   @Nullable
   public abstract String exclude();
 
   /**
-   * Whether or not the request had voice instructions set to true or not.
+   * Whether to return SSML marked-up text for voice guidance along the route (true) or not
+   * (false, default).
+   * Must be used in conjunction with {@link RouteOptions#steps()}=true.
    *
    * @return true if the original request included voice instructions
    * @since 3.0.0
@@ -210,7 +320,8 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract Boolean voiceInstructions();
 
   /**
-   * Whether or not the request had banner instructions set to true or not.
+   * Whether to return banner objects associated with the route steps (true) or not
+   * (false, default). Must be used in conjunction with {@link RouteOptions#steps()}=true
    *
    * @return true if the original request included banner instructions
    * @since 3.0.0
@@ -220,7 +331,10 @@ public abstract class RouteOptions extends DirectionsJsonObject {
   public abstract Boolean bannerInstructions();
 
   /**
-   * Whether or not the units used inside the voice instruction's string are in imperial or metric.
+   * A type of units to return in the text for voice instructions.
+   * Can be {@link DirectionsCriteria#IMPERIAL} (default) or {@link DirectionsCriteria#METRIC}.
+   * Must be used in conjunction with {@link RouteOptions#steps()}=true and
+   * {@link RouteOptions#voiceInstructions()} ()}=true.
    *
    * @return a string matching either imperial or metric
    * @since 3.0.0
@@ -248,25 +362,46 @@ public abstract class RouteOptions extends DirectionsJsonObject {
    * @since 3.0.0
    */
   @SerializedName("uuid")
-  @NonNull
+  @Nullable
   public abstract String requestUuid();
 
   /**
    * Indicates from which side of the road to approach a waypoint.
-   * Accepts  unrestricted (default) or  curb . If set to  unrestricted ,
-   * the route can approach waypoints from either side of the road.
-   * If set to curb, the route will be returned so that on arrival,
-   * the waypoint will be found on the side that corresponds with the driving_side of the region
-   * in which the returned route is located.
+   * Accepts  {@link DirectionsCriteria#APPROACH_UNRESTRICTED} (default) or
+   * {@link DirectionsCriteria#APPROACH_CURB} .
+   * If set to {@link DirectionsCriteria#APPROACH_UNRESTRICTED}, the route can approach waypoints
+   * from either side of the road.
+   * If set to {@link DirectionsCriteria#APPROACH_CURB}, the route will be returned so that on
+   * arrival, the waypoint will be found on the side that corresponds with the driving_side of the
+   * region in which the returned route is located.
    * If provided, the list of approaches must be the same length as the list of waypoints.
-   * However, you can skip a coordinate and show its position in the list with the ; separator.
    *
    * @return a string representing approaches for each waypoint
    * @since 3.2.0
+   * @deprecated use {@link #approachesList()}
    */
-
+  @Deprecated
   @Nullable
-  public abstract String approaches();
+  public String approaches() {
+    return FormatUtils.formatApproaches(approachesList());
+  }
+
+  /**
+   * Indicates from which side of the road to approach a waypoint.
+   * Accepts  {@link DirectionsCriteria#APPROACH_UNRESTRICTED} (default) or
+   * {@link DirectionsCriteria#APPROACH_CURB} .
+   * If set to {@link DirectionsCriteria#APPROACH_UNRESTRICTED}, the route can approach waypoints
+   * from either side of the road.
+   * If set to {@link DirectionsCriteria#APPROACH_CURB}, the route will be returned so that on
+   * arrival, the waypoint will be found on the side that corresponds with the driving_side of the
+   * region in which the returned route is located.
+   * If provided, the list of approaches must be the same length as the list of waypoints.
+   *
+   * @return a list of strings representing approaches for each waypoint
+   */
+  @SerializedName("approaches")
+  @Nullable
+  public abstract List<String> approachesList();
 
   /**
    * Indicates which input coordinates should be treated as waypoints.
@@ -279,22 +414,61 @@ public abstract class RouteOptions extends DirectionsJsonObject {
    *
    * @return a string representing indices to be used as waypoints
    * @since 4.4.0
+   * @deprecated use {@link #waypointIndicesList()}
+   */
+  @Nullable
+  @Deprecated
+  public String waypointIndices() {
+    return FormatUtils.join(";", waypointIndicesList());
+  }
+
+  /**
+   * Indicates which input coordinates should be treated as waypoints.
+   * <p>
+   * Most useful in combination with  steps=true and requests based on traces
+   * with high sample rates. Can be an index corresponding to any of the input coordinates,
+   * but must contain the first ( 0 ) and last coordinates' index.
+   * {@link #steps()}
+   * </p>
+   *
+   * @return a List of Integers representing indices to be used as waypoints
    */
   @SerializedName("waypoints")
   @Nullable
-  public abstract String waypointIndices();
+  public abstract List<Integer> waypointIndicesList();
 
   /**
-   * Custom names for waypoints used for the arrival instruction in banners and voice instructions,
-   * each separated by  ; . Values can be any string and total number of all characters cannot
-   * exceed 500. If provided, the list of waypoint_names must be the same length as the list of
-   * coordinates, but you can skip a coordinate and show its position with the ; separator.
+   * A semicolon-separated list of custom names for entries in the list of
+   * {@link RouteOptions#coordinates()}, used for the arrival instruction in banners and voice
+   * instructions. Values can be any string, and the total number of all characters cannot exceed
+   * 500. If provided, the list of waypoint_names must be the same length as the list of
+   * coordinates. The first value in the list corresponds to the route origin, not the first
+   * destination.
+   * Must be used in conjunction with {@link RouteOptions#steps()} = true.
    * @return  a string representing names for each waypoint
    * @since 3.3.0
+   * @deprecated use {@link #waypointNamesList()}
+   */
+  @Nullable
+  @Deprecated
+  public String waypointNames() {
+    return FormatUtils.formatWaypointNames(waypointNamesList());
+  }
+
+  /**
+   * A semicolon-separated list of custom names for entries in the list of
+   * {@link RouteOptions#coordinates()}, used for the arrival instruction in banners and voice
+   * instructions. Values can be any string, and the total number of all characters cannot exceed
+   * 500. If provided, the list of waypoint_names must be the same length as the list of
+   * coordinates. The first value in the list corresponds to the route origin, not the first
+   * destination.
+   * Must be used in conjunction with {@link RouteOptions#steps()} = true.
+   *
+   * @return  a list of strings representing names for each waypoint
    */
   @SerializedName("waypoint_names")
   @Nullable
-  public abstract String waypointNames();
+  public abstract List<String> waypointNamesList();
 
 
   /**
@@ -303,15 +477,31 @@ public abstract class RouteOptions extends DirectionsJsonObject {
    * If this parameter is provided, the Directions API will compute the side of the street,
    * left or right, for each target based on the waypoint_targets and the driving direction.
    * The maneuver.modifier, banner and voice instructions will be updated with the computed
-   * side of street. The number of waypoint targets must be the same as the number of coordinates,
-   * but you can skip a coordinate pair and show its position in the list with the ; separator.
-   * Must be used with steps=true.
-   * @return  a string representing coordinate pairs for drop-off locations
+   * side of street. The number of waypoint targets must be the same as the number of coordinates.
+   * Must be used with {@link RouteOptions#steps()} = true.
+   * @return  a list of Points representing coordinate pairs for drop-off locations
    * @since 4.3.0
+   * @deprecated use {@link #waypointTargetsList()}
+   */
+  @Nullable
+  @Deprecated
+  public String waypointTargets() {
+    return FormatUtils.formatPointsList(waypointTargetsList());
+  }
+
+  /**
+   * A list of points used to specify drop-off
+   * locations that are distinct from the locations specified in coordinates.
+   * If this parameter is provided, the Directions API will compute the side of the street,
+   * left or right, for each target based on the waypoint_targets and the driving direction.
+   * The maneuver.modifier, banner and voice instructions will be updated with the computed
+   * side of street. The number of waypoint targets must be the same as the number of coordinates.
+   * Must be used with {@link RouteOptions#steps()} = true.
+   * @return  a list of Points representing coordinate pairs for drop-off locations
    */
   @SerializedName("waypoint_targets")
   @Nullable
-  public abstract String waypointTargets();
+  public abstract List<Point> waypointTargetsList();
 
   /**
    * To be used to specify settings for use with the walking profile.
@@ -341,6 +531,7 @@ public abstract class RouteOptions extends DirectionsJsonObject {
    *   method
    * @since 3.4.0
    */
+  @NonNull
   public static RouteOptions fromJson(String json) {
     GsonBuilder gson = new GsonBuilder();
     gson.registerTypeAdapterFactory(DirectionsAdapterFactory.create());
@@ -388,20 +579,27 @@ public abstract class RouteOptions extends DirectionsJsonObject {
     public abstract Builder user(@NonNull String user);
 
     /**
-     * The directions profile that was used during the request time and resulted in this responses
-     * result.
+     * The routing profile to use. Possible values are
+     * {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC},
+     * {@link DirectionsCriteria#PROFILE_DRIVING}, {@link DirectionsCriteria#PROFILE_WALKING}, or
+     * {@link DirectionsCriteria#PROFILE_CYCLING}.
+     * The same profile which was used during the request that resulted in this root directions
+     * response. <tt>MapboxDirections.Builder</tt> ensures that a profile is always set even if the
+     * <tt>MapboxDirections</tt> requesting object doesn't specifically set a profile.
      *
      * @param profile One of the direction profiles defined in
-     *                {@link DirectionsCriteria#DirectionsCriteria()}
+     *                {@link DirectionsCriteria.ProfileCriteria}
      * @return this builder for chaining options together
      * @since 3.0.0
      */
     public abstract Builder profile(@NonNull @DirectionsCriteria.ProfileCriteria String profile);
 
     /**
-     * The coordinates used for the routes origin, destination, and optionally, waypoints. Note that
-     * these coordinates are different than the direction responses {@link DirectionsWaypoint}s in
-     * that these are the non-snapped coordinates.
+     * A list of Points to visit in order.
+     * There can be between two and 25 coordinates for most requests, or up to three coordinates for
+     * {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC} requests.
+     * Note that these coordinates are different than the direction responses
+     * {@link DirectionsWaypoint}s that these are the non-snapped coordinates.
      *
      * @param coordinates a list of {@link Point}s which represent the route origin, destination,
      *                    and optionally, waypoints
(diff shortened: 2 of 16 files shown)


services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
```diff
@@ -111,27 +126,44 @@ public static String formatRadiuses(double[] radiuses) {
...
    * Formats the bearing variables from the raw values to a string which can than be used for the
    * request URL.
    *
-   * @param bearings a List of doubles representing bearing values
+   * @param bearings a List of list of doubles representing bearing values
    * @return a string with the bearing values
    * @since 3.0.0
    */
-  public static String formatBearing(List<Double[]> bearings) {
+  public static String formatBearing(List<List<Double>> bearings) {
```
RingerJK: might `List<Double>` changed to `Pair` or custom model?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -111,11 +111,11 @@ public static Builder builder() {
   /**
    * The same radiuses were used during the request that resulted in this root directions response.
    *
-   * @return a string representing the radiuses
+   * @return a list of radiuses
    * @since 3.0.0
    */
   @Nullable
-  public abstract String radiuses();
+  public abstract List<Double> radiuses();
```
RingerJK: 👍

services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
korshaknn: yep, agree, I just wanted to make it in a separate PR to make review and testing easier

RingerJK: leave `TODO` to not forget 😃

korshaknn:
Also, maybe I should make all list params in `MapboxDirections` Nullable? to make it possible to remove remove unnecessary ones?

services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
```diff
@@ -111,27 +126,44 @@ public static String formatRadiuses(double[] radiuses) {
...
 
+  public static String formatRadiuses(List<Double> radiuses) {
+    if (radiuses == null || radiuses.size() == 0) {
+      return null;
+    }
+
+    String[] radiusesFormatted = new String[radiuses.size()];
+    for (int i = 0; i < radiuses.size(); i++) {
+      if (radiuses.get(i) == Double.POSITIVE_INFINITY) {
+        radiusesFormatted[i] = "unlimited";
```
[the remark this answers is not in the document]
Guardiola31337: Yeah 👀 https://docs.mapbox.com/api/navigation/#retrieve-directions

> The maximum distance a coordinate can be moved to snap to the road network in meters. There must be as many radiuses as there are coordinates in the request, each separated by ;. Values can be any number greater than 0 or the string unlimited. A NoSegment error is returned if no routable road is found within the radius.

services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
```diff
@@ -111,27 +126,44 @@ public static String formatRadiuses(double[] radiuses) {
...
     String[] bearingFormatted = new String[bearings.size()];
     for (int i = 0; i < bearings.size(); i++) {
-      if (bearings.get(i).length == 0) {
+      if (bearings.get(i).size() == 0) {
         bearingFormatted[i] = "";
       } else {
         bearingFormatted[i] = String.format(Locale.US, "%s,%s",
-          TextUtils.formatCoordinate(bearings.get(i)[0]),
-          TextUtils.formatCoordinate(bearings.get(i)[1]));
+                TextUtils.formatCoordinate(bearings.get(i).get(0)),
```
Guardiola31337: NIT formatting (actually `checkstyle` is ❌ in CI 😅 https://circleci.com/gh/mapbox/mapbox-java/3900?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link)

For Java code we use [this](https://github.com/mapbox/mapbox-navigation-android/tree/master/config/checkstyle) checkstyle and code style. You can find instructions on how to set both in [Setting up Mapbox checkstyle](https://github.com/mapbox/mapbox-gl-native-android/wiki/Setting-up-Mapbox-checkstyle).

services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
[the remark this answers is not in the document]
korshaknn: I can't have 2 methods
`formatBearing(List<Double[]> bearings)`
and
`formatBearing(List<List<Double>> bearings)`
because I get the next compile error:

Screen Shot 2020-01-31 at 12 32 14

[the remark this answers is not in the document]
korshaknn: so we can add new static parse methods to another file. To `MapboxDirections` for example
wdyt?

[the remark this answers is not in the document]
LukasPaczos: How about naming the method `formatBearings`? That would make sense from the data standpoint as well.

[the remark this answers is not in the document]
korshaknn: makes sense!

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -431,21 +455,21 @@ public static RouteOptions fromJson(String json) {
     /**
      * The radiuses in string format that were used during the original request.
      *
```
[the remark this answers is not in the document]
LukasPaczos: Could you also address ^^^ @korshaknn?

services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
```diff
@@ -86,117 +85,4 @@ public static String formatCoordinate(double coordinate, int precision) {
...
   }
-
-  /**
-   * Used in various APIs to format the user provided radiuses to a String matching the APIs format.
-   *
-   * @param radiuses a double array which represents the radius values
-   * @return a String ready for being passed into the Retrofit call
-   * @since 3.0.0
-   */
-  public static String formatRadiuses(double[] radiuses) {
```
LukasPaczos: We cannot remove these methods as that would break semver.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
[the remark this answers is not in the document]
LukasPaczos: Could you also address ^^^ @korshaknn?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,216 @@
...
+ */
+public class FormatUtils {
+  /**
+   * Returns a string containing the tokens joined by delimiters.
+   *
+   * @param delimiter the delimeter on which to split.
+   * @param tokens An array objects to be joined. Strings will be formed from the objects by calling
+   *               object.toString().
+   * @return {@link String}
+   * @since 1.0.0
```
RingerJK: wrong `since x.x.x`

korshaknn: removed

services-core/src/main/java/com/mapbox/core/utils/TextUtils.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -109,28 +118,76 @@ public static Builder builder() {
...
+  @Nullable
+  @Deprecated
+  public String radiuses() {
+    return FormatUtils.formatRadiuses(radiusesList());
+  }
+
+  /**
+   * The maximum distance a coordinate can be moved to snap to the road network in meters. There
+   * must be as many radiuses as there are coordinates in the request.
+   * Values can be any number greater than 0 or the string unlimited. A NoSegment error is returned
```
LukasPaczos: The "unlimited" string is not going to be returned here. What happens when we supply "unlimited" or "NoSegment" to `Double.valueOf`?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -109,28 +118,76 @@ public static Builder builder() {
...
+  public String bearings() {
+    return FormatUtils.formatBearings(bearingsList());
+  }
+
+  /**
+   * Influences the direction in which a route starts from a waypoint. Used to filter the road
+   * segment the waypoint will be placed on by direction. This is useful for making sure the new
+   * routes of rerouted vehicles continue traveling in their current direction. A request that does
+   * this would provide bearing and radius values for the first waypoint and leave the remaining
+   * values empty. Takes two comma-separated values per waypoint: an angle clockwise from true north
```
LukasPaczos: This does not take, but return a value that's a list of double.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -153,26 +213,37 @@ public static Builder builder() {
   public abstract Boolean roundaboutExits();
 
   /**
-   * Geometry type used to make the initial directions request.
+   * The format of the returned geometry. Allowed values are:
+   * {@link DirectionsCriteria#GEOMETRY_POLYLINE} (default, a polyline with a precision of five
+   * decimal places), {@link DirectionsCriteria#GEOMETRY_POLYLINE6} (a polyline with a precision
+   * of six decimal places).
+   * A null value will reset this field to the APIs default value vs this SDKs default value of
+   * {@link DirectionsCriteria#GEOMETRY_POLYLINE6}.
```
LukasPaczos: Sentences like this should only be in the builder methods, not getters. Same with other instances that mention `take`/`specify`/`set` instead of `return` etc. This is not a big deal though.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -429,26 +616,78 @@ public static RouteOptions fromJson(String json) {
...
+     */
+    @Deprecated
+    public Builder radiuses(String radiuses) {
+      return radiusesList(ParseUtils.parseToDoubles(radiuses));
+    }
+
+    /**
+     * The maximum distance a coordinate can be moved to snap to the road network in meters. There
+     * must be as many radiuses as there are coordinates in the request.
+     * Values can be any number greater than 0 or the string unlimited.
```
LukasPaczos: This does not accept the `unlimited`, but I think it does accept `null`? How can we phrase this better?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -429,26 +616,78 @@ public static RouteOptions fromJson(String json) {
...
+    public Builder bearings(String bearings) {
+      return bearingsList(ParseUtils.parseToListOfListOfDoubles(bearings));
+    }
+
+    /**
+     * Influences the direction in which a route starts from a waypoint. Used to filter the road
+     * segment the waypoint will be placed on by direction. This is useful for making sure the new
+     * routes of rerouted vehicles continue traveling in their current direction. A request that
+     * does this would provide bearing and radius values for the first waypoint and leave the
+     * remaining values empty. Takes two comma-separated values per waypoint: an angle clockwise
```
LukasPaczos: Similarly as above.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+      return null;
+    }
+
+    String[] bearingFormatted = new String[bearings.size()];
+    for (int i = 0; i < bearings.size(); i++) {
+      if (bearings.get(i).size() == 0) {
+        bearingFormatted[i] = "";
+      } else {
+        bearingFormatted[i] = String.format(Locale.US, "%s,%s",
+            formatCoordinate(bearings.get(i).get(0)),
```
LukasPaczos: The nested coordinate lists are nullable if we specify `null` as the bearing for a specific waypoint or origin.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,134 @@
...
+   * @param original an original String.
+   * @return List of Doubles
+   */
+  public static List<Double> parseToDoubles(String original) {
+    List<Double> doubles = new ArrayList<>();
+
+    if (original != null) {
+      String[] array = original.split(SEMICOLON);
+      for (String radius : array) {
+        if (radius != null && !radius.isEmpty()) {
```
LukasPaczos: We should add empty strings as `nulls`, otherwise, we are breaking the order of values in reference to the number of coordinates. Same in other occurrences. We should add tests around these scenarios.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -559,49 +820,174 @@ public abstract Builder overview(
...
+     * region in which the returned route is located.
+     * If provided, the list of approaches must be the same length as the list of waypoints.
+     * However, you can skip a coordinate and show its position in the list with null value.
+     * The same approaches the user originally made when the request was made.
+     *
+     * @param approaches a list of Strings
+     * @return this builder for chaining options together
      */
-    public abstract Builder approaches(String approaches);
+    public abstract Builder approachesList(List<String> approaches);
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+public class FormatUtils {
+  /**
+   * Returns a string containing the tokens joined by delimiters.
+   *
+   * @param delimiter the delimeter on which to split.
+   * @param tokens An array objects to be joined. Strings will be formed from the objects by calling
+   *               object.toString().
+   * @return {@link String}
+   */
+  public static String join(CharSequence delimiter, Object[] tokens) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+
+  /**
+   * Returns a string containing the tokens joined by delimiters.
+   *
+   * @param delimiter the delimiter on which to split.
+   * @param tokens A list of objects to be joined. Strings will be formed from the objects by
+   *               calling object.toString().
+   * @return {@link String}
+   */
+  public static String join(CharSequence delimiter, List<?> tokens) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Used in various APIs to format the user provided radiuses to a String matching the APIs
+   * format.
+   *
+   * @param radiuses a list of doubles represents the radius values
+   * @return a String ready for being passed into the Retrofit call
+   */
+  public static String formatRadiuses(List<Double> radiuses) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Formats the bearing variables from the raw values to a string which can than be used for the
+   * request URL.
+   *
+   * @param bearings a List of list of doubles representing bearing values
+   * @return a string with the bearing values
+   */
+  public static String formatBearings(List<List<Double>> bearings) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+    return join(";", bearingFormatted);
+  }
+
+  /**
+   * converts the list of integer arrays to a string ready for API consumption.
+   *
+   * @param distributions the list of integer arrays representing the distribution
+   * @return a string with the distribution values
+   */
+  public static String formatDistributions(List<Integer[]> distributions) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+  }
+
+  /**
+   * Converts String list with approaches values to a string ready for API consumption. An approach
+   * could be unrestricted, curb or null.
+   *
+   * @param approaches a list representing approaches to each coordinate.
+   * @return a formatted string.
+   */
+  public static String formatApproaches(List<String> approaches) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+    return join(";", approaches);
+  }
+
+  /**
+   * Converts String list with waypoint_names values to a string ready for API consumption.
+   *
+   * @param waypointNames a string representing approaches to each coordinate.
+   * @return a formatted string.
+   */
+  public static String formatWaypointNames(List<String> waypointNames) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+    return join(";", waypointNames);
+  }
+
+  /**
+   * Converts a list of Points to String.
+   *
+   * @param coordinates a list of coordinates.
+   * @return a formatted string.
+   */
+  public static String formatCoordinates(List<Point> coordinates) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,211 @@
...
+    return join(";", coordinatesFormatted);
+  }
+
+  /**
+   * Converts array of Points with waypoint_targets values to a string ready for API consumption.
+   *
+   * @param points a list representing approaches to each coordinate.
+   * @return a formatted string.
+   */
+  public static String formatPointsList(List<Point> points) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,134 @@
...
+public class ParseUtils {
+  private static final String SEMICOLON = ";";
+  private static final String COMMA = ",";
+
+  /**
+   * Parse a String to a list of Integers.
+   * @param original an original String.
+   * @return List of Integers
+   */
+  public static List<Integer> parseToIntegers(String original) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+    return integers;
+  }
+
+  /**
+   * Parse a String to a list of Strings.
+   * @param original an original String.
+   * @return  List of Strings
+   */
+  public static List<String> parseToStrings(String original) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+    return strings;
+  }
+
+  /**
+   * Parse a String to a list of Points.
+   * @param original an original String.
+   * @return List of Points
+   */
+  public static List<Point> parseToPoints(String original) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+    return points;
+  }
+
+  /**
+   * Parse a String to a list of Points.
+   * @param original an original String.
+   * @return List of Doubles
+   */
+  public static List<Double> parseToDoubles(String original) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+    return doubles;
+  }
+
+  /**
+   * Parse a String to a list of Longs.
+   * @param original an original String.
+   * @return List of Longs
+   */
+  public static List<Long> parseToLongs(String original) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
Guardiola31337: Yeah, as mentioned we should double check the formatting methods as there are cases in which `null` is valid and gets formatted to `""`.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
Guardiola31337: Yeah, as mentioned we should double check the formatting methods as there are cases in which `null` is valid and gets formatted to `""`.

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -416,17 +382,17 @@ public static Builder builder() {
    */
   @AutoValue.Builder
   public abstract static class Builder {
-
-    private List<Double[]> bearings = new ArrayList<>();
+    //TODO change List<Double> to a custom model or to a Pair
```
Guardiola31337: Do you mean `List<List<Double>>` for `bearings`? If so, let's clarify in the comment.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,134 @@
...
+
+    return longs;
+  }
+
+  /**
+   * Parse a String to a list of list of Doubles.
+   * @param original an original String.
+   * @return List of List of Doubles
+   */
+  public static List<List<Double>> parseToListOfListOfDoubles(String original) {
```
Guardiola31337: Same as above re: `@Nullable` / `@NonNull` annotations

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -517,13 +641,38 @@ public void testRadiusWithUnlimitedDistance() throws IOException {
...
-      .radiuses(100, Double.POSITIVE_INFINITY, 100)
+      .radiuses(100d, Double.POSITIVE_INFINITY, 100d)
       .build();
 
     assertEquals("100;unlimited;100",
       client.cloneCall().request().url().queryParameter("radiuses"));
   }
 
+  @Test
+  public void testRadiusesListWithUnlimitedDistance() throws IOException {
```
Guardiola31337: We should also remove `test` prefix from tests names, it's implicit with the usage of `@Test` annotation.

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -688,6 +837,27 @@ public void testApproaches() throws Exception {
     assertEquals("Ok", response.body().code());
   }
 
+  @Test
+  public void testApproachesList() throws Exception {
```
Guardiola31337: Same as above re: `test` / `@Test`

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -704,8 +874,34 @@ public void testRouteOptionsApproaches() throws Exception {
...
 
-    String approaches = routeOptions.approaches();
-    assertEquals("unrestricted;curb", approaches);
+    List<String> approaches = routeOptions.approachesList();
+    assertEquals("unrestricted", approaches.get(0));
+    assertEquals("curb", approaches.get(1));
+  }
+
+  @Test
+  public void testRouteOptionsApproachesList() throws Exception {
```
Guardiola31337: Same as above re: `test` / `@Test`

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -791,6 +1007,32 @@ public void testWithWaypointNames() throws Exception {
     assertEquals("Ok", response.body().code());
   }
 
+  @Test
+  public void testWithWaypointNamesList() throws Exception {
```
Guardiola31337: Same as above re: `test` / `@Test`

services-matching/src/test/java/com/mapbox/api/matching/v5/MapboxMapMatchingTest.java
```diff
@@ -599,7 +599,8 @@ public void testRouteOptionsApproaches() throws Exception {
     Response<MapMatchingResponse> response = mapMatching.executeCall();
     RouteOptions routeOptions = response.body().matchings().get(0).routeOptions();
 
-    assertEquals("unrestricted;curb", routeOptions.approaches());
+    assertEquals("unrestricted", routeOptions.approachesList().get(0));
```
Guardiola31337: To keep previous test, wouldn't `FormatUtils#formatApproaches` passing `routeOptions.approachesList()` work here and return `"unrestricted;curb"`?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: `println(Double.POSITIVE_INFINITY)` -> `Infinity`
`print("unlimited".toDouble())` -> `NumberFormatException`
`print("NoSegment".toDouble())` -> `NumberFormatException`

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: fixed

korshaknn:
@LukasPaczos , @Guardiola31337  what about `waypointIndices`, can we skip them passing an empty string or a null value?

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
korshaknn: I mean it would be useful to change
`List<List<Double>>` -> `List<Bearing>` of `List<Pair>`

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -279,22 +392,65 @@ public static Builder builder() {
...
+   * <p>
+   * Most useful in combination with  steps=true and requests based on traces
+   * with high sample rates. Can be an index corresponding to any of the input coordinates,
+   * but must contain the first ( 0 ) and last coordinates' index separated by  ; .
+   * {@link #steps()}
+   * </p>
+   *
+   * @return a List of Integers representing indices to be used as waypoints
    */
   @SerializedName("waypoints")
```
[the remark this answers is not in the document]
korshaknn: removes unnecessary `Serialized` annotations.
`toJson` and `fromGson` should work, because we don't use any custom model, just Lists

korshaknn:
also 
`uuid` and `walkingOptions` are `Nullable` in `DirectionsResponse`, but `NonNull` in `RouteOptions`. Which one should I use?

services-matching/src/test/java/com/mapbox/api/matching/v5/MapboxMapMatchingTest.java
korshaknn: will add a new one

LukasPaczos:
Are there any guidelines in the Directions API docs? Is it guaranteed that we'll receive those values?

korshaknn:
@LukasPaczos , @Guardiola31337 
BTW, found a confusing situation. With old(current) logic if we pass
`radiuses("7.4;;;")` it will be parsed with `split` method to array with size `1`, so we lose empty strings. It happens because `split` removes trailing empty strings. To avoid it we need to use `split(separator, -1)` .
some of the docs has a message that `a value can be skipped with ; symbol`. With such symbols at the end of the string we might have bugs

LukasPaczos:
Great catch! That's Directions API's design that we need to take into consideration here.

korshaknn:
@LukasPaczos I changed all parse methods in util classes to use `str.split(separator, -1)`. Added tests

Guardiola31337:
@korshaknn 
> what about `waypointIndices`, can we skip them passing an empty string or a null value?

Yeah, as indicated in the docs (👀 at the `Note`)

```java
   * @param waypointIndices     which input coordinates should be treated as waypoints/separate legs
   *                            Note: coordinate indices not added here act as silent waypoints
```

we can skip them. But we need to be careful in this particular case because the valid formatting that the Directions API is expecting (imagine we have 3 waypoints) would be `waypoints=0;2` (✅) and not `waypoints=0;;2` (❌) - note that `""` trail is invalid, while it's needed for other parameters. Does that make sense? Let me know if this example isn't clear enough to understand the subtle difference. cc @danpaz 

> `uuid` and `walkingOptions` are `Nullable` in `DirectionsResponse`, but `NonNull` in `RouteOptions`. Which one should I use?

`uuid` it's in the `DirectionsResponse` and it's not always set so it should be `@Nullable`.

Walking parameters are optional 👀 https://docs.mapbox.com/api/navigation/#optional-parameters-for-the-mapboxwalking-profile so `@Nullable` as well.

Guardiola31337:
In any case, in order to verify all these "unknowns" / questions and make sure everything is working as expected, it'd be 💯 if we can create a `SNAPSHOT` and test downstream from the Navigation SDK side.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -1,52 +1,303 @@
...
+        .approachesList(approaches)
+        .build();
+
+    assertEquals(approaches, routeOptions.approachesList());
+    assertEquals("curb;;;unrestricted;unrestricted;;;curb", routeOptions.approaches());
+  }
+
+  @Test
+  public void waypointIndicesString() {
+    String indicesStr = "5;1;8;4;;;4;7;;;";
```
Guardiola31337: This is not a valid waypoint indices output 👀 https://github.com/mapbox/mapbox-java/pull/1118#issuecomment-582521995 Am I missing something? Let's sync up internally if the example wasn't clear.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -1,52 +1,303 @@
...
+    assertEquals(Integer.valueOf(8), indices.get(2));
+    assertEquals(Integer.valueOf(4), indices.get(3));
+    assertEquals(null, indices.get(4));
+    assertEquals(null, indices.get(5));
+    assertEquals(Integer.valueOf(4), indices.get(6));
+    assertEquals(Integer.valueOf(7), indices.get(7));
+  }
+
+  @Test
+  public void waypointIndicesList() {
```
Guardiola31337: See comment above re: valid waypoint indices input / output

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
korshaknn: @Guardiola31337 we check that we build a valid list even if a customer passes an invalid string. so, the list will be `(5, 1, 8, 4, null, null, 4, 7)`
and when we build a query string to make API request, we will get "5;1;8;4;;4;7"

Guardiola31337: My points is that `"5;1;8;4;;4;7"` is not a valid input for the Directions API. It should be `"1;4;5;7;8"`  - in order, not duplicated, not trailing `""` (`;;`). Is that correct @danpaz?

danpaz: @Guardiola31337 is correct, there should not be null entries in the list. 

To expand a bit further: the purpose of the `waypoints` parameter is for the user to select which route coordinates should be treated as _break_ points along the route as opposed to _via_ points. Break points receive arrival and departure instructions and create multiple legs on the route. Via points are "silent" waypoints which do not receive arrival/departure instructions and don't break up a leg. The default behavior without waypoints is to treat every waypoint as a break point. 

For example, if the user supplied 5 lon,lat coordinate pairs:

```
0,0
1,1
2,2
3,3
4,4
```

and set waypoints=0;2;4, the result would be setting the first, third, and fifth coordinates as break points.

```
0,0  <-- treat as break point
1,1  <-- treat as via point
2,2  <-- treat as break point
3,3  <-- treat as via point
4,4  <-- treat as break point
```
The result would be a route with 2 legs instead of 4 legs. Hope this explanation helps.

Guardiola31337: Is it fair to say that only `waypoints` cannot have `;;`  inside a string? We want to double check @danpaz as there are other parameters that accept trailing `""` (`";;"`).

danpaz: Yes waypoints is unique in that sense. 

It doesn't make sense to have an empty entry in waypoints, because each value is an index into the coordinates. Other parameters like `radiuses` can have empty entries because there's a default value to fall back to in that case.

korshaknn: discussed. waypoint indices can't have empty values

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -420,46 +594,100 @@ public static RouteOptions fromJson(String json) {
...
-     * The language for instructions to be in when the response is given.
+     * The language of returned turn-by-turn text instructions. The default is en (English).
+     * Must be used in conjunction with {@link RouteOptions#steps()} = true.
      *
      * @param language a string with the language which was requested in the url
      * @return this builder for chaining options together
      * @since 3.0.0
      */
-    public abstract Builder language(String language);
+    public abstract Builder language(@Nullable String language);
```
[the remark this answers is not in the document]
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -109,28 +118,73 @@ public static Builder builder() {
...
+  @Nullable
+  @Deprecated
+  public String radiuses() {
+    return FormatUtils.formatRadiuses(radiusesList());
+  }
+
+  /**
+   * The maximum distance a coordinate can be moved to snap to the road network in meters. There
+   * must be as many radiuses as there are coordinates in the request.
+   * Values can be any number greater than 0 or the string unlimited.
```
LukasPaczos: Value can also be `null`.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -109,28 +118,73 @@ public static Builder builder() {
   public abstract String language();
 
   /**
-   * The same radiuses were used during the request that resulted in this root directions response.
+   * The maximum distance a coordinate can be moved to snap to the road network in meters. There
+   * must be as many radiuses as there are coordinates in the request, each separated by ;.
+   * Values can be any number greater than 0 or the string unlimited.
```
LukasPaczos: Value can also be an empty string.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -109,28 +118,73 @@ public static Builder builder() {
...
+  public String bearings() {
+    return FormatUtils.formatBearings(bearingsList());
+  }
+
+  /**
+   * Influences the direction in which a route starts from a waypoint. Used to filter the road
+   * segment the waypoint will be placed on by direction. This is useful for making sure the new
+   * routes of rerouted vehicles continue traveling in their current direction. A request that does
+   * this would provide bearing and radius values for the first waypoint and leave the remaining
+   * values empty. Returns two comma-separated values per waypoint: an angle clockwise from true
```
LukasPaczos: This returns a list of values where first is an angle and the second is the range of degrees.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -303,15 +450,31 @@ public static Builder builder() {
...
+   * @deprecated use {@link #waypointTargetsList()}
+   */
+  @Nullable
+  @Deprecated
+  public String waypointTargets() {
+    return FormatUtils.formatPointsList(waypointTargetsList());
+  }
+
+  /**
+   * A list of coordinate pairs used to specify drop-off
```
LukasPaczos:
```suggestion
   * A list of points used to specify drop-off
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,218 @@
...
+   * @param tokens A list of objects to be joined.. Strings will be formed from the objects by
+   *               calling object.toString().
+   * @return {@link String}
+   */
+  public static @Nullable String join(@NonNull CharSequence delimiter, @Nullable List<?> tokens) {
+    if (tokens == null || tokens.size() < 1) {
+      return null;
+    }
+
+    // trailing nulls should be dropped
```
LukasPaczos: Are we sure that this is correct @korshaknn @Guardiola31337 @danpaz?

If I understand correctly, the nulls/empty strings should only be avoided for waypoints. For fields like `bearings` or `radiuses` we still want to have all of the `nulls`/empty strings be represented, even if they are trailing because the number of arguments has to much the number of coordinates.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -1,52 +1,296 @@
...
+    radiuses.add(null);
+    radiuses.add(null);
+
+    RouteOptions routeOptions = routeOptions()
+        .toBuilder()
+        .radiusesList(radiuses)
+        .build();
+
+    assertEquals(radiuses, routeOptions.radiusesList());
+    assertEquals(";;5.7;;4.4;9.9", routeOptions.radiuses());
```
LukasPaczos:
```suggestion
    assertEquals(";;5.7;;4.4;9.9;;;", routeOptions.radiuses());
```

Based on the above comment, and If I'm not mistaken, this assertion should look like this.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
korshaknn: current code works with Strings and `split` is used o parse input string. 
`split` method removes trailing empty strings, so I just implemented the same logic, but with `Lists`
please check `MapboxDirectionsUtils`, to parse Strings we use Kotlin
`someString.split(SEMICOLON.toRegex()).dropLastWhile { it.isEmpty() }`
so, we drop empty strings. (it works exactly as java `split` does).

LukasPaczos: Is that what Directions API expects though? The number of input values is going to differ in comparison to the size of the coordinate, isn't it?

Guardiola31337: Yeah @LukasPaczos is right here 👇 

> If I understand correctly, the nulls/empty strings should only be avoided for waypoints. For fields like `bearings` or `radiuses` we still want to have all of the `nulls`/empty strings be represented, even if they are trailing because the number of arguments has to much the number of coordinates.

We need to make sure the number of coordinates match with the number of coordinates (except for `waypointIndices` as discussed above).

cc @danpaz

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
Guardiola31337: Yeah, that's correct 👀 https://github.com/mapbox/mapbox-java/pull/1118#discussion_r378329478

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,218 @@
...
+public class FormatUtils {
+  /**
+   * Returns a string containing the tokens joined by delimiters.
+   *
+   * @param delimiter the delimeter on which to split.
+   * @param tokens A list of objects to be joined.. Strings will be formed from the objects by
+   *               calling object.toString().
+   * @return {@link String}
+   */
+  public static @Nullable String join(@NonNull CharSequence delimiter, @Nullable List<?> tokens) {
```
Guardiola31337: NIT
```suggestion
  @Nullable
  public static String join(@NonNull CharSequence delimiter, @Nullable List<?> tokens) {
```

Same thing applies to the rest of the methods 👇

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,218 @@
...
+    return sb.toString();
+  }
+
+  /**
+   * Useful to remove any trailing zeros and prevent a coordinate being over 7 significant figures.
+   *
+   * @param coordinate a double value representing a coordinate.
+   * @return a formatted string.
+   */
+  public static String formatCoordinate(double coordinate) {
```
Guardiola31337: Missing `@Nullable` / `@NonNull` annotation?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -0,0 +1,150 @@
...
+  private static final String COMMA = ",";
+  private static final String UNLIMITED = "unlimited";
+
+  /**
+   * Parse a String to a list of Integers.
+   *
+   * @param original an original String.
+   * @return List of Integers
+   */
+  public static @Nullable List<Integer> parseToIntegers(@Nullable String original) {
```
Guardiola31337: Same as above

```suggestion
  @Nullable
  public static List<Integer> parseToIntegers(@Nullable String original) {
```

korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
korshaknn: fixed

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,243 @@
...
+   *
+   * @param delimiter the delimiter on which to split.
+   * @param tokens A list of objects to be joined. Strings will be formed from the objects by
+   *               calling object.toString().
+   * @param removeTrailingNulls true if trailing nulls should be removed.
+   * @return {@link String}
+   */
+  @Nullable
+  public static String join(@NonNull CharSequence delimiter, @Nullable List<?> tokens,
+      boolean removeTrailingNulls) {
```
LukasPaczos: is `removeTrailingNulls` always `false`? Maybe we can remove it if that's the case? One less thing to test and maintain.

LukasPaczos:
Awesome work!

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
korshaknn: it is true for `waypointIndices`
```
waypointIndices(FormatUtils.join(";", waypointIndices, true));
```

LukasPaczos: Do we need it though? The docs are clear that we are looking for indices and passing `null`s will be invalid, and the response error will point it out probably. Removing the values might just introduce unexpected behavior.

korshaknn: I added it just to try to handle customers invalid input String.
But maybe you are right and we should not do it under the hood.

circle.yml
```diff
@@ -78,7 +78,7 @@ jobs:
       - deploy:
           name: Publish Java libraries to Bintray
           command: |
-            if [[ $CIRCLE_BRANCH == master ]] || [[ $CIRCLE_TAG == v* ]]; then
+            if [[ $CIRCLE_BRANCH == knn-options-refactoring ]] || [[ $CIRCLE_TAG == v* ]]; then
```
Guardiola31337: Revert this before merging.

korshaknn:
@LukasPaczos @Guardiola31337 during snapshot testing I found some issues with `MapboxDirections` and `annotations` inside `RouteOptions`. Please check my last [commit](https://github.com/mapbox/mapbox-java/pull/1118/commits/1ebcd68cca00c3a6aafd3636f26c7dd8bbe71800)

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -597,21 +577,31 @@ public Builder language(@Nullable Locale language) {
...
-     *                    or null which will result in no annotations being used
+     * @param annotation string referencing one of the annotation direction criteria's.
      * @return this builder for chaining options together
      * @see <a href="https://www.mapbox.com/api-documentation/navigation/#route-leg-object">RouteLeg object
      * documentation</a>
      * @since 2.1.0
      */
-    public Builder annotations(@Nullable @AnnotationCriteria String... annotations) {
-      this.annotations = annotations;
+    public Builder addAnnotation(@AnnotationCriteria @NonNull String annotation) {
```
Guardiola31337: This is breaking SemVer. What about deprecating `annotations` and add `addAnnotation`?

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -666,15 +690,32 @@ public Builder addBearing(@Nullable @FloatRange(from = 0, to = 360) Double angle
      * If no routable road is found within the radius, a {@code NoSegment} error is returned.
      * </p>
      *
-     * @param radiuses double array containing the radiuses defined in unit meters.
+     * @param radiuses a list of radiuses defined in unit meters.
      * @return this builder for chaining options together
-     * @since 1.0.0
      */
-    public Builder radiuses(@FloatRange(from = 0) double... radiuses) {
+    public Builder radiuses(@NonNull List<Double> radiuses) {
```
Guardiola31337: This is breaking SemVer.

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -788,13 +829,31 @@ public Builder radiuses(@FloatRange(from = 0) double... radiuses) {
...
+     * If provided, the list of approaches must be the same length as the list of waypoints.
+     *
+     * @param approaches empty if you'd like the default approaches,
      *                   else one of the options found in
      *                   {@link com.mapbox.api.directions.v5.DirectionsCriteria.ApproachesCriteria}.
      * @return this builder for chaining options together
-     * @since 3.2.0
      */
-    public Builder addApproaches(String... approaches) {
+    public Builder approaches(@NonNull List<String> approaches) {
```
Guardiola31337: Same as above re: breaking SemVer

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -807,50 +866,97 @@ public Builder addApproaches(String... approaches) {
...
+     * Most useful in combination with  steps=true and requests based on traces
+     * with high sample rates. Can be an index corresponding to any of the input coordinates,
+     * but must contain the first ( 0 ) and last coordinates' indices.
+     * {@link #steps()}
+     * </p>
+     *
+     * @param waypointIndices a list of coordinate indices to be used as waypoints
+     * @return this builder for chaining options together
+     */
+    public Builder waypointIndices(@NonNull List<Integer> waypointIndices) {
```
Guardiola31337: Same as above re: breaking SemVer

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -807,50 +866,97 @@ public Builder addApproaches(String... approaches) {
...
+    /**
+     * Custom names for waypoints used for the arrival instruction.
+     * Values can be any string and total number of all characters cannot exceed 500.
+     * If provided, the list of waypointNames must be the same length as the list of
+     * coordinates, but you can skip a coordinate and show its position with the null value.
+     *
+     * @param waypointNames a list of custom names for waypoints used for the arrival instruction.
+     * @return this builder for chaining options together
+     */
+    public Builder waypointNames(@NonNull List<String> waypointNames) {
```
Guardiola31337: Same as above re: breaking SemVer

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
```diff
@@ -807,50 +866,97 @@ public Builder addApproaches(String... approaches) {
...
+     * A list of points used to specify drop-off locations that are distinct from the locations
+     * specified in coordinates.
+     * The number of waypoint targets must be the same as the number of coordinates,
+     * but you can skip a coordinate with a null value.
+     * Must be used with steps=true.
+     *
+     * @param waypointTargets list of coordinate points for drop-off locations
+     * @return this builder for chaining options together
+     */
+    public Builder waypointTargets(@NonNull List<Point> waypointTargets) {
```
Guardiola31337: Same as above re: breaking SemVer

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -240,4 +241,36 @@ public static String formatPointsList(@Nullable List<Point> points) {
...
+   */
+  @Nullable
+  public static String formatAnnotations(@Nullable List<String> annotations) {
+    if (annotations == null || annotations.isEmpty()) {
+      return null;
+    }
+
+    for (String annotation : annotations) {
+      if (annotation != null
+          && !annotation.equals(DirectionsCriteria.ANNOTATION_DURATION)
```
Guardiola31337: What happens if in the future we want to add a new annotation in `DirectionsCriteria`? I believe we shouldn't check if the annotations passed in are not equal to the current ones available because if a new annotation is added we'll need to change this code. How can we avoid that situation? What do you think?

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -279,6 +284,48 @@ public void waypointTargetsList() {
     assertEquals(";;5.55,7.77;1.22,3.44;;1.55,8.99", routeOptions.waypointTargets());
   }
 
+  @Test
+  public void annotationsString() {
+    String annotationsStr = ";" + ANNOTATION_MAXSPEED + ";" + ";" + ANNOTATION_DURATION + ";";
```
Guardiola31337: This is not valid for annotations:

> InvalidInput
>
> annotations value must be one of duration, distance, speed, congestion

`congestion;duration` ✅ 
`;congestion;duration;` ❌ 

Annotations only accepts a list of valid `DirectionsCriteria` annotations separated by `;`

The same thing applies to 👇

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
korshaknn: fixed

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
korshaknn: fixed

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
korshaknn: fixed

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
korshaknn: fixed

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
korshaknn: fixed

services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
korshaknn: we can't iterate on them, they are just static constants.
I can remove that check, docs have full description according valid params.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
korshaknn: fixed

circle.yml
Guardiola31337: This still applies.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -0,0 +1,243 @@
...
+        bearingsToJoin.add(String.format(Locale.US, "%s,%s",
+            formatCoordinate(bearing.get(0)),
+            formatCoordinate(bearing.get(1))));
+      }
+    }
+    return join(";", bearingsToJoin);
+  }
+
+  /**
+   * converts the list of integer arrays to a string ready for API consumption.
```
Guardiola31337: NIT

```suggestion
   * Converts the list of integer arrays to a string ready for API consumption.
```

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -1,52 +1,337 @@
...
+  }
+
+  @Test
+  public void annotationsList() {
+    List<String> annotations = new ArrayList<>();
+    annotations.add(ANNOTATION_CONGESTION);
+    annotations.add(ANNOTATION_DISTANCE);
+    annotations.add(ANNOTATION_MAXSPEED);
+    annotations.add(ANNOTATION_SPEED);
+    annotations.add(null);
```
Guardiola31337: Not sure if we want to accept `null` and then trim them 🤔 But it's good for now.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -1,52 +1,337 @@
...
+        .waypointTargetsList(targets)
+        .build();
+
+    assertEquals(targets, routeOptions.waypointTargetsList());
+    assertEquals(";;5.55,7.77;1.22,3.44;;1.55,8.99", routeOptions.waypointTargets());
+  }
+
+  @Test
+  public void annotationsString() {
+    String annotationsStr = ANNOTATION_MAXSPEED + ";" + ANNOTATION_DURATION + ";" + ";";
```
Guardiola31337: Same here. Not sure if we want to accept unnecessary `;` and then trim them 🤔 In any case, it's good for now 👍

Guardiola31337:
@korshaknn https://github.com/mapbox/mapbox-java/pull/1118#discussion_r379494625
> we can't iterate on them, they are just static constants.
I can remove that check, docs have full description according valid params.

Yeah, before we had `@AnnotationCriteria` https://github.com/mapbox/mapbox-java/blob/6b53e5e62b32c0a0cb934b2fc363da320b69cb88/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java#L593-L596 but now we cannot control that in `List<String>` unless we create an `Annotation` type. But yeah, as you mentioned for now it's fine as it's properly documented.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
korshaknn: fixed

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
korshaknn: passed Annotations List might have `null` values. I think we can't forbid it, so we need to have logic to remove them. Also we can update docs (in a separate PR) to make it valid params clearer.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
korshaknn: see the comment above

PR #1121, opened 2020-02-20
Bearing formatting fix

### LINKED ISSUE #1120: RouteOptions parsing issuse
I was working with changes from https://github.com/mapbox/mapbox-java/pull/1118 today and noticed 2 issues so far:

1. 
I run into this NPE:
```
java.lang.NullPointerException: Attempt to invoke virtual method 'java.util.List com.mapbox.geojson.Point.coordinates()' on a null object reference
    at com.mapbox.geojson.BaseCoordinatesTypeAdapter.writePoint(BaseCoordinatesTypeAdapter.java:29)
    at com.mapbox.geojson.PointAsCoordinatesTypeAdapter.write(PointAsCoordinatesTypeAdapter.java:21)
    at com.mapbox.geojson.PointAsCoordinatesTypeAdapter.write(PointAsCoordinatesTypeAdapter.java:16)
    at com.google.gson.internal.bind.TypeAdapterRuntimeTypeWrapper.write(TypeAdapterRuntimeTypeWrapper.java:69)
    at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:97)
    at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.write(CollectionTypeAdapterFactory.java:61)
    at com.mapbox.api.directions.v5.models.AutoValue_RouteOptions$GsonTypeAdapter.write(AutoValue_RouteOptions.java:300)
    at com.mapbox.api.directions.v5.models.AutoValue_RouteOptions$GsonTypeAdapter.write(AutoValue_RouteOptions.java:41)
    at com.google.gson.Gson.toJson(Gson.java:704)
    at com.google.gson.Gson.toJson(Gson.java:683)
    at com.google.gson.Gson.toJson(Gson.java:638)
    at com.google.gson.Gson.toJson(Gson.java:618)
    at com.mapbox.api.directions.v5.models.DirectionsJsonObject.toJson(DirectionsJsonObject.java:30)
```
when initially provided w`aypoint_targets` were ";lng,lat". That's because the serialization of `Point` doesn't expect `nulls` in `BaseCoordinatesTypeAdapter`.

2.
Which brings us to issue number 2 - the newly created list of arguments do not accept nulls. An example can be the above crash, but also a way we need to provide other fields like `bearingsList`:
```
routeOptionsBuilder.bearingsList(listOf(listOf(angle, BEARING_TOLERANCE), listOf(null, null)))
```
We should instead be able to do:
```
routeOptionsBuilder.bearingsList(listOf(listOf(angle, BEARING_TOLERANCE), null))
```
but this will crash at `bearings.size()`:
https://github.com/mapbox/mapbox-java/blob/123e7cd6c7aac9b125ed2a679c09f68b6e5dadd7/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java#L681-L686
which will throw and prevent us from making a request.

We'll need to iteratively iron out those conversion issues and cover them with tests, especially the JSON conversion.

@korshaknn would you be able to look into this?

/cc @mapbox/navigation-android if you see any other incompatibilities

korshaknn:
removed `bearing` validation logic from `MapboxDirections`
now both `RouteOptions` and `MapboxDirections` use `FormatUtils.formatBearings()` to build a valid string from a list of bearings

closes #1120

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
@@ -125,12 +125,26 @@ public class FormatUtils {
 
     List<String> bearingsToJoin = new ArrayList<>();
     for (List<Double> bearing : bearings) {
-      if (bearing == null || bearing.size() == 0) {
+      if (bearing == null) {
         bearingsToJoin.add(null);
       } else {
-        bearingsToJoin.add(String.format(Locale.US, "%s,%s",
-            formatCoordinate(bearing.get(0)),
-            formatCoordinate(bearing.get(1))));
+        if (bearing.size() != 2) {
+          throw new RuntimeException("Bearing size should be 2.");
+        }
+
+        Double angle = bearing.get(0);
+        Double tolerance = bearing.get(1);
+        if (angle == null || tolerance == null) {
+          bearingsToJoin.add(null);
+        } else {
+          if (angle < 0 || angle > 360 || tolerance < 0 || tolerance > 360) {
+            throw new RuntimeException("Angle and tolerance have to be from 0 to 360.");
+          }
+
+          bearingsToJoin.add(String.format(Locale.US, "%s,%s",
+              formatCoordinate(angle),
+              formatCoordinate(tolerance)));
+        }
       }
     }
     return join(";", bearingsToJoin);
--- a/services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
+++ b/services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
@@ -8,6 +8,7 @@ import com.mapbox.api.directions.v5.models.DirectionsResponse;
 import com.mapbox.api.directions.v5.models.DirectionsRoute;
 import com.mapbox.api.directions.v5.models.LegAnnotation;
 import com.mapbox.api.directions.v5.models.RouteOptions;
+import com.mapbox.api.directions.v5.utils.ParseUtils;
 import com.mapbox.core.TestUtils;
 import com.mapbox.core.exceptions.ServicesException;
 import com.mapbox.geojson.Point;
@@ -495,6 +496,27 @@ public class MapboxDirectionsTest extends TestUtils {
         .build();
     assertEquals(";;45,90;2,90;",
         directions.cloneCall().request().url().queryParameter("bearings"));
+  }
+
+  @Test
+  public void checksParseToListOfListOfDoublesEmptyTrailing() {
+    List<List<Double>> bearings = ParseUtils.parseToListOfListOfDoubles(";;45,90;2,90;");
+
+    assertEquals(5, bearings.size());
+
+    List<Double> bearing1 = new ArrayList<>();
+    bearing1.add(45d);
+    bearing1.add(90d);
+
+    List<Double> bearing2 = new ArrayList<>();
+    bearing2.add(2d);
+    bearing2.add(90d);
+
+    assertEquals(null, bearings.get(0));
+    assertEquals(null, bearings.get(1));
+    assertEquals(bearing1, bearings.get(2));
+    assertEquals(bearing2, bearings.get(3));
+    assertEquals(null, bearings.get(4));
   }
 
   @Test
--- a/services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
+++ b/services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
@@ -471,6 +471,33 @@ public class MapboxDirectionsTest extends TestUtils {
   }
 
   @Test
+  public void addNullBearings_doesGetFormattedInUrlCorrectly() throws Exception {
+    List<Double> bearing1 = new ArrayList<>();
+    bearing1.add(45d);
+    bearing1.add(90d);
+
+    List<Double> bearing2 = new ArrayList<>();
+    bearing2.add(2d);
+    bearing2.add(90d);
+
+    List<List<Double>> bearings = new ArrayList<>();
+    bearings.add(null);
+    bearings.add(null);
+    bearings.add(bearing1);
+    bearings.add(bearing2);
+    bearings.add(null);
+
+    MapboxDirections directions = MapboxDirections.builder()
+        .destination(Point.fromLngLat(13.4930, 9.958))
+        .origin(Point.fromLngLat(1.234, 2.345))
+        .bearings(bearings)
+        .accessToken(ACCESS_TOKEN)
+        .build();
+    assertEquals(";;45,90;2,90;",
+        directions.cloneCall().request().url().queryParameter("bearings"));
+  }
+
+  @Test
   public void bearing_doesGetFormattedInUrlCorrectly() throws Exception {
     List<List<Double>> bearings = new ArrayList<>();
     bearings.add(Arrays.asList(45d, 90d));
--- a/services-geojson/src/main/java/com/mapbox/geojson/BaseCoordinatesTypeAdapter.java
+++ b/services-geojson/src/main/java/com/mapbox/geojson/BaseCoordinatesTypeAdapter.java
@@ -25,8 +25,11 @@ import java.util.List;
 abstract class BaseCoordinatesTypeAdapter<T> extends TypeAdapter<T> {
 
 
-  protected void writePoint(JsonWriter out, Point value) throws  IOException {
-    writePointList(out, value.coordinates());
+  protected void writePoint(JsonWriter out, Point point) throws  IOException {
+    if (point == null) {
+      return;
+    }
+    writePointList(out, point.coordinates());
   }
 
   protected Point readPoint(JsonReader in) throws IOException {
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
@@ -481,13 +481,10 @@ public abstract class RouteOptions extends DirectionsJsonObject {
    * Must be used with {@link RouteOptions#steps()} = true.
    * @return  a list of Points representing coordinate pairs for drop-off locations
    * @since 4.3.0
-   * @deprecated use {@link #waypointTargetsList()}
    */
+  @SerializedName("waypoint_targets")
   @Nullable
-  @Deprecated
-  public String waypointTargets() {
-    return FormatUtils.formatPointsList(waypointTargetsList());
-  }
+  public abstract String waypointTargets();
 
   /**
    * A list of points used to specify drop-off
@@ -499,9 +496,10 @@ public abstract class RouteOptions extends DirectionsJsonObject {
    * Must be used with {@link RouteOptions#steps()} = true.
    * @return  a list of Points representing coordinate pairs for drop-off locations
    */
-  @SerializedName("waypoint_targets")
   @Nullable
-  public abstract List<Point> waypointTargetsList();
+  public List<Point> waypointTargetsList() {
+    return ParseUtils.parseToPoints(waypointTargets());
+  }
 
   /**
    * To be used to specify settings for use with the walking profile.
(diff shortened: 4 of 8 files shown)


services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -125,12 +125,26 @@ public static String formatBearings(@Nullable List<List<Double>> bearings) {
...
-            formatCoordinate(bearing.get(0)),
-            formatCoordinate(bearing.get(1))));
+        if (bearing.size() != 2) {
+          throw new RuntimeException("Bearing size should be 2.");
+        }
+
+        Double angle = bearing.get(0);
+        Double tolerance = bearing.get(1);
+        if (angle == null || tolerance == null) {
+          bearingsToJoin.add(null);
```
LukasPaczos: I feel like we're safe to throw an exception here. If we chose to provide angle, we have to provide a tolerance as well. And vice versa.

services-geojson/src/main/java/com/mapbox/geojson/BaseCoordinatesTypeAdapter.java
```diff
@@ -25,8 +25,11 @@
 abstract class BaseCoordinatesTypeAdapter<T> extends TypeAdapter<T> {
 
 
-  protected void writePoint(JsonWriter out, Point value) throws  IOException {
-    writePointList(out, value.coordinates());
+  protected void writePoint(JsonWriter out, Point point) throws  IOException {
+    if (point == null) {
```
LukasPaczos: We are now choosing to ignore nulls, while we should actually be able to support them when serializing and deserializing `List`s in the `BaseCoordinatesTypeAdapter`. I think we'll need to refactor both read and write methods.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
LukasPaczos: Ah now I see that the `addBearing` method accepts nulls for angle and tolerance. Not sure if that's correct, but I guess we can leave this as is.

services-geojson/src/main/java/com/mapbox/geojson/BaseCoordinatesTypeAdapter.java
korshaknn: @LukasPaczos what logic should `writePoint` has in case of `null` Point?
also `readPoint` always returns `non null` Point or throws an exception.

LukasPaczos: > also readPoint always returns non null Point or throws an exception

That's what we'd need to refactor. Could we use `JsonToken.NULL` or a similar construct?

Guardiola31337: Is this because before `null` wasn't an option when reach serialization / deserialization and now it is? I believe the problem is that before we were always dealing with `String`s and not anymore 🤔 How serialization / deserialization was working before?

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -470,6 +470,33 @@ public void addBearing_doesGetFormattedInUrlCorrectly() throws Exception {
...
+    bearings.add(bearing2);
+    bearings.add(null);
+
+    MapboxDirections directions = MapboxDirections.builder()
+        .destination(Point.fromLngLat(13.4930, 9.958))
+        .origin(Point.fromLngLat(1.234, 2.345))
+        .bearings(bearings)
+        .accessToken(ACCESS_TOKEN)
+        .build();
+    assertEquals(";;45,90;2,90;",
```
Guardiola31337: Debugging this PR I found that `ParseUtils#parseToListOfListOfDoubles` in this concrete example is not working as expected

```java
@Test
public void checksParseToListOfListOfDoublesEmptyTrailing() {
  List<List<Double>> bearings = ParseUtils.parseToListOfListOfDoubles(";;45,90;2,90;");

  assertEquals(5, bearings.size());
}
```

It should return a list of 5 elements and it's returning a list of 4 because of not splitting with trailing empty strings https://github.com/mapbox/mapbox-java/blob/123e7cd6c7aac9b125ed2a679c09f68b6e5dadd7/services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java#L141

After changing to `original.split(SEMICOLON, -1);` other tests start failing. We should revisit the empty trailing scenarios for other fields too.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
korshaknn: ok, resolve for now

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
korshaknn: @Guardiola31337 @LukasPaczos 
when you parse a `";;45,90;2,90;"` with `ParseUtils#parseToListOfListOfDoubles` you get a list with size 4 because it removes trailing empty strings. And it worked the same way  before refactoring, `split(SEMICOLON)` is used in both cases

korshaknn: so, it works as expected

korshaknn: tested it with @kmadsen , @Guardiola31337 you are right, `bearing` string should HAVE trailing `;` if we pass null bearings. 
I'll fix logic and tests

LukasPaczos: @korshaknn is `";;45,90;2,90;"` and `";;45,90;2,90"` going to be equal in this case? I think it shouldn't be. We should take the strategy that Directions API uses:
- what happens if we pass `"45,90;"` as the request parameter? Is it considered as 2 values (1 empty)?
- what happens if we pass `"45,90;;"` as the request parameter? Is it considered as 3 values (2 empty)?
If yes, we shouldn't trim the trailing empty string.

/cc @danpaz

LukasPaczos: Oh, looks like we commented at the same time @korshaknn :) Thanks for verifying!

korshaknn: @LukasPaczos 
`";;45,90;2,90;"` will be valid for 5 coordinates
`";;45,90;2,90"` will be valid for 4 coordinates

services-geojson/src/main/java/com/mapbox/geojson/BaseCoordinatesTypeAdapter.java
korshaknn: @Guardiola31337 @LukasPaczos  I think it's only because of `waypointTargetsList()` in `RouteOptions`. 
Previously waypointTargets were stored as a String, so `PointAsCoordinatesTypeAdapter` was not used. But now they are a `List<Point>`, so `PointAsCoordinatesTypeAdapter` calls `BaseCoordinatesTypeAdapter.writePoint()`.
It is possible to store `nulls` in `waypointTargetsList`, so we need to check for a null inside `writePoint()` method

Guardiola31337: @korshaknn could we add some tests in `RouteOptions` verifying that the serialization / deserialization works 👌 (`toJson` / `fromJson`)?

LukasPaczos: That's right @korshaknn, but we still need to serialize and deserialize this `null`, not skip it. I think the inconsistency will be caught with a test that @Guardiola31337 suggested.

korshaknn: @LukasPaczos how that `null` Point should be serialized?

Guardiola31337: > how that `null` Point should be serialized?

Probably `""` (empty)? Could we run a quick test of `toJson` / `fromJson` before and after https://github.com/mapbox/mapbox-java/pull/1118 to check how was working, compare and do the same?

LukasPaczos: Empty string sounds correct.

korshaknn:
@LukasPaczos  @Guardiola31337 
I tested `RouteOptions.waypointTargets` logic before refactoring and got the next results:
```
waypointTargets("11,22;33,44;55,66") -> "waypoint_targets":"11,22;33,44;55,66"

waypointTargets("11,22;33,44;") -> "waypoint_targets":"11,22;33,44;"

waypointTargets("11,22;;55,66") -> "waypoint_targets":"11,22;;55,66"

waypointTargets(";33,44;55,66") -> "waypoint_targets":";33,44;55,66"
```
I decided to return old logic and use String to store `waypointTargets`, in that case there is no need to change serialization/deserialization logic, it won't break anything.
so there will be 2 methods in builder:
`abstract waypointTargets(String)`
`nonAbstract waypointTargetsList(List)` (it uses `FormatUtils.formatPointsList` to store as s string)

to get `waypointTargets` from `RouteOptions` we can call 
`String waypointTargets`
or `List<Point> waypointTargetsList` (it uses `ParseUtils.parseToPoints` to get a List from a string)

LukasPaczos:
The only concern I have is that if the end-users (and our SDKs) are going to rely on the `List`s implementation (because it's more convenient and that was the goal), with every call to the non-abstract `List` method we will have to parse the string and create new objects, again and again.

Did you run into any issues when trying to serialize/deserialize nulls?

korshaknn:
@LukasPaczos no, I haven't had any issues with serialization/deserialization. 
BTW with old implementation we got a string and converted it to a new List. So now we just have the same logic inside `RouteOptions`

LukasPaczos:
> BTW with old implementation we got a string and converted it to a new List. So now we just have the same logic inside RouteOptions

The question is whether we can hopefully skip one of those steps and optimize the process. If you had no issues with serialization, could we try it?

korshaknn:
if we store waypointTargets as `List<Point>`, `RouteOptions` will have 2 `List<Point>` :
`List<Point> coordinates`
and 
`List<Point> waypointTargets`
so, they both will use the same serialization/deserialization adapter. But they have to be serialized in a different way. 
`"coordinates":[[1.0,2.0],[3.0,4.0]]` - a list
`"waypoint_targets":"33,44;55,66"` - a string with semicolons

Guardiola31337:
@korshaknn 
> But they have to be serialized in a different way. 

Is this needed? Can't those be serialized / deserialized equally? Would that help solving @LukasPaczos's concern?

korshaknn:
@Guardiola31337 it was serialized in a different way before refactoring.
If we change serialization logic It may work incorrect in some cases. That's why I'm a little afraid of those changes. 
Where do we send a json? Can it break any external (API) logic?
Where do we receive json from? Currently we receive a json which stores `coordinates` as a list and `waypointTragets` as a String. What if we change logic to handle `waypointTragets` as List, but someone sends us a String?

or maybe serialization/deserialization of `RouteOptions` is just an internal logic and doesn't affect other parts of the SDK and we can  change it without any problems?

Guardiola31337:
@korshaknn 
> it was serialized in a different way before refactoring.
If we change serialization logic It may work incorrect in some cases. That's why I'm a little afraid of those changes. 
Where do we send a json? Can it break any external (API) logic?
Where do we receive json from? Currently we receive a json which stores `coordinates` as a list and `waypointTragets` as a String. What if we change logic to handle `waypointTragets` as List, but someone sends us a String?

These are valid questions 👍 

> or maybe serialization/deserialization of `RouteOptions` is just an internal logic and doesn't affect other parts of the SDK and we can  change it without any problems?

That's a great question 🤔 Events maybe? It'd be great to find out testing downstream in the Navigation SDK and 👀 if everything is working 👌 or if there are any 💥 We can do that in a separate PR, so feel free to merge here 🚀

Issue #1124, opened 2020-03-04
Add support for Matrix API calls with more than 25 coordinates


langsmith:
In certain situations and with Mapbox approval, Matrix API calls can be made with 50 coordinates instead of the normal 25 maximum. This ticket tracks the idea of adjusting the Java SDK so that the SDK supports the 50 coordinate calls.

https://github.com/mapbox/mapbox-java/blob/master/services-matrix/src/main/java/com/mapbox/api/matrix/v1/MapboxMatrix.java#L306-L312

Related:

- https://docs.mapbox.com/android/java/overview/directions-matrix/
- https://docs.mapbox.com/android/java/examples/get-a-travel-time-matrix/

cc @brandihaskins

langsmith:
@danpat 's thoughts on this

"the SDKs shouldn't have any limits in them - limits are customizable per-account and controlled server/API side"

This is probably the right way to go. Remove the coordinate size check at https://github.com/mapbox/mapbox-java/blob/master/services-matrix/src/main/java/com/mapbox/api/matrix/v1/MapboxMatrix.java#L306-L312 and let the server/API side handle limits.

zugaldia:
I'd prefer setting a default limit that is configurable by the developer to avoid running into memory issues (e.g. https://github.com/mapbox/mapbox-java/issues/893).

danpat:
@zugaldia We'd want to document that pretty clearly - it seems like it'll be confusing to have limits configurable both in the SDK, *and* by contacting sales reps.

Besides, if someone has a high-configured limit server-side, and allows high-coordinate-counts in the SDK, it seems like the issues in #893 would occur anyway, as the root cause of that bug was never really resolved that I can see.

zugaldia:
@danpat that's fair. I'm only worried that the SDK would not provide any safeguards for limits that are documented and that might surprise the dev. On the other hand, keeping SDK docs in sync with API docs have been a challenge in the past because we follow a different release pace.

@langsmith Let's get some customer feedback on your approach here https://github.com/mapbox/mapbox-java/compare/matrix-max-50-snapshot-fix.

langsmith:
> your approach here https://github.com/mapbox/mapbox-java/compare/matrix-max-50-snapshot-fix.

[...]

PR #1133, opened 2020-03-28
Adding flexibility to Matrix API builder to set custom max coordinate list size limit

langsmith:
Resolves https://github.com/mapbox/mapbox-java/issues/1124 by adding support for Matrix API calls with a coordinate list size that's greater than 25. 

After approval from the Mapbox team, a Mapbox account can make Matrix API calls with a coordinate list that's larger than the default of 25. This pr unblocks users from using the Java SDK's `MapboxMatrix.builder()` to make those 26+ calls.

This pr introduces a `.coordinateListSizeLimit()` method, which takes an `Integer` parameter. The parameter represents the new max number of coordinates per call. If the number passed through is higher than what the Mapbox account is allowed (e.g. passing `.coordinateListSizeLimit(50)` when 25 is still the max limit for that Mapbox account/token), the backend API response will respond with an error message. However, at least the Matrix API call will now be allowed/made by the Java SDK rather than an exception being thrown before the call is even made.

cc @juliawinder @chloekraw @brandihaskins @zugaldia

### DIFF
--- a/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
+++ b/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java
@@ -382,7 +382,7 @@ public abstract class MapboxDirections extends
    */
   @AutoValue.Builder
   public abstract static class Builder {
-    //TODO change List<Double> to a custom model or to a Pair
+    // TODO: change List<Double> to a custom model or to a Pair
     private List<List<Double>> bearings = new ArrayList<>();
     private List<Point> coordinates = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


langsmith:
The only reason why this isn't ready for review is because I'd like to test this work with a token that's been given permission to make calls with a coordinate list size greater than 25.

langsmith:
>I'd like to test this work with a token that's been given permission to make calls with a coordinate list size greater than 25.

✔️ . Used a token with increased permission. Calls are working as expected. This is good for review.

zugaldia:
@langsmith Approving to unblock the release.

@LukasPaczos Great if you can give it a second pass when you have a chance.

Issue #987, opened 2019-04-02
Add Turf buffer method 


danesfeder:
Turf.js has a [`Turf#buffer`](http://turfjs.org/docs/#buffer) method that would be useful for defining offline map regions.  An example use case in navigation would be defining a buffer around a route geometry for an offline download: 

![buffer](https://user-images.githubusercontent.com/8434572/55410157-618e8780-5531-11e9-83b3-d532aa730209.png)

@zugaldia mentioned that this may not be a straightforward port to java, possibly necessitating bringing in a third party [library](https://github.com/locationtech/jts) to get the job done.  We need to look into this a bit further / discuss if that makes sense.  

cc @osana @kevinkreiser

zugaldia:
Thanks @danesfeder for cutting this ticket.

The old JS Turf Buffer implementation used https://github.com/bjornharrtell/jsts which is in turn a port of the Java library https://github.com/locationtech/jts. The latest JS Turf buffer seems to use something different (`d3-geo`?). I'd love to use this ticket to understand our options:

- Implement buffer from scratch
- Bring in JTS
- Bring in another library (to be researched)

Particularly in terms of size implications. We're trying to keep MAS lightweight and this might require adding a separate artifact.

langsmith:
Based on https://github.com/Turfjs/turf/blob/master/packages/turf-buffer/index.js#L1, I believe we'll also need to port [`center()`](https://github.com/Turfjs/turf/blob/master/packages/turf-center/index.ts) before we can add the buffer method. `buffer()` could also unlock cool visuals like 👇 (Off the top of my head, I'm not sure how they're creating the white `FillLayer` area right now without `buffer()` 🤔 ). GIF is from the `Flyover Country` app: https://play.google.com/store/apps/details?id=com.ionicframework.app303739&hl=en_US

![ezgif com-resize](https://user-images.githubusercontent.com/4394910/56518473-6ae98f00-64f4-11e9-8bad-8518a564a12b.gif)

langsmith:
> Based on https://github.com/Turfjs/turf/blob/master/packages/turf-buffer/index.js#L1, I believe we'll also need to port center() before we can add the buffer method

Porting the center method is moving at https://github.com/mapbox/mapbox-java/pull/1150

langsmith:
Having this buffer method would also help with detecting whether a line has been clicked on. For example, selecting an alternate route that the Mapbox Navigation SDK displays.

[...]

PR #1150, opened 2020-05-22
Porting TurfMeasurement.center

langsmith:
This pr ports http://turfjs.org/docs/#center (https://github.com/Turfjs/turf/blob/master/packages/turf-center/index.ts) to the Mapbox Java SDK. 

Having `TurfMeasurement.center()` inches us one step closer to porting http://turfjs.org/docs/#buffer. The center method is declared [here](https://github.com/Turfjs/turf/blob/master/packages/turf-buffer/index.js#L1) in the turf.js buffer file. Buffer porting is being tracked at https://github.com/mapbox/mapbox-java/issues/987.

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
@@ -631,7 +631,7 @@ public final class TurfMeasurement {
   }
 
   /**
-   * Takes one {@link Geometry} and returns it's area in square meters.
+   * Takes one {@link Geometry} and returns its area in square meters.
    *
    * @param geometry input {@link Geometry}
    * @return area in square meters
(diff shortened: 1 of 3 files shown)


services-turf/src/main/java/com/mapbox/turf/TurfMeasurement.java
```diff
@@ -719,4 +719,67 @@ private static double ringArea(@NonNull List<Point> coordinates) {
...
+
+  /**
+   * Takes {@link FeatureCollection} and returns the absolute center
+   * of the {@link Feature}s in the {@link FeatureCollection}.
+   *
+   * @param featureCollection the single {@link FeatureCollection} to find the center of.
+   * @return a {@link Feature} with a {@link Point} geometry type.
+   * @since 5.3.0
+   */
+  public static Feature center(FeatureCollection featureCollection) {
```
pengdev, line 779: Seems this function duplicates with 
```
public static Feature center(FeatureCollection featureCollection,
                               @Nullable JsonObject properties,
                               @Nullable String id) 
```
Since the properties and id are nullable. could call `center(featureCollection, null, null)` instead.

langsmith: whoops, yea. I think I originally meant to do that. Fixed and rebased 👍

Issue #1119, opened 2020-02-04
Error when batch forward geocoding a single query


sduc:
This seems to be similar to this issue on the swift repo: https://github.com/mapbox/MapboxGeocoder.swift/issues/147

When trying to call enqueue batch call on a single query, the response will fail to be deserialized. For example when doing:

    MapboxGeocoding client = MapboxGeocoding.builder()
	    .accessToken(MAPBOX_ACCESS_TOKEN)
	    .mode(GeocodingCriteria.MODE_PLACES_PERMANENT)
	    .query("11 W 53rd St, New York, NY 10019, USA")
	    .build();
    client.enqueueBatchCall(new Callback<List<GeocodingResponse>>() {..});

The error I get when trying to batch call a single query is the following:

    java.lang.IllegalStateException: Expected BEGIN_ARRAY but was BEGIN_OBJECT at line 1 column 2 path $
        at com.google.gson.stream.JsonReader.beginArray(JsonReader.java:351)
        at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:80)
        at com.google.gson.internal.bind.CollectionTypeAdapterFactory$Adapter.read(CollectionTypeAdapterFactory.java:61)
        at retrofit2.converter.gson.GsonResponseBodyConverter.convert(GsonResponseBodyConverter.java:39)
        at retrofit2.converter.gson.GsonResponseBodyConverter.convert(GsonResponseBodyConverter.java:27)
        at retrofit2.ServiceMethod.toResponse(ServiceMethod.java:122)
        at retrofit2.OkHttpCall.parseResponse(OkHttpCall.java:217)
        at retrofit2.OkHttpCall$1.onResponse(OkHttpCall.java:116)
        at okhttp3.RealCall$AsyncCall.execute(RealCall.java:216)
        at okhttp3.internal.NamedRunnable.run(NamedRunnable.java:32)
        ... 3 more

I believe the problem here is that

    GET /geocoding/v5/{mode}/{query}.json

won't return a list when querying a single element which Gson will fail to deserialize.
It seems a list is only returned when more than one query is sent in `{query}`.

Note: this was tried on version 5.0.0

carefulAxe:
Also seeing this as well with version 5.1.0, exactly as described above. I tried with 4.9.0 and saw the same.

moritzzzzz:
[...]

PR #1160, opened 2020-07-27
Correct handling of a single element batch geocoding response

DzmitryFomchyn:
Closes https://github.com/mapbox/mapbox-java/issues/1119

Geocoding search service response has a different format depending on the query type
- if a query contains `;`, the service returns an array of results, represented as JSON array `[{...}, {...}, ...]`
- otherwise, the server returns JSON object `{...}`

This PR adds flexible JSON parsing that can handle both cases.

### DIFF
--- a/services-geocoding/src/test/java/com/mapbox/api/geocoding/v5/MapboxGeocodingTest.java
+++ b/services-geocoding/src/test/java/com/mapbox/api/geocoding/v5/MapboxGeocodingTest.java
@@ -46,6 +46,19 @@ public class MapboxGeocodingTest extends GeocodingTestUtils {
   }
 
   @Test
+  public void sanity_batchGeocodeSingleItemRequest() throws Exception {
+    MapboxGeocoding mapboxGeocoding = MapboxGeocoding.builder()
+            .mode(GeocodingCriteria.MODE_PLACES_PERMANENT)
+            .accessToken(ACCESS_TOKEN)
+            .query("20001")
+            .baseUrl(mockUrl.toString())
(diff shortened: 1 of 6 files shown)


services-geocoding/src/test/java/com/mapbox/api/geocoding/v5/MapboxGeocodingTest.java
```diff
@@ -45,6 +45,19 @@ public void sanity_batchGeocodeRequest() throws Exception {
     assertEquals(200, response.code());
   }
 
+  @Test
+  public void sanity_batchGeocodeSingleItemRequest() throws Exception {
+    MapboxGeocoding mapboxGeocoding = MapboxGeocoding.builder()
+            .mode(GeocodingCriteria.MODE_PLACES_PERMANENT)
```
cafesilencio, line 51: This test is a lot like the one below it with .mode(GeocodingCriteria.MODE_PLACES_PERMANENT) being the differentiator. 

If .mode(GeocodingCriteria.MODE_PLACES_PERMANENT) is left out there's an error message saying "Use getCall() for non-batch calls."  I wonder if it makes sense to change this message to also indicate that ".mode(GeocodingCriteria.MODE_PLACES_PERMANENT)" could be added as well in order to avoid the error condition.

DzmitryFomchyn: Makes sense, thank you! I've added clarification for the error message.

Issue #682, opened 2017-12-07
MapboxDirections: onResponse called twice


palto-blubek:
When using the directions API like so:
```
builder = (...)
MapboxDirections client = builder.build();
 client.enqueueCall(new Callback<DirectionsResponse>() {
                @Override
                public void onResponse(@NonNull Call<DirectionsResponse> call, @NonNull Response<DirectionsResponse> response) {
                    (...)
                }

                @Override
                public void onFailure(@NonNull Call<DirectionsResponse> call, @NonNull Throwable throwable) {
                    (...)
                }
            });

```
if no route can be found, then `onResponse` will be called twice.

This is highly surprising/unexpected, and in our case, caused a crash.

#### Expected Behavior
This should be called only once.

#### Current Behavior
It is called twice.

#### Possible Solution
a `return` should be added right after this line:
https://github.com/mapbox/mapbox-java/blob/master/services-directions/src/main/java/com/mapbox/api/directions/v5/MapboxDirections.java#L161

#### Steps to Reproduce
A call should be made that results in a `NoRoute` response (for instance, 2.308538,48.852698 to 120.960515,23.69781).
See above.

#### Context (Environment)
In our case this produced a crash because we use an Rx `SingleEmitter` and `onError` must not be called twice.

#### Detailed Description
See "possible solution" above.

#### Possible Implementation
See "possible solution" above.

cammace:
Thanks for reporting, I suspect you're correct with the possible solution but will need to double check this and add a fix.

PR #1166, opened 2020-08-12
Update annotation for NonNull attributes. #1087

### LINKED ISSUE #1087: Fix DirectionsRoute annotations so they match what actually comes from the API
Currently we have `distance` and `duration` (from `DirectionsRoute`) marked as `@Nullable` in the java wrapper https://github.com/mapbox/mapbox-java/blob/9bd6158292374fc6dd5f7eba33e7586ca1d76177/services-directions/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java#L50-L51 https://github.com/mapbox/mapbox-java/blob/9bd6158292374fc6dd5f7eba33e7586ca1d76177/services-directions/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java#L59-L60 and @mapbox/navigation-api has confirmed that if a valid route is returned it has to get these two values, so they can be marked as `@NonNull`.

Refs https://docs.mapbox.com/api/navigation/#route-object

JunDai:
* Fix #1087 : base on discussion in the ticket and the [comment](https://github.com/mapbox/mapbox-java/issues/1087#issuecomment-573165291) from @danpaz , we can mark `distance` and `duration` to `@NonNull`
* For the `requestUuid`, the doc clearly says it will be `NonNull`


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -353,7 +353,7 @@ public static Builder builder() {
    * @since 3.0.0
    */
   @SerializedName("uuid")
-  @Nullable
+  @NonNull
```
danpaz, line 356: Note that a NoRoute response will not include a `uuid` field. Not sure if that's a problem here or not.

JunDai: thanks for pointing it out.
I tried to request a route to an island without ferry but I still got a route :(

Do you have such case that will have a `NoRoute` response so I can try?

Guardiola31337: I think it's safe because we're not parsing if there are no routes 👀 

https://github.com/mapbox/mapbox-java/blob/baef4494dcda2812569b5af676d85b45a0dbfb21/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsResponseFactory.java#L78
1:56
https://github.com/mapbox/mapbox-java/blob/baef4494dcda2812569b5af676d85b45a0dbfb21/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsResponseFactory.java#L44-L48

In any case, definitely worth checking. You can try `2.308538,48.852698 to 120.960515,23.69781` from https://github.com/mapbox/mapbox-java/issues/682 Also I think if you reverse the order of `lat` `lon` should also return `NoRoute`. Could you confirm @danpaz?

danpaz: You can make a request between disconnected continents, like Europe and North America.

JunDai: Just verified that in `NoRoute` case, the `uuid` is an empty string.

also, discussed with @Guardiola31337 offline, we don't parse the response if there are no routes.
see [here](https://github.com/mapbox/mapbox-java/blob/baef4494dcda2812569b5af676d85b45a0dbfb21/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsResponseFactory.java#L78) and [here](https://github.com/mapbox/mapbox-java/blob/baef4494dcda2812569b5af676d85b45a0dbfb21/services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsResponseFactory.java#L44-L48).

We should be good here.

cc: @danpaz / @Guardiola31337

danpaz: Thanks for checking 👍  🙏

PR #1182, opened 2020-09-14
Intersection fields

korshaknn:
added fields
`isUrban`
`Street`

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/StepIntersection.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/StepIntersection.java
@@ -8,6 +8,7 @@ import com.google.gson.GsonBuilder;
 import com.google.gson.TypeAdapter;
 import com.google.gson.annotations.SerializedName;
 import com.mapbox.api.directions.v5.DirectionsAdapterFactory;
+import com.mapbox.api.directions.v5.DirectionsCriteria;
 import com.mapbox.geojson.Point;
 
 import java.util.List;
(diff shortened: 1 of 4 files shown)


korshaknn:
@LukasPaczos, @Guardiola31337 doc's is the most difficult part) need your help)

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/StepIntersection.java
```diff
@@ -146,6 +147,28 @@ public Point location() {
...
+  /**
+   * An object containing detailed information about the road exiting the intersection along the
+   * route. Properties in this object correspond to properties in the {@link #classes()}
+   * specification. Only available on the {@link DirectionsCriteria#PROFILE_DRIVING} profile.
+   *
+   * @return an object containing detailed road information.
+   */
+  @Nullable
+  @SerializedName("mapbox_streets_v8")
+  public abstract Street streets();
```
LukasPaczos, line 170:
```suggestion
  public abstract MapboxStreetsV8 mapboxStreetsV8();
```

This value corresponds to streets v8 sources, so I think we need to be explicit here, even if it looks ugly.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Street.java
```diff
@@ -0,0 +1,91 @@
...
+import com.mapbox.api.directions.v5.DirectionsAdapterFactory;
+import com.mapbox.api.directions.v5.DirectionsCriteria;
+
+/**
+ * An object containing detailed information about the road exiting the intersection along the
+ * route.
+ * Only available on the {@link DirectionsCriteria#PROFILE_DRIVING} profile.
+ */
+@AutoValue
+public abstract class Street extends DirectionsJsonObject {
```
LukasPaczos, line 18: Same as above.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/StepIntersection.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Street.java
korshaknn: fixed

LukasPaczos (approved):
:ship:

PR #1187, opened 2020-09-21
- RouteOptions: added `departAt` and `arriveBy`; - Direction Request: arriveBy and departAt params

RingerJK:
Pull request introduces two new fields to Directions' requests(also to `RouteOptions` as it provides fields for `MapboxDirections`): 
- _arriveBy_: the desired arrival time, formatted as a timestamp in ISO-8601 format in the local time at the route **destination**
- _departAt_: the departure time, formatted as a timestamp in ISO-8601 format in the local time at the route **origin**

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
@@ -15,6 +15,7 @@ import com.mapbox.api.directions.v5.utils.FormatUtils;
 import com.mapbox.api.directions.v5.utils.ParseUtils;
 import com.mapbox.geojson.Point;
 import com.mapbox.geojson.PointAsCoordinatesTypeAdapter;
+
 import java.util.List;
 
 /**
@@ -73,7 +74,7 @@ public abstract class RouteOptions extends DirectionsJsonObject {
    * <tt>MapboxDirections</tt> requesting object doesn't specifically set a profile.
    *
    * @return string value representing the profile defined in
-   *         {@link DirectionsCriteria.ProfileCriteria}
+   *   {@link DirectionsCriteria.ProfileCriteria}
    * @since 3.0.0
    */
   @NonNull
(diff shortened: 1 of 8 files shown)


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -493,6 +495,30 @@ public static Builder builder() {
   @Nullable
   public abstract WalkingOptions walkingOptions();
 
+  /**
+   * The desired arrival time, the local time at the route <b>destination</b>.
```
korshaknn, line 499:
```suggestion
   * The departure time, the local time at the route <b>destination</b>.
```

looks like docs for `departAt` and `arriveBy` are mixed in both `RouteOptions` and `RouteOptions.Builder`
and inside `MapboxDirections`

korshaknn (commented):
please add tests to `RouteOptions` and `MapboxDirections`

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
RingerJK: yep, nice note. They're mixed.

RingerJK:
> please add tests to `RouteOptions` and `MapboxDirections`

added 👍

korshaknn (approved):
LGTM

RingerJK:
I've re-written current PR a bit, now instead `java.util.Date` we're using `java.time.LocalDateTime`.
`Date` is error prone class and recommendations is to avoid using one. I've got some problems with time-zone when serialize-deserialize `Date`. 
btw another option to works with date-time is use [Joda time](https://www.joda.org/joda-time/) or any fork of Java 8 time package(that looks awesome) but it increase library size so I believe we don't want to use them

I'm gonna test snapshot version with Navigation SDK, but pr is ready for review 

**cc** @korshaknn @Guardiola31337 @LukasPaczos

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -254,4 +262,23 @@ public static String formatPointsList(@Nullable List<Point> points) {
     }
     return join(";", coordinatesToJoin);
   }
+
+  private static DateTimeFormatter provideDateTimeFormat() {
+    return DateTimeFormatter.ofPattern(ISO_8601_PATTERN);
+  }
```
korshaknn, line 268: do we need a separate method for a formatter?
There is only a single line of code, so I would prefer just to use
`DateTimeFormatter.ofPattern(ISO_8601_PATTERN)`inside `formatDateTimeToIso8601 ` method
or we might have an input param to save a method and make it more flexible

```
private static DateTimeFormatter provideDateTimeFormatter(String format) {
    return DateTimeFormatter.ofPattern(format);
  }
```

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -230,6 +230,40 @@ public void build_walkingAlleyBiasOptions() {
...
+    MapboxDirections directions = MapboxDirections.builder()
+      .destination(Point.fromLngLat(13.4930, 9.958))
+      .origin(Point.fromLngLat(1.234, 2.345))
+      .accessToken(ACCESS_TOKEN)
+      .arriveBy(arriveBy)
+      .build();
+
+    assertEquals(
+            directions.cloneCall().request().url().queryParameter("arrive_by"),
+            FormatUtils.formatDateTimeToIso8601(arriveBy)
```
korshaknn, line 246: I guess we also need a test that `FormatUtils.formatDateTimeToIso8601` returns a String with needed pattern 
to check  if we get a valid `iso8601` String 
something like this:
assertEquals("2020-11-12'T'22:22", FormatUtils.formatDateTimeToIso8601(input))
wdyt?

RingerJK: agree, but it has to be tests of FormatUtils, not MapboxDirections. Will add them 👍

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
RingerJK: Seems that it's redundant with implementation using `java.util.Date`. I'll move body to method

korshaknn (approved):

LukasPaczos:
Closing in favor of https://github.com/mapbox/mapbox-java/pull/1265.

PR #1207, opened 2020-10-19
Move Incidents from LegStep to RouteLeg

korshaknn:
A regular json from directions API looks like
```
{
  "routes": [
    {
      "legs": [
        {
          "steps": [],
          "incidents": [],
          ...
        }
      ],
      ...
    }
  ],
  "waypoints": [],
  ...
}
```
`Incidents` should be moved from `LegStep` to `RouteLeg`


Guardiola31337:
Did we test that Directions requests return incidents and if they're parsed properly when using the `DirectionsService`? It'd be also great to test with a `SNAPSHOT` from the Navigation SDK and if it works release a beta so we can integrate it.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteLeg.java
```diff
@@ -85,6 +85,14 @@ public static Builder builder() {
   @Nullable
   public abstract List<LegStep> steps();
 
+  /**
+   * A list of incidents that occur on the route.
```
LukasPaczos, line 89:
```suggestion
   * A list of incidents that occur on this leg.
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteLeg.java
```diff
@@ -199,6 +207,14 @@ public static RouteLeg fromJson(String json) {
      */
     public abstract Builder steps(@Nullable List<LegStep> steps);
 
+    /**
+     * A list of incidents that occur on the route.
```
LukasPaczos, line 211:
```suggestion
     * A list of incidents that occur on this leg.
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteLeg.java
korshaknn: fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteLeg.java
korshaknn: fixed

korshaknn:
@Guardiola31337 have added an `Incidents` parsing test with `MockWebServer`.

korshaknn:
@Guardiola31337 ready

Screen Shot 2020-10-20 at 16 02 41

Guardiola31337 (approved):
:shipit:

PR #1212, opened 2020-11-06
publisher production gitignore

RingerJK:
`publisher-production`' .gitignore


danswick (approved):
Awesome. Thank you for adding this!

PR #1213, opened 2020-11-06
Docs v5.6.0

RingerJK:


danswick (approved):
These changes look good to me! Once this PR has published, you can open a PR in the main docs site to update the version number here: https://github.com/mapbox/android-docs/blob/publisher-production/src/constants.json#L5. This will update the URLs here: https://docs.mapbox.com/android/java/overview/

PR #1225, opened 2021-02-08
Refresh Route: refactored

RingerJK:


services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRouteRefresh.java
```diff
@@ -0,0 +1,89 @@
...
+    return new AutoValue_DirectionsRouteRefresh.Builder();
+  }
+
+  /**
+   * A Leg Refresh is an object contain refresh data between only two two {@link DirectionsWaypoint}.
+   *
+   * @return list of {@link RouteLegRefresh} objects
+   */
+  @Nullable
+  public abstract List<RouteLegRefresh> legs();
```
LukasPaczos, line 33: Should this be a list? Isn't refresh API returning a value only for one leg?
/cc @danpat @danpaz

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRouteRefresh.java
```diff
@@ -0,0 +1,89 @@
...
+   * Create a new instance of this class by using the {@link Builder} class.
+   *
+   * @return this classes {@link Builder} for creating a new instance
+   */
+  public static Builder builder() {
+    return new AutoValue_DirectionsRouteRefresh.Builder();
+  }
+
+  /**
+   * A Leg Refresh is an object contain refresh data between only two two {@link DirectionsWaypoint}.
```
LukasPaczos, line 28:
```suggestion
   * A Leg Refresh is an object contain refresh data between two {@link DirectionsWaypoint}.
```

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
```diff
@@ -0,0 +1,29 @@
...
+
+  /**
+   * This takes the currently defined values found inside this instance and converts it to a json
+   * string.
+   *
+   * @return a JSON string which represents this DirectionsJsonObject
+   */
+  public String toJson() {
+    GsonBuilder gson = new GsonBuilder();
+    gson.registerTypeAdapterFactory(DirectionsAdapterFactory.create());
```
LukasPaczos, line 24: Should this be here?

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRouteRefresh.java
RingerJK: No, it's a list. You could run https://github.com/mapbox/mapbox-java/blob/master/samples/src/main/java/com/mapbox/samples/BasicDirectionsRefresh.java and ger instance of response

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
RingerJK: Sure, we need that because using Annotations which is a part of DirectionsRoute

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRouteRefresh.java
LukasPaczos: I see, but the list will always have only one element since we're defining the leg index as an argument, is that right?

RingerJK: param `leg index` means "current leg index", so the response contains annotations for current and following legs(annotations for passed legs aren't present)

LukasPaczos: gotcha, thanks :+1:

LukasPaczos (commented):
Could you add one test for a multi-legged route? (I don't see one currently but maybe I'm missing it). Otherwise lgtm :+1:

RingerJK:
> Could you add one test for a multi-legged route? (I don't see one currently but maybe I'm missing it). Otherwise lgtm 👍

Nice suggestion 🚀 added to `BasicDirectionsRefresh`

LukasPaczos:
@RingerJK could that be a unit test instead?

LukasPaczos (approved):
:ship:

PR #1242, opened 2021-03-24
Add snapping closures to a response options

korshaknn:
SnappingClosures options was not added to a response options. This PR fixes it

we should cover such cases with tests https://github.com/mapbox/mapbox-java/issues/1243


LukasPaczos (approved):
Do we have a test that also needs to be updated?

korshaknn:
@LukasPaczos no, we don't. DirectionsResponseFactory is not covered. I've cut a ticket https://github.com/mapbox/mapbox-java/issues/1243

PR #1250, opened 2021-03-29
Adding closures to route leg.

### LINKED ISSUE #1249: Add closures to route leg
As can be seen here: https://github.com/mapbox/api-documentation/pull/1105#pullrequestreview-596905970 there should be some closure data in the RouteLeg under certain circumstances.  The closures item needs to be added to the RouteLeg class.

cafesilencio:
Resolves #1249 by adding closures to the `RouteLeg` class.


kmadsen:
Let's add a test with real json response, verify the models are reading the values from json.

After merging we can cut a `v5.9.0-alpha.3`
https://github.com/mapbox/android-internal/blob/master/releases/mapbox-java.md 

We could also tell BMW to try watching the new auto snapshots. `v5.9.0-SNAPSHOT` should automatically get their updates

services-directions-models/src/test/resources/directions-with-closure.json
```diff
@@ -0,0 +1,332 @@
+{
```
kmadsen, line 1: should we rename the file to "fit in" `directions_v5_closure.json`

Screen Shot 2021-03-30 at 1 35 18 PM

cafesilencio: renamed.

kmadsen (approved):
🚢

PR #1265, opened 2021-06-18
refactor `RouteOptions` and make them a `MapboxDirections` parameter, fix smaller issues

LukasPaczos:
Closes https://github.com/mapbox/mapbox-java/issues/1262.

This PR:
- Refactored `MapboxDirections` to accept `RouteOptions` as an argument that defines request parameters.
- Moved the `requestUuid` object out of `RouteOptions` and into `DirectionsRoute` since the UUID is not a request parameter, it's a response value.
- Adds `enableRefresh` to `RouteOptions` since it is a route request parameter.
- Removed a separate `WalkingOptions` in favor of storing the parameter directly in the `RouteOptions`.
- Exposed `arriveBy` and `departAt` arguments, superseding https://github.com/mapbox/mapbox-java/pull/1187.
- Exposed `ANNOTATION_CONGESTION_NUMERIC` parameter.
- Updated documentation for various route request parameters.
- Exposes `DirectionsResponse#fromJson(json, route options, request uuid)` that allows developers to provide information about the original route request which response they are deserializing so that Nav SDK can operate correctly.
-  Exposed `DirectionsRoute#fromJson(json, route options, request uuid)` for same reason as above.
- Exposed `RouteOptions#fromUrl` utility function that creates the object from the request URL which might be useful when paired with the above functions.
- Added default values for `RouteOptions`: base URL, user, and geometry encoding (polyline6).
- All comma or semicolon separated lists had a `List` overload in `RouteOptions` besides `coordinates`, now this is aligned and there is `RouteOptions#coordinates(string)` and `RouteOptions#coordinatesList(list)`. :warning: If there are routes that were previously serialized with `toJson` methods and contained `RouteOptions`, they will not be possible to deserialize anymore since coordinates are now stored as a string and not an array. This breaking change can typically impact test fixtures.
- Fixed an issue where coordinates were rounded to 6 decimal places instead of 7 decimal places that the GeoJSON specification expects.
- Changed the default encoding for `MapMatching` to be polyline6 and aligns it with regular route requests.


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java
```diff
@@ -163,15 +169,44 @@ public static Builder builder() {
...
+   * @param requestUuid  UUID of the request found in the body of the original response,
+   *                     see "response.body.uuid"
+   * @return a new instance of this class defined by the values passed inside this static factory
+   *   method
+   * @see RouteOptions#fromUrl(java.net.URL)
+   * @see RouteOptions#fromJson(String)
+   */
+  public static DirectionsRoute fromJson(
+    @NonNull String json, @Nullable RouteOptions routeOptions, @Nullable String requestUuid) {
+    GsonBuilder gson = new GsonBuilder();
```
RingerJK, line 200: nit
```suggestion
  public static DirectionsRoute fromJson(
    @NonNull String json, @Nullable RouteOptions routeOptions, @Nullable String requestUuid
  ) {
    GsonBuilder gson = new GsonBuilder();
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -17,189 +16,201 @@
...
+   * <p>
+   * Elements of the string should have a comma separated double values, for example:
+   * ";;10.1,47.3,33.09,79.111;84.45,45.4;"
    *
    * @param original an original String.
-   * @return List of List of Doubles
+   * @return List of List of Doubles or null if the original string is null
    */
   @Nullable
   public static List<List<Double>> parseToListOfListOfDoubles(@Nullable String original) {
```
RingerJK, line 201: I'd restrict `List<List<Double>>` at all, and replace it with `List<Point>` in all places of mapbox-java

LukasPaczos: This is actually used for `bearings` parameter which is a pair of doubles. Those are different than points and unfortunately, we can't use the same method.

services-matching/src/main/java/com/mapbox/api/matching/v5/models/MapMatchingMatching.java
```diff
@@ -252,6 +262,15 @@ public DirectionsRoute toDirectionRoute() {
      */
     public abstract Builder voiceLanguage(@Nullable String voiceLanguage);
 
+    /**
+     * The universally unique identifier of the request that produced this route.
+     *
+     * @param requestUuid uuid
+     * @return this builder for chaining options together
+     */
+    @NonNull
```
RingerJK, line 271: should be `@NonNull` added to every `Builder`'s methods?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
RingerJK: 👍 my bad

Should we add a wrapper for bearings? that list of beings looks like `List<Bearings>`. It would be good for documentation as well: we can add javadocs for each value of `Beraings`

LukasPaczos:
Here's a Nav SDK PR that integrates the changes from this PR which might be helpful during the review process to visualize the scope of breaking changes - https://github.com/mapbox/mapbox-navigation-android/pull/4526.

services-matching/src/main/java/com/mapbox/api/matching/v5/models/MapMatchingMatching.java
LukasPaczos: I don't really want to touch too many more classes in this PR unless I have to. We can follow up separately.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/ParseUtils.java
```diff
@@ -17,189 +17,200 @@
...
-   * Parse a String to a list of list of Doubles.
+   * Parse a semicolon separated String to a list of lists of doubles.
+   * If separation finds empty strings, those will be added as nulls to the resulting list.
+   * <p>
+   * Elements of the string should have a comma separated double values, for example:
+   * ";;10.1,47.3,33.09,79.111;84.45,45.4;"
    *
    * @param original an original String.
-   * @return List of List of Doubles
+   * @return List of List of Doubles or null if the original string is null
```
RingerJK, line 198:
```suggestion
   * @return List<Bearing> or null if the original string is null
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Bearing.java
```diff
@@ -0,0 +1,85 @@
...
+    public abstract Builder angle(double angle);
+
+    /**
+     * Range of degrees by which the angle can deviate.
+     *
+     * @param degrees value
+     * @return this builder
+     */
+    @NonNull
+    public abstract Builder degrees(double degrees);
```
RingerJK, line 75:
```suggestion
    public abstract Builder degrees(@FloatRange(from = 0, to = 360) double degrees);
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Bearing.java
```diff
@@ -0,0 +1,85 @@
...
+  public abstract static class Builder {
+
+    /**
+     * Clockwise value from true north between 0 and 360.
+     *
+     * @param angle value
+     * @return this builder
+     */
+    @NonNull
+    public abstract Builder angle(double angle);
```
RingerJK, line 66:
```suggestion
    public abstract Builder angle(@FloatRange(from = 0, to = 360) double angle);
```

RingerJK (approved):
LGTM 👍 few nits

LukasPaczos:
Additional changes:
- Added a `Bearing` object that represents a pair of values for `{angle, degrees}`.
- As per suggestion from @danpat, removed route request parameters verification in favor of letting the service respond with a correct error. Theoretically, this makes sense and would not require library updates if rules for a specific parameter were relaxed, for example, if `degrees` tolerance was relaxed from 180° to 270° we'd have to issue a new version of the library because there's a runtime check in place that throws an exception if the value exceeds 180°. Now, we won't have to do that, and we won't need to track minor changes like these that might occasionally be deployed. We have to ensure that the service always responds with actionable errors cc @kshehadeh @PureW @mandeepsandhu. Is this also how Mapbox Directions Swift approaches the problem @mapbox/navigation-ios?

We will still add runtime checks in the SDK that verifies if a feature was requested, like if traffic is enabled in the SDK but developers didn't request congestion annotation. We won't check the validity of the request params themselves though.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Bearing.java
LukasPaczos: Refs https://github.com/mapbox/mapbox-java/pull/1265#issuecomment-865887864, we are not validating request parameters.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
```diff
@@ -88,42 +93,53 @@
...
-   * The distance, in meters, between each pair of coordinates.
+   * The distance between each pair of coordinates in meters.
    *
    * @since 2.1.0
    */
   public static final String ANNOTATION_DISTANCE = "distance";
 
   /**
-   * The speed, in km/h, between each pair of coordinates.
+   * The speed between each pair of coordinates in meters per second.
```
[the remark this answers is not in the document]
LukasPaczos: The text now matches Directions API docs, so I think it was wrong before.

cc @danpat

services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsResponseFactory.java
```diff
@@ -26,13 +26,8 @@
...
           .routes(generateRouteOptions(response))
           .build(),
-        new okhttp3.Response.Builder()
-          .code(200)
-          .message("OK")
-          .protocol(response.raw().protocol())
-          .headers(response.headers())
-          .request(response.raw().request())
-          .build());
+        response.raw()
```
Guardiola31337, line 29: ❤️

Guardiola31337 (commented):
Overall this is looking good. Great work @LukasPaczos 💪 

Left some comments.

Also, we should probably list the breaking changes in the release notes as we don't have an automated tool to detect breaking changes in `mapbox-java`.

services-core/src/main/java/com/mapbox/core/MapboxService.java
```diff
@@ -21,7 +21,12 @@
  * @since 1.0.0
  */
 public abstract class MapboxService<T, S> {
-  protected static final int MAX_URL_SIZE = 1024 * 8;
+
+  /**
+   * If the url length exceeds this value, the POST method is chosen by default.
+   */
+  public static final int MAX_URL_SIZE = 1024 * 8;
```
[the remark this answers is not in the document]
LukasPaczos: I wanted to link it from the public docs for `get`/`post` method flag so figure why not making the info public as well.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsResponse.java
```diff
@@ -126,13 +125,50 @@ public static Builder builder() {
    *   method
    * @since 3.0.0
    */
-  public static DirectionsResponse fromJson(String json) {
+  public static DirectionsResponse fromJson(@NonNull String json) {
```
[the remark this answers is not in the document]
LukasPaczos: I'd like to avoid any additional changes here, let's audit separately.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsResponse.java
```diff
@@ -126,13 +125,50 @@ public static Builder builder() {
...
+   * @param routeOptions options that were used during the original route request
+   * @param requestUuid  UUID of the request found in the body of the original response,
+   *                     see "response.body.uuid"
+   * @return a new instance of this class defined by the values passed inside this static factory
+   *   method
+   * @see RouteOptions#fromUrl(java.net.URL)
+   * @see RouteOptions#fromJson(String)
+   */
+  public static DirectionsResponse fromJson(
+    @NonNull String json, @Nullable RouteOptions routeOptions, @Nullable String requestUuid) {
```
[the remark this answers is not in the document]
LukasPaczos: > Confirming here: Is this needed to inject the RouteOptions and the requestUuid from a DirectionsResponse that was generated by not using mapbox-java?

Yes, it is needed for the Nav SDK, and Nav SDK only. That's why the parameters are optional to not break compatibility for developers that do not use `mapbox-java` with the Nav SDK.

> I'm not seeing it in mapbox/mapbox-navigation-android#4526

The options and UUID is injected automatically for routes requested via `mapbox-java` so we don't need to do it manually in the Nav SDK. If a route was serialized from a `mapbox-java` object, the data is already there. I can audit the test fixtures in the Nav SDK  again to see if there are any routes that were serialized from a raw response.

korshaknn (approved):
Great work! 👍

PR #1267, opened 2021-06-29
Do not serialize access tokens

LukasPaczos:
Ensures that access tokens are not serialized in order to avoid often unintentional leaks of the data when paired with other tools and utilities that serialize and store either the whole route response, route, or only the options. Unfortunately, due to the limitations of the code generation tools used we need to make the token field nullable until we find a better solution.

I also tried upgrading the https://github.com/rharter/auto-value-gson library that is used in this project to the latest version but it didn't seem to have the features needed to support a non-nullable property that is also not serializable and initialize it externally. We can cut a feature request, contribute, or think about a different solution in the future.

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
@@ -14,6 +14,7 @@ import com.mapbox.api.directions.v5.DirectionsCriteria;
 import com.mapbox.api.directions.v5.utils.FormatUtils;
 import com.mapbox.api.directions.v5.utils.ParseUtils;
 import com.mapbox.geojson.Point;
+import com.ryanharter.auto.value.gson.Ignore;
 import java.io.UnsupportedEncodingException;
 import java.net.URL;
 import java.net.URLDecoder;
@@ -366,11 +367,15 @@ public abstract class RouteOptions extends DirectionsJsonObject {
 
   /**
    * A valid Mapbox access token used to making the request.
+   * <p>
+   * Avoiding to provide a token will most-likely result in a failure, however,
+   * it's annotated as nullable to prevent serialization of tokens.
    *
    * @return a string representing the Mapbox access token
    */
   @SerializedName("access_token")
-  @NonNull
+  @Ignore(Ignore.Type.SERIALIZATION)
+  @Nullable
   public abstract String accessToken();
 
   /**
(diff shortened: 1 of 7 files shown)


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -672,12 +677,18 @@ public static Builder builder() {
...
    *
    * @param json a formatted valid JSON string defining a RouteOptions
+   * @param accessToken the Mapbox Access Token
    * @return a new instance of this class defined by the values passed inside this static factory
    *   method
    * @see #fromUrl(URL)
    */
   @NonNull
-  public static RouteOptions fromJson(String json) {
+  public static RouteOptions fromJson(@NonNull String json, @Nullable String accessToken) {
```
korshaknn, line 686: just to clarify, is this method used anywhere except tests?
will we always have `accessToken=null` in the `DirectionsResponse `?

Guardiola31337:
> I also tried upgrading the https://github.com/rharter/auto-value-gson library that is used in this project to the latest version but it didn't seem to have the features needed to support a non-nullable property that is also not serializable and initialize it externally. We can cut a feature request, contribute, or think about a different solution in the future.

@rharter First of all, thanks for the great library! Do you know off the top of your head if above is possible with some workaround or if not if would be feasible and how big of a lift would be? We're going to create a ticket in https://github.com/rharter/auto-value-gson but wanted to hear your initial thoughts first as creator of the extension. Looking forward to hearing from you. Thanks again! 🙇‍♂️

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
LukasPaczos: This is a utility so that devs don't need to do `fromJson.toBuilder.accessToken.build` and it's clear that `accessToken` needs to be provided again, it's not serialized.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -272,6 +275,16 @@ public void routeOptionsList_toJson() {
...
   }
 
+  @Test
+  public void routeOptions_toJson_tokenNotSerialized() {
+    String json = routeOptions().toJson();
+
+    Gson gson = new Gson();
+    JsonObject object = gson.fromJson(json, JsonObject.class);
+
+    Assert.assertNull(object.get("access_token"));
```
Guardiola31337, line 285: NIT

```suggestion
    assertNull(object.get("access_token"));
```

Guardiola31337 (approved):

rharter:
I don't have any specific approaches to solve the problem above. It sounds like the desire is really to have AVG deserialize a populated builder that the user can then provide an access token to, since you wouldn't be able to deserialize a valid object with a missing, but non-nullable, property. Offhand that would be my first inclination. The above suggested API looks good to me, internally the fromJson method could just deserialize a package private object like a builder and construct the route options by combining that with the provided token.

LukasPaczos:
Thanks for chiming in @rharter! The idea of deserializing into an intermediate, private structure gave me the idea to deserialize into a generic `JsonObject` first, append the token, and then transform the structure into the proper target object. This is not as safe as a dedicated structure but offers the same functionality with hopefully smaller maintenance needs. Up for discussion in https://github.com/mapbox/mapbox-java/pull/1270.

Guardiola31337:
Thanks a lot @rharter for your input. Really appreciate it 🙇‍♂️ 

As @LukasPaczos mentioned take a look at https://github.com/mapbox/mapbox-java/pull/1270 if you're curious about the solution we opted for 🚀 

Thanks again!

PR #1270, opened 2021-07-07
make the access token non-nullable in RouteOptions

LukasPaczos:
This refactors the access token in the RouteOptions to be non-nullable again by first deserializing the json into a generic data structure, appending the provided token, and only then transforming into the type-safe RouteOptions object.

Refs https://github.com/mapbox/mapbox-java/pull/1267.

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java
@@ -5,6 +5,7 @@ import androidx.annotation.Nullable;
 import com.google.auto.value.AutoValue;
 import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
+import com.google.gson.JsonObject;
 import com.google.gson.TypeAdapter;
 import com.google.gson.annotations.SerializedName;
 import com.mapbox.api.directions.v5.DirectionsAdapterFactory;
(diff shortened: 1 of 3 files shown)


services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/DirectionsRouteTest.java
```diff
@@ -73,4 +77,66 @@ public void directionsRoute_doesContainOptionsAndUuid() throws Exception {
...
+      .accessToken("token")
+      .build();
+    String uuid = "123";
+    DirectionsRoute route = DirectionsRoute.fromJson(json, options, uuid);
+
+    String newRouteJson = route.toJson();
+
+    DirectionsRoute newRoute = DirectionsRoute.fromJson(newRouteJson);
+
+    assertEquals(route, newRoute);
```
Guardiola31337, line 124:
```suggestion
    DirectionsRoute.fromJson(newRouteJson);
```

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/DirectionsRouteTest.java
```diff
@@ -73,4 +77,66 @@ public void directionsRoute_doesContainOptionsAndUuid() throws Exception {
...
+      "Provided serialized route does not contain RouteOptions. "
+        + "Use DirectionsRoute#fromJson(json) instead.");
+    String json = loadJsonFixture("directions_v5-with-closure_precision_6.json");
+    DirectionsRoute route = DirectionsRoute.fromJson(json);
+
+    String newRouteJson = route.toJson();
+
+    DirectionsRoute newRoute = DirectionsRoute.fromJson(newRouteJson, "token");
+
+    assertEquals(route, newRoute);
```
Guardiola31337, line 140:
```suggestion
    DirectionsRoute.fromJson(newRouteJson, "token");
```

Guardiola31337 (commented):
Left some minor comments 💪

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsRoute.java
```diff
@@ -165,26 +166,66 @@ public static Builder builder() {
...
+   * If you're using the provided route with the Mapbox Navigation SDK, also see
+   * {@link #fromJson(String, RouteOptions, String)}.
    *
    * @param json a formatted valid JSON string defining a GeoJson Directions Route
    * @return a new instance of this class defined by the values passed inside this static factory
    *   method
-   * @see #fromJson(String, RouteOptions, String)
-   * @since 3.0.0
    */
   public static DirectionsRoute fromJson(@NonNull String json) {
```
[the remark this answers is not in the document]
LukasPaczos: This actually is not true, Gson will return `null` if the input string is `null` or empty.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/DirectionsRouteTest.java
LukasPaczos: I don't think I understand this suggestion :thinking: could you explain?

Guardiola31337: It's simple the `assert` is never executed 😅 

What makes the test to pass / throw the exception is `DirectionsRoute.fromJson(newRouteJson);` the rest is unnecessary.

LukasPaczos: lol, good catch :+1:

Guardiola31337 (approved):
Thanks for addressing the comments 🚀

Issue #1275, opened 2021-07-27
Remove access token from RouteOptions


LukasPaczos:
Refs https://github.com/mapbox/mapbox-java/pull/1265, https://github.com/mapbox/mapbox-java/pull/1267, and https://github.com/mapbox/mapbox-java/pull/1270 where we tightened the `RouteOptions` integration while also stopped serializing the access token for security reasons. In https://github.com/mapbox/mapbox-navigation-android/issues/4662#issuecomment-886719332 we were discussing the implications and potential improvements.

We should be able to entirely remove the access token from the `RouteOptions` to avoid the problems of deserialization and usability of route objects. Each entity that does need to make a resources request using a route reference (like annotations refresh or reroute) will most likely get a token from a different source, the route object shouldn't need to transport it.

PR #1276, opened 2021-07-27
Removed access token from RouteOptions

LukasPaczos:
Closes https://github.com/mapbox/mapbox-java/issues/1275. Moves the `accessToken` parameter from `RouteOptions` to `MapboxDirections`.

### DIFF
--- a/services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
+++ b/services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
@@ -43,11 +43,15 @@ public class MapboxDirectionsTest extends TestUtils {
   private static final String DIRECTIONS_V5_ANNOTATIONS_FIXTURE = "directions_annotations_v5.json";
   private static final String DIRECTIONS_V5_NO_ROUTE = "directions_v5_no_route.json";
   private static final String DIRECTIONS_V5_SPEED_LIMIT = "directions_v5_speedlimit.json";
-  private static final String DIRECTIONS_V5_MAX_SPEED_ANNOTATION = "directions_v5_max_speed_annotation.json";
-  private static final String DIRECTIONS_V5_BANNER_INSTRUCTIONS = "directions_v5_banner_instructions.json";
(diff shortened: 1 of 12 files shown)


LukasPaczos:
Integration of this change can be seen in https://github.com/mapbox/mapbox-navigation-android/pull/4670.

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -544,14 +589,15 @@ public void callForUrlLength_longUrl() {
...
       .routeOptions(
         RouteOptions.builder()
           .baseUrl(mockUrl.toString())
           .coordinatesList(fakeCoordinates)
           .profile(DirectionsCriteria.PROFILE_DRIVING_TRAFFIC)
-          .accessToken("token")
           .build()
-      );
+      )
+      .accessToken("token");
```
korshaknn, line 600: token set twice

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -669,13 +723,21 @@ public void depart_at() throws Exception {
...
+      .accessToken("token")
       .routeOptions(routeOptions.toBuilder().baseUrl(mockUrl.toString()).build())
       .build();
 
     assertEquals("true",
       mapboxDirections.cloneCall().request().url().queryParameter("enable_refresh"));
   }
 
+  @Test(expected = Exception.class)
+  public void token_required() throws Exception {
```
korshaknn, line 735: do we provide any readable message?
can we test it?

LukasPaczos: We can't this is implemented directly in the `AutoValue` library - if there's a non-nullable field that wasn't initialized, `AutoValue` will throw an exception saying which field is missing. I added the test mostly for us to remember to keep the `accessToken` as a `non-nullable field in the future.

RingerJK (approved):
LGTM

korshaknn (approved):

PR #1279, opened 2021-08-16
Add support of layer to RouteOptions

SiarheiFedartsou:
.


services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/FormatUtils.java
```diff
@@ -169,4 +169,27 @@ public static String formatPointsList(@Nullable List<Point> points) {
...
   }
+
+  /**
+   * Converts list of Z-levels to a string ready for API consumption.
+   *
+   * @param zLevels a list representing Z-levels
+   * @return a formatted string with semicolon separated integers
+   */
+  @Nullable
+  public static String formatZLevels(@Nullable List<Integer> zLevels) {
```
LukasPaczos, line 180:
```suggestion
  public static String formatIntegers(@Nullable List<Integer> integers) {
```
this can be generic, there's not logic specific to z-levels in this function

services-directions-models/src/test/resources/route_options_v5.json
```diff
@@ -7,6 +7,7 @@
   "language": "ru",
   "radiuses": ";unlimited;5.1",
   "bearings": "0,90;90,0;",
+  "layers": "-42;;0",
```
LukasPaczos, line 10: Could you also add this to the other file? Refs https://github.com/mapbox/mapbox-java/issues/1286.

LukasPaczos (approved):

services-directions-models/src/test/resources/route_options_v5.json
SiarheiFedartsou: done

PR #1282, opened 2021-08-23
add metadata deserializers

### LINKED ISSUE #1272: Support metadata response from Directions API
The Directions API supports a `metadata=true` parameter that provides versioning information that looks like the following:

```
"metadata": {
    "map": {
      "tileset_version": "2021_07_14-03_00_00"
    }
  }
```

We should update the directions client to support both the request parameter as well as its response.

abhishek1508:
Fixes #1272


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Metadata.java
```diff
@@ -0,0 +1,90 @@
...
+    return new AutoValue_Metadata.Builder();
+  }
+
+  /**
+   * Map containing a key/value pair of all the properties inside `map` element.
+   * @return map of key/value pairs
+   */
+  @Nullable
+  @SerializedName("map")
+  public abstract Map<String, String> versionMap();
```
stari4ek, line 34: metadata response does not limit this key-valued container to be **versions** only. While right now all possible values are versions, I'd expect that other non-version keys might be added also.
is there a better name to be more generic? (`mapProperties`? `mapMetadata`? `mapInfo`? having key name as `map` and java structure as `Map` does not help here)

services-directions-models/src/test/resources/route_options_v5.json
```diff
@@ -27,5 +27,6 @@
   "walkway_bias": -0.2,
   "arrive_by": "2021-01-01'T'01:01",
   "depart_at": "2021-02-02'T'02:02",
-  "enable_refresh": true
+  "enable_refresh": true,
+  "metadata": true
```
LukasPaczos, line 31: This also should be added to `services-directions/src/test/resources/route_options_v5.json`. Unfortunately, those are currently duplicated.

LukasPaczos (commented):
Could you update the docs based on the comments above, that the metadata map is generic and can carry more than version information?

services-directions-models/src/test/resources/route_options_v5.json
abhishek1508: It is already added 😉

abhishek1508:
> Could you update the docs based on the comments above, that the metadata map is generic and can carry more than version information?

I like `mapInfo` better. Will update the name and also the docs based comments

stari4ek (approved):
lgtm from api side. thank you.

PR #1294, opened 2021-09-15
add active direction to banner component

abhishek1508:
Adding `active_direction` to `BannerComponents` which was added in directions response based on valhalla routing engine. This is required for refactoring turn lane.


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
```diff
@@ -250,6 +250,15 @@ public static Builder builder() {
   @Nullable
   public abstract Boolean active();
 
+  /**
+   * A property shows which of the lane's {@link BannerComponents#directions()} is applicable
+   * to the current route, when there is more than one.
```
LukasPaczos, line 255: Could you paste the full description from the Directions API docs, including the given example?

LukasPaczos (approved):

Issue #1293, opened 2021-09-13
Expose include-hov/hot params in `RouteOptions`


Guardiola31337:
> We already have `exclude` parameters, e.g., `exclude=toll`.  These parameters can be thought of as, "we will route onto tolls by default; to avoid tolls you have to opt-out using `exclude=toll`".
>
> For hov/hot I'm proposing a new `include` key with these possible values:
>`include=hov2`
>`include=hov3`
>`include=hot`
>
> These parameters can be thought of as, "we will avoid routing onto hov/hot by default; if you want to route onto hov/hot you have to opt-in using `include=hov/hot`".

cc @browndp08 @ktatterso @kelseylivingston @zugaldia

Guardiola31337:
cc'ing @1ec5 for visibility on the iOS side

1ec5:
mapbox/mapbox-directions-swift#592 tracks the corresponding changes in Swift.

PR #1296, opened 2021-09-16
Expose include-hov/hot parameters and add exclude list option

Guardiola31337:
- Fixes https://github.com/mapbox/mapbox-java/issues/1293
- Adds `exclude` list option https://github.com/mapbox/mapbox-java/pull/1296#discussion_r714831962

Opening this as a `Draft` until Directions team deploys `include` params from their side in production. For now, we can test it in Staging using a `SNAPSHOT` from https://github.com/mapbox/mapbox-navigation-android

cc @browndp08 @ktatterso @kelseylivingston @zugaldia @jill-cardamon


services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
```diff
@@ -178,6 +178,21 @@
    */
   public static final String EXCLUDE_RESTRICTED = "restricted";
 
+  /**
+   * TBD.
```
Guardiola31337, line 182: @ktatterso could you share the docs?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
```diff
@@ -178,6 +178,21 @@
...
+   */
+  public static final String INCLUDE_HOV2 = "hov2";
+
+  /**
+   * TBD.
```
Guardiola31337, line 187: @ktatterso could you share the docs?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -353,6 +353,26 @@ public static Builder builder() {
   @Nullable
   public abstract String exclude();
 
+  /**
+   * Include certain road types from routing. The default is to not include anything from the
+   * profile selected. The following include flags are available for each profile:
```
Guardiola31337, line 358: @ktatterso is this accurate? If not, could you share the docs?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
```diff
@@ -178,6 +178,21 @@
...
+   */
+  public static final String INCLUDE_HOV3 = "hov3";
+
+  /**
+   * TBD.
```
Guardiola31337, line 192: @ktatterso could you share the docs?

ktatterso:
@Guardiola31337 The doc changes for `include` are here:  https://github.com/mapbox/api-documentation/pull/1237

The api-valhalla changes for this parameter were only approved a couple hours ago.  They will be released on Monday (if things go according to schedule).  At which point I will merge these doc changes into master.

As you might imagine, we don't release documentation changes before the feature has been released.  Hence why I'm waiting.

Here are the "pretty" docs from that PR:
image

zugaldia:
API support for this feature has landed, we're good to proceed merging this PR and issuing a release.

/cc: @browndp08

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
LukasPaczos:
```suggestion
   * Include certain road types in routing. By default, none of the road types listed below are included. The following include flags are available for each profile:
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1195,6 +1215,28 @@ public Builder annotationsList(@Nullable List<String> annotations) {
     @NonNull
     public abstract Builder exclude(@Nullable @DirectionsCriteria.ExcludeCriteria String exclude);
 
+    /**
+     * Include certain road types from routing. The default is to not include anything from the
+     * profile selected. The following include flags are available for each profile:
```
LukasPaczos, line 1220: Same as above.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1195,6 +1215,28 @@ public Builder annotationsList(@Nullable List<String> annotations) {
...
+     * {@link DirectionsCriteria#PROFILE_WALKING}: No includes supported
+     * <p>
+     * {@link DirectionsCriteria#PROFILE_CYCLING}: No includes supported
+     *
+     * @param include a string matching one of the {@link DirectionsCriteria.IncludeCriteria}
+     *                inclusions
+     * @return this builder for chaining options together
+     */
+    @NonNull
+    public abstract Builder include(@Nullable @DirectionsCriteria.IncludeCriteria String include);
```
LukasPaczos, line 1238: Can you include only one parameter, or can it be a list? If the latter, we should also expose a `List` method variant and test it.

Guardiola31337: Similar to `exclude` https://github.com/mapbox/mapbox-navigation-android/blob/d1218a77e305e04672754835df5fe22a4a68985a/libnavigation-base/src/main/java/com/mapbox/navigation/base/route/RouteExclusions.kt#L15-L21 should we do implement that for `include` from the SDK side for consistency? 🤔

LukasPaczos: Hmm, this should be part of the `mapbox-java` for both include and exclude, same as all the other list paramaters.

Guardiola31337:
Addressed the feedback. This is ready for another round of eyes @mapbox/navigation-android

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -387,7 +395,14 @@ private RouteOptions routeOptionsList() {
         add(0);
       }})
       .continueStraight(false)
-      .exclude(DirectionsCriteria.EXCLUDE_TOLL)
+      .excludeList(new ArrayList<String>() {{
+          add(DirectionsCriteria.EXCLUDE_TOLL);
+          add(DirectionsCriteria.EXCLUDE_TUNNEL);
```
abhishek1508, line 400: This test uses a `DirectionsCriteria.PROFILE_DRIVING_TRAFFIC` and excludes `TUNNEL`, but `TUNNEL` exclude is not mentioned in the documentation above. If `DRIVING` and `DRIVING_TRAFFIC` both support `TUNNEL` as exclude could you please add it to the documentation

abhishek1508 (commented):
Could we add some more tests on what happens if we exclude or include some criteria for a profile that is not supported? For ex: a test for what happens if `Profile.Walking` and you add `Include.HOV2`

Guardiola31337:
> Could we add some more tests on what happens if we exclude or include some criteria for a profile that is not supported?

I think those are developer misconfigurations that we shouldn't handle from the client side. We threw exceptions in the past but we decided it wasn't a good idea and let Directions API handle it. Actually if tomorrow Directions API wants to support one more profile we'd need to change it from our side and cut a new release. I'd vote for avoiding that.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
Guardiola31337: That is correct. That being said, Directions team is working on supporting the `DRIVING_TRAFFIC` profile in the short term https://github.com/mapbox/mapbox-navigation-android/issues/3844#issuecomment-743361069

In order to be more accurate, I can fix the tests to use `DRIVING` profile instead 👍

Guardiola31337: Addressed this.

Guardiola31337:
Thanks for the feedback @abhishek1508 I've addressed it so this is ready again for another round of 👀

services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
```diff
@@ -24,7 +24,7 @@
    * @param user                    {@link RouteOptions#user()}
    * @param profile                 {@link RouteOptions#profile()}
    * @param coordinates             {@link RouteOptions#coordinatesList()}
-   * @param accessToken             {@link RouteOptions#accessToken()}
+   * @param accessToken             the access token
```
kmadsen, line 27: Why is this changing?

Guardiola31337: Broken Javadocs, I was practicing the [The Boy Scout Rule](https://biratkirat.medium.com/step-8-the-boy-scout-rule-robert-c-martin-uncle-bob-9ac839778385) 🧹

kmadsen: Oh never mind, RouteOptions does not have accessToken https://github.com/mapbox/mapbox-java/pull/1276

Guardiola31337: Yeah, not anymore, some leftovers from https://github.com/mapbox/mapbox-java/pull/1276

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
abhishek1508: Thanks for addressing the feedback, but I meant this:

```
A list of exclude. Exclude certain road types from routing.
     * The default is to not exclude anything from the profile selected.
     * The following exclude flags are available for each profile:
     * <p>
     * {@link DirectionsCriteria#PROFILE_DRIVING}: One of {@link DirectionsCriteria#EXCLUDE_TOLL},
     * {@link DirectionsCriteria#EXCLUDE_MOTORWAY}, or {@link DirectionsCriteria#EXCLUDE_FERRY}.
```
In the test you are using `EXCLUDE_TUNNEL` for `PROFILE_DRIVING`, but the documentation does not mention that it is supported. Could you add `{@link DirectionsCriteria#EXCLUDE_TUNNEL}` to the documenation?

Guardiola31337: Going to fix the tests and use `ferry` for example. `tunnel` is not supported based on https://docs.mapbox.com/api/navigation/directions/

| Profile                | Available excludes                               |
|------------------------|--------------------------------------------------|
| mapbox/driving         | One or more of `toll`, `motorway`, `ferry`, or `unpaved` |
| mapbox/driving-traffic | One or more of `toll`, `motorway`, `ferry`, or `unpaved` |
| mapbox/walking         | No excludes supported                            |
| mapbox/cycling         | `ferry`                                            |

Thanks for confirming @danpat

Guardiola31337:
@kmadsen @abhishek1508 OK ready for a final round of 👀

abhishek1508 (approved):
Thanks for addressing the feedback @Guardiola31337

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -353,6 +353,72 @@ public static Builder builder() {
   @Nullable
   public abstract String exclude();
 
+  /**
+   * A list of exclude.
```
kmadsen, line 357: Don't think this is needed. Redundant with the `@return` and there is already a description
```suggestion
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -353,6 +353,72 @@ public static Builder builder() {
...
+   * <p>
+   * {@link DirectionsCriteria#PROFILE_CYCLING}: No includes supported
+   *
+   * @return a string matching one of the {@link DirectionsCriteria.IncludeCriteria} inclusions
+   */
+  @Nullable
+  public abstract String include();
+
+  /**
+   * A list of include.
```
kmadsen, line 400: Same as above, don't think this is needed.
```suggestion
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1195,6 +1262,84 @@ public Builder annotationsList(@Nullable List<String> annotations) {
     @NonNull
     public abstract Builder exclude(@Nullable @DirectionsCriteria.ExcludeCriteria String exclude);
 
+    /**
+     * A list of exclude. Exclude certain road types from routing.
```
kmadsen, line 1266:
```suggestion
     * Exclude certain road types from routing.
```

PR #1310, opened 2021-10-21
RouteOptions and Directions API: add `avoidManeuverRadius` ("avoid_maneuver_radius") param

RingerJK:
RouteOptions and Directions API: add `avoidManeuverRadius` ("avoid_maneuver_radius") param


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -199,6 +200,25 @@ public static Builder builder() {
...
+   * speed to avoid dangerous maneuvers when re-routing. If a route is not found using
+   * the specified value, it will be ignored. Note that if a large radius is used, the routers'
+   * provider may ignore an important turn and return a long straight path before the first
+   * maneuver.
+   * <p>
+   * Note: the param cannot be used with {@link #departAt()} and {@link #arriveBy()},
+   * since these modes are used for reference requests, not for real-time routing.
+   *
+   * @return a radius around a starting point where router's provider tries to avoid any
+   *   significant maneuvers.
```
LukasPaczos, line 216: Let's mention the unit here as well (meters).

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1160,6 +1183,25 @@ public Builder bearingsList(@Nullable List<Bearing> bearings) {
...
+     * <p>
+     * Note: the param cannot be used with {@link #departAt(String)} and {@link #arriveBy(String)},
+     * since these modes are used for reference requests, not for real-time routing.
+     *
+     * @param avoidManeuverRadius avoid maneuver radius, in meters. Possible range is [0, 1000].
+     * @return this builder for chaining options together
+     */
+    @NonNull
+    public abstract Builder avoidManeuverRadius(
+      @Nullable @IntRange(from = 0, to = 1000) Integer avoidManeuverRadius
```
LukasPaczos, line 1202: Let's not specify a range here and let server validate it instead. This allows us more flexibility in case the range is expanded.

LukasPaczos: If we provide more than `1000`, is an error returned or is it clamped to `1000`?

RingerJK: server returns error 
```
{
  "message": "should be <= 1000",
  "code": "InvalidInput"
}
```

RingerJK: we should mention 1000 meters in the docs anyway, so it will be limited anyway

LukasPaczos: Okay, that's fine and matches all other APIs, let's not restrict the range in `mapbox-java` and let server validate.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1,6 +1,7 @@
 package com.mapbox.api.directions.v5.models;
 
 import androidx.annotation.FloatRange;
+import androidx.annotation.IntRange;
```
Zayankovsky, line 4: Is this still needed?

RingerJK: removed 👍

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -201,6 +201,25 @@ public static Builder builder() {
...
+   * speed to avoid dangerous maneuvers when re-routing. If a route is not found using
+   * the specified value, it will be ignored. Note that if a large radius is used, the routers'
+   * provider may ignore an important turn and return a long straight path before the first
+   * maneuver.
+   * <p>
+   * Note: the param cannot be used with {@link #departAt()} and {@link #arriveBy()},
+   * since these modes are used for reference requests, not for real-time routing.
+   *
+   * @return a radius around a starting point where router's provider tries to avoid any
+   *   significant maneuvers. Possible range is [0, 1000].
```
LukasPaczos, line 217:
```suggestion
   * @return a radius around a starting point where router's provider tries to avoid any
   *   significant maneuvers. Possible range is [0, 1000] in meters.
```

LukasPaczos (approved):

PR #1320, opened 2021-11-11
Added code owners for publisher-production

VysotskiVadim:
@colleenmcginnis and @danswick will be assigned to every PR in publisher-production automatically

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,1 @@
+*       @mapbox/docs @colleenmcginnis


.github/CODEOWNERS
```diff
@@ -0,0 +1 @@
+*       @colleenmcginnis @danswick
```
RingerJK, line 1: should be @mapbox/docs I think 🤔 cc @colleenmcginnis

Issue #1309, opened 2021-10-20
Add DirectionsCriteria to exclude toll roads that only accept cash payment


1ec5:
The Directions API now supports an `exclude=cash_only_tolls` parameter, regardless of routing profile. We should add an `EXCLUDE_CASH_ONLY_TOLLS` constant to the `DirectionsCriteria` class.

/cc @mapbox/navigation-android @ktatterso

LukasPaczos:
Land in https://github.com/mapbox/mapbox-java/pull/1327.

VysotskiVadim:
Tested on the mapbox-java `6.2.0-SNAPSHOT`. Directions API doesn't fail when I use `DirectionsCriteria.EXCLUDE_CASH_ONLY_TOLLS` from an example Android app.

Issue #1308, opened 2021-10-20
Add DirectionsCriteria to exclude unpaved roads


1ec5:
The Directions API’s mapbox/driving and mapbox/driving-traffic profiles now support an `exclude=unpaved` parameter. We should add an `EXCLUDE_UNPAVED` constant to the `DirectionsCriteria` class.

/cc @mapbox/navigation-android @ktatterso

LukasPaczos:
Land in https://github.com/mapbox/mapbox-java/pull/1327.

VysotskiVadim:
Tested on the mapbox-java `6.2.0-SNAPSHOT`. Directions API doesn't fail when I use `DirectionsCriteria.EXCLUDE_UNPAVED` from an example Android app.

PR #1327, opened 2021-11-16
Added exclude unpaved and cash only tolls options

VysotskiVadim:
Implementation of the #1309 and the #1308

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
@@ -179,6 +179,16 @@ public final class DirectionsCriteria {
   public static final String EXCLUDE_RESTRICTED = "restricted";
 
   /**
+   * Exclude all cash only tolls along the returned directions route.
+   */
+  public static final String EXCLUDE_CASH_ONLY_TOLLS = "cash_only_tolls";
+
+  /**
+   * Exclude all unpaved roads along the returned directions route.
+   */
+  public static final String EXCLUDE_UNPAVED = "unpaved";
+
+  /**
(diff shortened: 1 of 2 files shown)


VysotskiVadim:
I believe this kind of changes doesn't require auto tests, do they?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
```diff
@@ -178,6 +178,16 @@
    */
   public static final String EXCLUDE_RESTRICTED = "restricted";
 
+  /**
+   * Exclude all cash only tolls along the returned directions route.
+   */
+  public static final String EXCLUDE_CASH_ONLY_TOLLS = "cash_only_tolls";
+
+  /**
+   * Exclude all unpaved roads along the returned directions route.
```
LukasPaczos, line 187:
```suggestion
   * Exclude all unpaved roads along the returned directions route.
   * Only available for {@link DirectionsCriteria#PROFILE_DRIVING} and {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC}
```

LukasPaczos (approved):

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
VysotskiVadim: It's good idea! But none of other comments mention limitation.
image
I would update all or none 🙂
@LukasPaczos , what do you think about it?

LukasPaczos: Since we don't have an architectural way to enforce this, let's skip the doc in this case and let the server respond with an error.

Issue #1316, opened 2021-11-08
Setup services for styles API


abhishek1508:
As an effort to introduce the new Mapbox designed route shields, Navigation SDK needs to make following 2 API calls to `styles` endpoint in order to retrieve the blank route shield and placeholder values for positioning the text on top of the shield. 

##### Request a shield:
```
https://api.mapbox.com/styles/v1/:ownerId/:styleId/draft?/sprite/:imageId
```

##### Request text placeholders:
```
https://api.mapbox.com/styles/v1/:ownerId/:styleId/:revision?/sprite(:pixelRatio?).json
```

In order to add these services to this repository there are 2 options:
- Add a new module `services-shields / services-styles`
- Use an existing module to host these new services; probably `services-directions`

I am leaning more towards the 1st option as it clearly states the purpose of this new module and services which kind of doesn't make sense to add it to the `services-directions` or any other module.

I would like to know your thoughts or any other suggestions

cc @mapbox/navigation-android

zugaldia:
I agree that a new `services-styles` makes more sense here.

abhishek1508:
More details on the use case are as follows:

```
https://api.mapbox.com/styles/v1/:ownerId/:styleId/:revision?/sprite(:pixelRatio?).json
```
The request above needs to be made every time there is a change in the map style[eg: style change from light to dark mode, street to satellite mode]. This particular request returns a set of all the shields available and would be favorable to not make this request every time there is a style change. Rather, it would be ideal to cache the request/response and only hit the network when the data becomes stale.

```
https://api.mapbox.com/styles/v1/:ownerId/:styleId/draft?/sprite/:imageId
```
The request above results in a black shield image as shown below
![Screen Shot 2021-08-25 at 10 24 59 AM](https://user-images.githubusercontent.com/1590802/130837200-9ae83f94-de4f-41f4-b3e8-84c6a6238d9a.png) ![Screen Shot 2021-08-25 at 10 27 08 AM](https://user-images.githubusercontent.com/1590802/130837278-f872b43d-2724-4474-877a-a4aae9c25368.png)

It would be ideal to cache these blank shields as they are unlikely to change in real world scenarios and the shields can be cached for a longer period of time ranging from months to an year.

Both these endpoints can be queried for using:
- MapboxJava services
- Common SDK httpclient

[...]

PR #1336, opened 2021-11-24
Add styles API models

Guardiola31337:
Fixes https://github.com/mapbox/mapbox-java/issues/1316

Opening this as a `Draft` as there are still some open questions that need to be addressed downstream in the BE:

- As a response we're getting

```json
{
	"turning-circle-outline": {
		"width": 138,
		"height": 138,
		"x": 0,
		"y": 0,
		"pixelRatio": 1,
		"visible": true
	},
	"turning-circle": {
		"width": 126,
		"height": 126,
		"x": 138,
		"y": 0,
		"pixelRatio": 1,
		"visible": true
	},
	"us-interstate-truck-2": {
		"width": 60,
		"height": 120,
		"x": 0,
		"y": 138,
		"pixelRatio": 1,
		"placeholder": [0, 17, 20, 23],
		"visible": true
	}
}
```

Can we get instead 

```json
{
  "sprites": [{
    "sprite_name": "turning-circle-outline",
    "sprite_attributes": {
      "width": 138,
      "height": 138,
      "x": 0,
      "y": 0,
      "pixel_ratio": 1,
      "visible": true
    }
  }, {
    "sprite_name": "turning-circle",
    "sprite_attributes": {
      "width": 126,
      "height": 126,
      "x": 138,
      "y": 0,
      "pixel_ratio": 1,
      "visible": true
    }
  }, {
    "sprite_name": "us-interstate-truck-2",
    "sprite_attributes": {
      "width": 60,
      "height": 120,
      "x": 0,
      "y": 138,
      "pixel_ratio": 1,
      "placeholder": [0.0, 17.0, 20.0, 23.0],
      "visible": true
    }
  }]
}
```

- Also noticed that the `placeholder` is not an integer, could we return floating point values instead? I.e. instead of `"placeholder": [0, 17, 20, 23]` get `"placeholder": [0.0, 17.0, 20.0, 23.0]`.

@mandeepsandhu @dereklieu

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/ShieldSprites.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/ShieldSprites.java
@@ -1,0 +1,128 @@
+package com.mapbox.api.directions.v5.models;
+
+import androidx.annotation.NonNull;
+
+import com.google.auto.value.AutoValue;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.TypeAdapter;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * ShieldSprites.
+ */
+@AutoValue
+public abstract class ShieldSprites implements Serializable {
+
+  /**
+   * Create a new instance of this class by using the {@link ShieldSprites.Builder} class.
+   *
+   * @return {@link ShieldSprites.Builder} for creating a new instance
+   */
+  public static Builder builder() {
+    return new AutoValue_ShieldSprites.Builder();
+  }
+
+  /**
+   * Create a new instance of this class by passing in a formatted valid JSON String.
+   *
+   * @param json a formatted valid JSON string defining a shield sprite
+   * @return a new instance of this class defined by the values passed inside this static factory
+   *   method
+   */
+  @NonNull
+  public static ShieldSprites fromJson(@NonNull String json) {
+    List<ShieldSprite> sprites = new ArrayList<>();
+    GsonBuilder gson = new GsonBuilder();
+    JsonObject jsonObject = gson.create().fromJson(json, JsonObject.class);
+    Set<Map.Entry<String, JsonElement>> entries = jsonObject.entrySet();
+    for (Map.Entry<String, JsonElement> entry : entries) {
+      String spriteName = entry.getKey();
+      ShieldSpriteAttribute spriteAttribute = ShieldSpriteAttribute.fromJson(
+          jsonObject.get(spriteName).toString()
+      );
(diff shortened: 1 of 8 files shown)


Guardiola31337:
@mandeepsandhu @dereklieu do we have any documentation available to add it as part of Javadocs?

dereklieu:
Hey @Guardiola31337, public documentation for the sprite endpoint lives here: https://docs.mapbox.com/api/maps/styles/#sprites

The sprite JSON metadata endpoint has numerous consumers, including gl-js, Studio, anything that renders a map, and we cannot just go changing the format of the response. I recommend parsing and formatting the output to resemble what you're asking for on the client instead.

> Also noticed that the placeholder is not an integer, could we return floating point values instead?

The `placeholder` key is an array of numbers, which in java will look like integers when they are whole, and floats when they are not. Most `JSON.stringify` implementations will remove trailing 0's, so doing what you're asking would not be possible, unless we use a string type. This has its own problems which I would rather avoid.

Screen Shot 2021-11-29 at 10 38 16 AM

Guardiola31337:
> The sprite JSON metadata endpoint has numerous consumers, including gl-js, Studio, anything that renders a map, and we cannot just go changing the format of the response. I recommend parsing and formatting the output to resemble what you're asking for on the client instead.

Got it. Don't like it but as it may impact the overall performance but understandable.

> The `placeholder` key is an array of numbers, which in java will look like integers when they are whole, and floats when they are not. Most `JSON.stringify` implementations will remove trailing 0's, so doing what you're asking would not be possible, unless we use a string type. This has its own problems which I would rather avoid.

As always we're suffering from JS and JSON of not being type safe 😬 

I've re-worked the implementation and it's working without the changes requested. This is ready for review @abhishek1508

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/ShieldSprites.java
```diff
@@ -0,0 +1,128 @@
...
+   *
+   * @param json a formatted valid JSON string defining a shield sprite
+   * @return a new instance of this class defined by the values passed inside this static factory
+   *   method
+   */
+  @NonNull
+  public static ShieldSprites fromJson(@NonNull String json) {
+    List<ShieldSprite> sprites = new ArrayList<>();
+    GsonBuilder gson = new GsonBuilder();
+    JsonObject jsonObject = gson.create().fromJson(json, JsonObject.class);
```
abhishek1508, line 45: Since `gson.create().fromJson` throws a JSONSyntaxException in case of invalid representation for an object, should we wrap this entire piece inside a `try/catch`?
cc @Guardiola31337

Guardiola31337: If that's the case the exception will be thrown. We don't need to `try/catch` unless we want to fail silently or do something in addition before throwing the exception.

abhishek1508: Okay I think it makes sense to throw exception and let the SDK/app user handle the exception upstream.

abhishek1508 (approved):

PR #1342, opened 2021-12-06
Promote `Search SDK` as a replacement of `Geocoder`

friendoye:
[Search SDK for Android](https://docs.mapbox.com/android/search/guides/) now is the recommended way to access Geocoding API on Android platform.  
To promote Search SDK usage this PR:
1) Adds `Geocoder` deprecation note;
2) Assigns `services-geocoding` ownership to @mapbox/search-sdk-android  team

### DIFF
--- a/CODEOWNERS.md
+++ b/CODEOWNERS.md
@@ -1,0 +1,2 @@
+# Transfer ownership over Geocoder related files to Search SDK Android team
+services-geocoding/ @mapbox/search-sdk-android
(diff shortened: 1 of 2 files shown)


friendoye:
@LukasPaczos I wouldn't mind adding deprecation notes and annotation to `Geocoder`'s classes, but I'm not fully convinced it's necessary:
1) `Search SDK` is available only for Android platform, while `Java SDK`'s artifacts are available for JVM as well. Customers using `Geocoder` on JVM will not be able to migrate to `Search SDK`, making deprecation notes a bit misleading;
2) `Search SDK` doesn't have plans to support batch geocoding right now, but `Geocoder` provides such functionality ([migration guide](https://docs.mapbox.com/android/search/guides/geocoder-migration/) emphasises this fact). Thus we can only deprecate a few methods of `MapboxGeocoding` class.

LukasPaczos:
Thanks, that makes total sense :+1:

PR #1348, opened 2022-01-14
Primitive implementation of line intersect

OttyLab:
## Summary
Implement line intersect which is equivalent to [turf.lineIntersect](https://turfjs.org/docs/#lineIntersect) with a primitive manner.

## Discussion

- [ ] It's O(nm) implementation as same as [Turf-Swift](https://github.com/mapbox/turf-swift/blob/v2.1.0/Sources/Turf/Geometries/LineString.swift#L384). Please judge if this algorithm is acceptable or not as a stopgap solution.
- [ ] The handling of line segment endpoint is different from Turf-Swift. In Turf-Swift endpoint is included while not in Turf-Java. [This](https://github.com/mapbox/mapbox-java/blob/v6.1.0/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java#L349) and [this](https://github.com/mapbox/mapbox-java/blob/v6.1.0/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java#L353) `if` conditions causes the difference. If they are like `varA >= 0 && varA <= 1`, the behavior becomes same. Which Swift/Java is the expected? (This is because the test is slightly different from Turf-Swift)

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
@@ -27,6 +27,44 @@ public final class TurfMisc {
   }
 
   /**
+   * Takes lines {@link LineString} and returns intersect points. The time complexity is O(nm)
+   * that is not as efficient as Turf.js implementation.
+   *
+   * @param line1 LineString 1
+   * @param line2 LineString 2
+   * @return Intersect points
+   * @see <a href="https://turfjs.org/docs/#lineIntersect">Turf Line intersect documentation</a>
+   * @since 6.2.0
+   */
+  @NonNull
+  public static List<Point> lineIntersect(@NonNull LineString line1, @NonNull LineString line2) {
+    List<Point> result = new ArrayList<>();
+    Point[] line1Points = line1.coordinates().toArray(new Point[line1.coordinates().size()]);
+    Point[] line2Points = line2.coordinates().toArray(new Point[line2.coordinates().size()]);
+
+    for (int i = 0; i < line1Points.length - 1; i++) {
+      for (int j = 0; j < line2Points.length - 1; j++) {
+        LineIntersectsResult intersects = lineIntersects(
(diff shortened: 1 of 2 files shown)


services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
```diff
@@ -26,6 +26,43 @@ private TurfMisc() {
     throw new AssertionError("No Instances.");
   }
 
+  /**
+   * Takes lines {@link LineString} and returns intersect points.
```
LukasPaczos, line 30: It'd be good to mention time complexity and that it's not as efficient as the JS counterpart.

LukasPaczos (commented):
LGTM with one small change request

> The handling of line segment endpoint is different from Turf-Swift.

Looks like [TurfJS is inclusive](https://github.com/Turfjs/turf/blob/6bc5fbe927b85b2be8e38600b5ffb29525a6f925/packages/turf-line-intersect/index.ts#L124) as well so we might want to fix this up on the Java side.

Would you be able to cut a separate ticket/PR?

services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
OttyLab: Thanks, I updated the comment

OttyLab:
@LukasPaczos 

I opened a [PR](https://github.com/mapbox/mapbox-java/pull/1350) which fixes `lineIntersects`. I'll update current PR after `lineIntersects` fix is merged because I want to change the test code.

OttyLab:
@LukasPaczos @pengdev 

By following [this comment](https://github.com/mapbox/mapbox-java/pull/1350#issuecomment-1021173663), I've picked up the fix of `lineIntersects` in this PR and added the edge test.

pengdev (approved):
LGTM

LukasPaczos (approved):
LGTM :ship:

PR #1350, opened 2022-01-18
Fix lineIntersects to align with Turf.js implementation

OttyLab:
## Summary
`lineIntersects` excludes line segment endpoints which is not consistent with Turf.js / Turf-Swift.

## Detail
[This](https://github.com/mapbox/mapbox-java/blob/v6.1.0/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java#L349) and [this](https://github.com/mapbox/mapbox-java/blob/v6.1.0/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java#L353) `if` conditions causes the difference. Turf.js implementation is [here](https://github.com/Turfjs/turf/blob/6bc5fbe927b85b2be8e38600b5ffb29525a6f925/packages/turf-line-intersect/index.ts#L124).

[Reference](https://github.com/mapbox/mapbox-java/pull/1348#pullrequestreview-854414293)

### DIFF
--- a/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
+++ b/services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
@@ -346,11 +346,11 @@ public final class TurfMisc {
       + (varA * (line1EndY - line1StartY))).build();
 
     // if line1 is a segment and line2 is infinite, they intersect if:
-    if (varA > 0 && varA < 1) {
+    if (varA >= 0 && varA <= 1) {
       result = result.toBuilder().onLine1(true).build();
     }
     // if line2 is a segment and line1 is infinite, they intersect if:
-    if (varB > 0 && varB < 1) {
+    if (varB >= 0 && varB <= 1) {
       result = result.toBuilder().onLine2(true).build();
     }
     // if line1 and line2 are segments, they intersect if both of the above are true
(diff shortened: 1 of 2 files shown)


services-turf/src/main/java/com/mapbox/turf/TurfMisc.java
```diff
@@ -346,11 +346,11 @@ private static LineIntersectsResult lineIntersects(double line1StartX, double li
       + (varA * (line1EndY - line1StartY))).build();
 
     // if line1 is a segment and line2 is infinite, they intersect if:
-    if (varA > 0 && varA < 1) {
+    if (varA >= 0 && varA <= 1) {
```
pengdev, line 349: could we add tests for the changes?

OttyLab: @pengdev Thanks, I added the test.

LukasPaczos:
I'd be in favor of merging this PR with https://github.com/mapbox/mapbox-java/pull/1348 and using the public API to test the boundaries, instead of using reflection.

services-turf/src/test/java/com/mapbox/turf/TurfMiscTest.java
```diff
@@ -490,4 +493,27 @@ public void testShortLine() throws IOException, TurfException {
     assertEquals(sliced.coordinates().get(sliced.coordinates().size() - 1).coordinates(),
       end_point.coordinates());
   }
+
+  @Test
+  public void testLineIntersects() throws
+          NoSuchMethodException, SecurityException,
+          IllegalAccessException, IllegalArgumentException, InvocationTargetException {
+    Method method = TurfMisc.class.getDeclaredMethod("lineIntersects",
```
pengdev, line 501: I'd like to suggest testing against the public API instead of using reflections.

OttyLab:
I close this PR because the fix is merged into #1348 as I mentioned [here](https://github.com/mapbox/mapbox-java/pull/1348#issuecomment-1022731055).

PR #1359, opened 2022-02-04
downgrade docker image to use Java 8

LukasPaczos:
Latest docker image swapped Java version used from 8 to 11 which broke this project's compilation. This PR downgrades the image version to the latest based on Java 8.

We can iterate from here and migrate `mapbox-java` to be compatible with Java 11 but proposing this quick fix to unblock snapshot releases needed for downstream project.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,13 @@
 version: 2.1
+
+executors:
+  java-executor:
+    docker:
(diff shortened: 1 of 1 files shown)


VysotskiVadim (approved):
It would be nice to have a defined executor in the config:
```
executors:
  default-executor:
    docker:
      - image: mbgl/android-ndk-r22:f547210689
```
so that all changes could be made in one place next time

PR #1360, opened 2022-02-07
Updated DirectionsWaypoint object

LukasPaczos:
Updated nullability annotations and exposed optional `distance` property. Refs https://docs.mapbox.com/api/navigation/directions/#waypoint-object.

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsWaypoint.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsWaypoint.java
@@ -1,6 +1,6 @@
 package com.mapbox.api.directions.v5.models;
 
-import androidx.annotation.Nullable;
+import androidx.annotation.NonNull;
 import com.google.auto.value.AutoValue;
 import com.google.gson.Gson;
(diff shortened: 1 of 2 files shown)


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsWaypoint.java
```diff
@@ -33,31 +34,32 @@ public static Builder builder() {
    * @return string with the name of the way the coordinate snapped to
    * @since 1.0.0
    */
-  @Nullable
+  @NonNull
```
Guardiola31337, line 37: This is technically SemVer although we could treat it as a bug.

Guardiola31337 (approved):

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsWaypoint.java
LukasPaczos: That's my thinking, not a semver for the getter only for the setter, but more of a bug nonetheless.

PR #1361, opened 2022-02-10
Need feedback about exclude waypoint parameter

VysotskiVadim:
## Description

Direction API provides a new feature "exclude waypoint". To exclude a coordinate form your route add it to the list of `exclude` like `point(lon lat)`. For example if you want to exclude motorways and coordinate `1.0 1.0` use following exclude param: `exclude=motorway,point(1.0%201.0)`

## Please provide a feedback

You will find the most simple and straightforward implementation which I can came up with. If you don't like how it looks I can try implement it in OOP style, but I'm not sure if it's worth.

### DIFF
--- a/services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
+++ b/services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
@@ -1,14 +1,17 @@
 package com.mapbox.api.directions.v5.models;
 
 import static com.google.gson.JsonParser.parseString;
+import static org.junit.Assert.*;
+
+import java.net.MalformedURLException;
 import java.net.URL;
-import static org.junit.Assert.assertEquals;
-import static org.junit.Assert.assertNull;
+
 import com.mapbox.api.directions.v5.DirectionsCriteria;
 import com.mapbox.core.TestUtils;
 import com.mapbox.geojson.Point;
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import org.junit.Test;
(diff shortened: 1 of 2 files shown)


services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -437,6 +440,23 @@ public void routeOptionsWithUTF8Chars_toUrlWithEncodedChars() {
     assertEquals(expectedEncodedUrl, url.toString());
   }
 
+  @Test
+  public void routeOptionWithExcludedWaypoint() {
+    RouteOptions options = RouteOptions.builder()
+        .coordinates("1.0,1.0;2.0,2.0")
+        .profile(DirectionsCriteria.PROFILE_DRIVING)
+        .excludeList(Collections.singletonList(
```
RingerJK, line 448: let's add separate property to RouteOptions, like `excludePointsList(List<Point> excludePointsList)` and process exclude profiles and coordinates separately

RingerJK: that true, that each RouteOptions field corresponded to request parameter, but previously parameters weren't mixed in the same place. 

Where it's actually mixed is URL, which we handle under the hood. So it's not a problem for customers I think

VysotskiVadim:
Another option I consider is having an OOP API for exclude together with an existing API:
```
interface ExcludeParameter {
    String toHttpParameter();
}

final class StaticExclude {
    internal StaticExclude(String rawValue) { ... }
    public String toHttpParameter() { 
         return this.rawValue;
    }
}

final class PointExclude {
    internal PointExclude(Point point) { ... }
     public String toHttpParameter() { 
         return "point(" + this.point.longitude() + " " + this.point.latitude() + ")";    
     }
}


public final class DirectionsCriteria {
    public static final ExcludeParameter EXLUDE_MOTORWAY_OBJECT  = StaticExclude(EXCLUDE_MOTORWAY)
    pubic static final ExcludeParameter exludePoint(Point point) { return PointExclude(point); }
}

public class RoutOptions {

    public class Builder {
        public void excludeObjectsList(List<ExcludeCriteria> excludeObjectsList) {...}
    }

    public List<ExcludeCriteria> excludeObjectsList() {...}
}
```

As a disadvantage I see that deserialisation from a string `exclude` field to list of objects won't be simple and I'm not sure that OOP API worth the complexity it brings.

VysotskiVadim:
I think the main question is: do our customers want to get a `Point` object from excludes in the `RouteObject`'s API. 
**If yes** then @RingerJK's idea of [separate processing](https://github.com/mapbox/mapbox-java/pull/1361#discussion_r803779788) or my [OOP idea](https://github.com/mapbox/mapbox-java/pull/1361#issuecomment-1035119093) is the way to go.
**If no** then current simple solution is sufficient. 

If there are no strong opinions I would prefer to go with simple solution. If we realise that somebody wants to access a `Point` object from excludes later then we can add more complex solution later 🙂 
@mapbox/navigation-android , what do you think?

VysotskiVadim:
> we cannot re-assing exclude fields, it will be breaking changes.

@RingerJK , why do you think it's a breaking change? Do you think that developers don't expect that we will add new flags and rely heavily on the ones we listed in the documentation? What can we break by providing points among excludes as Directions API have already done in their API?

VysotskiVadim:
> my [OOP idea](https://github.com/mapbox/mapbox-java/pull/1361#issuecomment-1035119093) is the way to go.

the OOP idea doesn't provide a type safe idimic java API for getting points back from the RouteOptions. So I don't see any benefits over the approach from PR

VysotskiVadim:
Closing PR in favour of the second iteration #1362

PR #1362, opened 2022-02-15
Exclude waypoints from routing 

VysotskiVadim:
## Description

Direction API provides a new feature "exclude waypoint". To exclude a coordinate form your route add it to the list of `exclude` like `point(lon lat)`. For example if you want to exclude motorways and coordinate `1.0 1.0` use following exclude param: `exclude=motorway,point(1.0%201.0)`

You're reviewing a second iteration of the exclude waypoints implementation. First iteration is #1361.

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
@@ -133,9 +133,9 @@ public abstract class Exclude {
     }
 
     return Exclude.builder()
-        .criteria(criteria)
-        .points(points)
-        .build();
+      .criteria(criteria)
+      .points(points)
+      .build();
   }
 
   @Nullable
@@ -185,14 +185,14 @@ public abstract class Exclude {
       String latitude = point.substring(delimiter + 1, point.length() - 1);
       try {
         Point result = Point.fromLngLat(
-            Double.parseDouble(longitude),
-            Double.parseDouble(latitude)
+          Double.parseDouble(longitude),
+          Double.parseDouble(latitude)
         );
         return result;
       } catch (NumberFormatException numberFormatError) {
         throw new IllegalArgumentException(
-            pointParsingErrorMessage(point),
-            numberFormatError
+          pointParsingErrorMessage(point),
+          numberFormatError
         );
       }
     } else {
@@ -202,9 +202,9 @@ public abstract class Exclude {
 
   private void appendPoint(StringBuilder result, Point point) {
     result.append("point(")
-        .append(point.longitude())
-        .append(' ')
-        .append(point.latitude())
-        .append(')');
+      .append(point.longitude())
+      .append(' ')
+      .append(point.latitude())
+      .append(')');
   }
 }
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
@@ -1,0 +1,210 @@
+package com.mapbox.api.directions.v5.models;
+
+import androidx.annotation.NonNull;
+import androidx.annotation.Nullable;
+import com.google.auto.value.AutoValue;
+import com.mapbox.api.directions.v5.DirectionsCriteria;
+import com.mapbox.api.directions.v5.utils.ParseUtils;
+import com.mapbox.geojson.Point;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+
+/**
+ * Defines excludes for {@link RouteOptions.Builder#excludeObject(Exclude)}.
+ */
+@AutoValue
+public abstract class Exclude {
+
+  /**
+   * Build a new instance of {@link Exclude}. Nothing is excluded by default.
+   */
+  @NonNull
+  public static Builder builder() {
+    return new AutoValue_Exclude.Builder();
+  }
+
+  /**
+   * Exclude certain road types from routing. The default is to not exclude anything from the
+   * profile selected. The following exclude flags are available for each profile:
+   * <p>
+   * {@link DirectionsCriteria#PROFILE_DRIVING}: One of {@link DirectionsCriteria#EXCLUDE_TOLL},
+   * {@link DirectionsCriteria#EXCLUDE_MOTORWAY}, or {@link DirectionsCriteria#EXCLUDE_FERRY}.
+   * <p>
+   * {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC}: One of
+   * {@link DirectionsCriteria#EXCLUDE_TOLL}, {@link DirectionsCriteria#EXCLUDE_MOTORWAY}, or
+   * {@link DirectionsCriteria#EXCLUDE_FERRY}.
+   * <p>
+   * {@link DirectionsCriteria#PROFILE_WALKING}: No excludes supported
+   * <p>
+   * {@link DirectionsCriteria#PROFILE_CYCLING}: {@link DirectionsCriteria#EXCLUDE_FERRY}
+   *
+   * @return a list of strings matching one
+   *   of the {@link DirectionsCriteria.ExcludeCriteria} exclusions
+   */
+  @Nullable
+  public abstract List<String> criteria();
+
+  /**
+   * Exclude certain points from routing. The default is to not exclude anything.
+   *
+   * @return a list of points excluded from the routing.
+   */
+  @Nullable
+  public abstract List<Point> points();
+
+  /**
+   * Use this builder to build an {@link Exclude} object.
+   */
+  @AutoValue.Builder
+  public abstract static class Builder {
+
+    /**
+     * Exclude certain road types from routing. The default is to not exclude anything from the
+     * profile selected. The following exclude flags are available for each profile:
+     * <p>
+     * {@link DirectionsCriteria#PROFILE_DRIVING}: One of {@link DirectionsCriteria#EXCLUDE_TOLL},
+     * {@link DirectionsCriteria#EXCLUDE_MOTORWAY}, or {@link DirectionsCriteria#EXCLUDE_FERRY}.
+     * <p>
+     * {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC}: One of
+     * {@link DirectionsCriteria#EXCLUDE_TOLL}, {@link DirectionsCriteria#EXCLUDE_MOTORWAY}, or
+     * {@link DirectionsCriteria#EXCLUDE_FERRY}.
+     * <p>
+     * {@link DirectionsCriteria#PROFILE_WALKING}: No excludes supported
+     * <p>
+     * {@link DirectionsCriteria#PROFILE_CYCLING}: {@link DirectionsCriteria#EXCLUDE_FERRY}
+     *
+     * @param criteria a list of exclude criteria
+     *   matching {@link DirectionsCriteria.ExcludeCriteria}
+     * @return this builder.
+     */
+    public abstract Builder criteria(@Nullable List<String> criteria);
+
+    /**
+     * Exclude certain points from routing. The default is to not exclude anything.
+     *
+     * @param points a list of points to exclude.
+     * @return this builder.
+     */
+    public abstract Builder points(@Nullable List<Point> points);
+
+    /**
+     * Builds the object.
+     *
+     * @return a new instance of {@link Exclude}
+     */
+    public abstract Exclude build();
+  }
+
+  @Nullable
+  static Exclude fromUrlQueryParameter(@Nullable String queryParameter) {
+    if (queryParameter == null || Objects.equals(queryParameter, "")) {
+      return null;
+    }
+
+    ArrayList<Point> points = new ArrayList<>();
+    ArrayList<String> criteria = new ArrayList<>();
+
+    List<String> excludeItems = ParseUtils.parseToStrings(queryParameter, ",");
+    if (excludeItems != null) {
+      for (String excludeItem : excludeItems) {
+        if (excludeItem.startsWith("point(") && excludeItem.endsWith(")")) {
+          Point point = parsePoint(excludeItem);
+          points.add(point);
+        } else if (excludeItem.contains("(") && excludeItem.contains(")")) {
+          // Ignoring the unexpected type of data. Update the library to get support of new types
+        } else if (excludeItem.contains("(") || excludeItem.contains(")")) {
+          throw new IllegalArgumentException("Can't parse parameter " + excludeItem);
+        } else {
+          criteria.add(excludeItem);
+        }
+      }
+    }
+
+    if (criteria.size() == 0) {
+      criteria = null;
+    }
+    if (points.size() == 0) {
+      points = null;
+    }
+    if (criteria == null && points == null) {
+      return null;
+    }
+
+    return Exclude.builder()
+        .criteria(criteria)
+        .points(points)
+        .build();
+  }
+
+  @Nullable
+  String toUrlQueryParameter() {
+    if (points() == null && criteria() == null) {
+      return null;
+    }
+
+    StringBuilder result = new StringBuilder();
+
+    appendPoints(result);
+    appendCriterias(result);
+
+    return result.toString();
+  }
+
+  private void appendCriterias(StringBuilder result) {
+    if (criteria() != null) {
+      for (String item : criteria()) {
+        if (result.length() != 0) {
+          result.append(',');
+        }
+        result.append(item);
+      }
+    }
+  }
+
+  private void appendPoints(StringBuilder result) {
+    if (points() != null) {
+      for (Point point : points()) {
+        if (result.length() != 0) {
+          result.append(",");
+        }
+        appendPoint(result, point);
+      }
+    }
+  }
+
+  private static String pointParsingErrorMessage(String point) {
+    return "Can't parse a point:" + point + ". Expected format: point(lon lat)";
+  }
+
+  private static Point parsePoint(String point) {
+    int delimiter = point.indexOf(' ');
+    if (delimiter != -1) {
+      String longitude = point.substring(6, delimiter);
+      String latitude = point.substring(delimiter + 1, point.length() - 1);
+      try {
+        Point result = Point.fromLngLat(
+            Double.parseDouble(longitude),
+            Double.parseDouble(latitude)
+        );
+        return result;
+      } catch (NumberFormatException numberFormatError) {
+        throw new IllegalArgumentException(
+            pointParsingErrorMessage(point),
+            numberFormatError
+        );
+      }
+    } else {
+      throw new IllegalArgumentException(pointParsingErrorMessage(point));
+    }
+  }
+
+  private void appendPoint(StringBuilder result, Point point) {
+    result.append("point(")
+        .append(point.longitude())
+        .append(' ')
+        .append(point.latitude())
+        .append(')');
+  }
+}
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
@@ -8,11 +8,19 @@ import com.mapbox.api.directions.v5.utils.ParseUtils;
 import com.mapbox.geojson.Point;
 
 import java.util.ArrayList;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Objects;
+import java.util.Set;
 
 /**
  * Defines excludes for {@link RouteOptions.Builder#excludeObject(Exclude)}.
+ *
+ * This class provides type-safe way to read and build {@link RouteOptions#exclude()} parameter.
+ *
+ * All properties are strictly categorized after parsing. Unknown data types or flags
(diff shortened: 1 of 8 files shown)


services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -437,6 +444,107 @@ public void routeOptionsWithUTF8Chars_toUrlWithEncodedChars() {
...
+    List<String> queryParameters = Arrays.asList(url.getQuery().split("&"));
+    assertTrue(
+        "url doesn't contain excluded point: " + url.toString(),
+        queryParameters.contains("exclude=point(1.0%202.0),point(6.03%208.07)")
+    );
+  }
+
+  @Test
+  public void routeOptionsWithExcludePointToFromUrl() throws MalformedURLException {
+    URL testUrl = new URL("https://api.mapbox.com/directions/v5/mapbox/driving-traffic/18.60289583391497,54.41121871390118;18.59400217318438,54.40983705017376.json?access_token=testToken&alternatives=true&annotations=distance%2Cduration%2Ccongestion%2Cspeed&geometries=geojson&language=en&overview=full&steps=true&exclude=point(18.595875353791087%2054.41000119108463)");
```
LukasPaczos, line 486: Could you add additional test that has a URL with both exclusion types to verify we're parsing correctly?

VysotskiVadim: added

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1870,5 +1877,108 @@ public Builder snappingIncludeClosuresList(@Nullable List<Boolean> snappingClosu
...
+  public abstract static class Exclude {
+
+    @NonNull
+    public static Builder builder() {
+      return new AutoValue_RouteOptions_Exclude.Builder();
+    }
+
+    @Nullable
+    static Exclude fromUrlQueryParameter(@Nullable String queryParameter) {
+      if (queryParameter == null || queryParameter == "") {
```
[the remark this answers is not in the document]
VysotskiVadim: Thank you, mr Bot

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -384,16 +395,17 @@ public void baseUrlWithLastSlash() {
...
     List<Point> coordinates = new ArrayList<>();
     coordinates.add(Point.fromLngLat(-122.4003312, 37.7736941));
     coordinates.add(Point.fromLngLat(-122.4187529, 37.7689715));
 
     RouteOptions options = RouteOptions.builder()
-      .profile(DirectionsCriteria.PROFILE_DRIVING)
-      .coordinatesList(coordinates)
-      .waypointNames("my starting position;my destination")
-      .build();
+        .profile(DirectionsCriteria.PROFILE_DRIVING)
```
VysotskiVadim, line 405: sometimes we use 2 and sometimes 4 spaces. Which option is correct, I will reformat all the file.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -370,14 +370,17 @@ public static Builder builder() {
...
+   * @return a comma separated string where each element matches one of
+   *   the {@link DirectionsCriteria.ExcludeCriteria} exclusion or a point
    */
   @Nullable
   public abstract String exclude();
 
   /**
-   * Exclude certain road types from routing. The default is to not exclude anything from the
-   * profile selected. The following exclude flags are available for each profile:
+   * Exclude certain road types and points from routing. The default is to not exclude anything
```
VysotskiVadim, line 382: Attention ⚠️ 
How do you think, is that a logical breaking change? I have doubts about whether I need to filter out points from that filed. I think no. What do you think?

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
Guardiola31337: Historically `mapbox-java` follows the 4 spaces Java formatting and https://github.com/mapbox/mapbox-java/tree/main/config/checkstyle as opposed to the Kotlin 2 spaces code style we have in https://github.com/mapbox/mapbox-navigation-android. I'd rather to stick with the 4 spaces for consistency.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1460,6 +1483,24 @@ public Builder excludeList(@Nullable List<String> exclude) {
...
+    /**
+     * Exclude certain road types or points from routing.
+     * The default is to not exclude anything from the profile selected.
+     *
+     * @param exclude an object of excludes that are used during the request.
+     *   Use {@link Exclude.Builder} to build exclude.
+     * @return this builder for chaining options together
+    */
+    @NonNull
+    public Builder excludeObject(@Nullable Exclude exclude) {
```
Guardiola31337, line 1495: Why is this `@Nullable`?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
```diff
@@ -0,0 +1,210 @@
...
+    ArrayList<String> criteria = new ArrayList<>();
+
+    List<String> excludeItems = ParseUtils.parseToStrings(queryParameter, ",");
+    if (excludeItems != null) {
+      for (String excludeItem : excludeItems) {
+        if (excludeItem.startsWith("point(") && excludeItem.endsWith(")")) {
+          Point point = parsePoint(excludeItem);
+          points.add(point);
+        } else if (excludeItem.contains("(") && excludeItem.contains(")")) {
+          // Ignoring the unexpected type of data. Update the library to get support of new types
```
Guardiola31337, line 116: Does this mean that you can't provide "exclude points" using the old `@Deprecated` APIs?

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsExcludeTest.java
```diff
@@ -0,0 +1,133 @@
...
+
+import java.util.Arrays;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+public class RouteOptionsExcludeTest {
+  @Test
+  public void parseTwoPointsWithTwoCriteriaMixed() {
```
Guardiola31337, line 15: NIT Normally is good to structure the tests following the triple A 👀 https://speakerdeck.com/guardiola31337/elegant-unit-testing-mobilization-2016?slide=40 http://wiki.c2.com/?ArrangeActAssert

```java
@Test
public void parseTwoPointsWithTwoCriteriaMixed() {
    // Arrange
    
    // Act
    
    // Assert
}
```

Note the empty lines between Arrange / Act and Act / Assert.

The same thing applies to the tests below 👇

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -421,22 +433,157 @@ public void routeOptions_toUrl_fromUrl_withEncodedChars() {
...
+        .waypointNames(";Улица Яна Чечота 7, Minsk 220045, Belarus")
+        .build();
 
     URL url = options.toUrl(ACCESS_TOKEN);
 
     assertEquals(expectedEncodedUrl, url.toString());
   }
 
+  @Test
+  public void routeOptionWithExcludedPoint() {
```
Guardiola31337, line 454: Same as above re: triple A

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -1,24 +1,34 @@
...
+
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
+
 import org.junit.Test;
 
 public class RouteOptionsTest extends TestUtils {
```
Guardiola31337, line 25: Should we add tests for checking backwards compatibility with old `@Deprecated` APIs to catch any potential regressions? What do you think?

```diff
diff --git a/services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java b/services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
index fba06a73..f8a10f6b 100644
--- a/services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
+++ b/services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
@@ -28,7 +28,7 @@ public class RouteOptionsTest extends TestUtils {
    */
   private static final String ROUTE_OPTIONS_JSON = "route_options_v5.json";
   private static final String ROUTE_OPTIONS_URL =
-      "https://api.mapbox.com/directions/v5/mapbox/driving/-122.4003312,37.7736941;-122.4187529,37.7689715;-122.4255172,37.7775835?access_token=pk.token&geometries=polyline6&alternatives=false&overview=full&radiuses=;unlimited;5.1&steps=true&avoid_maneuver_radius=200&bearings=0,90;90,0;&layers=-42;;0&continue_straight=false&annotations=congestion,distance,duration&language=ru&roundabout_exits=false&voice_instructions=true&banner_instructions=true&voice_units=metric&exclude=toll,ferry&include=hot,hov2&approaches=;curb;&waypoints=0;1;2&waypoint_names=;two;&waypoint_targets=;12.2,21.2;&enable_refresh=true&walking_speed=5.11&walkway_bias=-0.2&alley_bias=0.75&snapping_include_closures=;false;true&arrive_by=2021-01-01'T'01:01&depart_at=2021-02-02'T'02:02&max_height=1.5&max_width=1.4&metadata=true";
+      "https://api.mapbox.com/directions/v5/mapbox/driving/-122.4003312,37.7736941;-122.4187529,37.7689715;-122.4255172,37.7775835?access_token=pk.token&geometries=polyline6&alternatives=false&overview=full&radiuses=;unlimited;5.1&steps=true&avoid_maneuver_radius=200&bearings=0,90;90,0;&layers=-42;;0&continue_straight=false&annotations=congestion,distance,duration&language=ru&roundabout_exits=false&voice_instructions=true&banner_instructions=true&voice_units=metric&exclude=toll,ferry,point(11.0%2022.0)&include=hot,hov2&approaches=;curb;&waypoints=0;1;2&waypoint_names=;two;&waypoint_targets=;12.2,21.2;&enable_refresh=true&walking_speed=5.11&walkway_bias=-0.2&alley_bias=0.75&snapping_include_closures=;false;true&arrive_by=2021-01-01'T'01:01&depart_at=2021-02-02'T'02:02&max_height=1.5&max_width=1.4&metadata=true";
   private static final String ACCESS_TOKEN = "pk.token";
 
   private final String optionsJson = loadJsonFixture(ROUTE_OPTIONS_JSON);
@@ -180,7 +180,7 @@ public class RouteOptionsTest extends TestUtils {
   public void excludeIsValid_fromJson() {
     RouteOptions routeOptions = RouteOptions.fromJson(optionsJson);
 
-    assertEquals("toll,ferry", routeOptions.exclude());
+    assertEquals("toll,ferry,point(11.0 22.0)", routeOptions.exclude());
   }
 
   @Test
@@ -603,7 +603,7 @@ public class RouteOptionsTest extends TestUtils {
       .avoidManeuverRadius(200)
       .layers("-42;;0")
       .continueStraight(false)
-      .exclude(DirectionsCriteria.EXCLUDE_TOLL + "," + DirectionsCriteria.EXCLUDE_FERRY)
+      .exclude(DirectionsCriteria.EXCLUDE_TOLL + "," + DirectionsCriteria.EXCLUDE_FERRY + ",point(11.0 22.0)")
       .include(DirectionsCriteria.INCLUDE_HOT + "," + DirectionsCriteria.INCLUDE_HOV2)
       .geometries(DirectionsCriteria.GEOMETRY_POLYLINE6)
       .overview(DirectionsCriteria.OVERVIEW_FULL)
@@ -666,6 +666,7 @@ public class RouteOptionsTest extends TestUtils {
         .excludeList(new ArrayList<String>() {{
           add(DirectionsCriteria.EXCLUDE_TOLL);
           add(DirectionsCriteria.EXCLUDE_FERRY);
+          add("point(11.0 22.0)");
         }})
         .includeList(new ArrayList<String>() {{
           add(DirectionsCriteria.INCLUDE_HOT);
diff --git a/services-directions-models/src/test/resources/route_options_v5.json b/services-directions-models/src/test/resources/route_options_v5.json
index 8a96bc22..6af89f13 100644
--- a/services-directions-models/src/test/resources/route_options_v5.json
+++ b/services-directions-models/src/test/resources/route_options_v5.json
@@ -15,7 +15,7 @@
   "overview": "full",
   "steps": true,
   "annotations": "congestion,distance,duration",
-  "exclude": "toll,ferry",
+  "exclude": "toll,ferry,point(11.0 22.0)",
   "include": "hot,hov2",
   "voice_instructions": true,
   "banner_instructions": true,
```

Guardiola31337 (commented):
Overall is looking good @VysotskiVadim 

Left some minor comments.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
VysotskiVadim: To let users cleanup the exclude. 
```kotlin
  @Test
  public void cleanExcludes() {
    RouteOptions routeOptions = routeOptions().toBuilder()
        .excludeObject(null)
        .build();

    assertNull(routeOptions.exclude());
  }
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
VysotskiVadim: You can. This line for different case:
```kotlin
  @Test
  public void parseUnexpectedTypeOfParameterMixedWithPointAndCriteria() {
    Exclude exclude = Exclude.fromUrlQueryParameter("ferry,new(test),point(0.5 -0.5)");
    assertNotNull(exclude);
    assertEquals(1, exclude.criteria().size());
    assertEquals(1, exclude.points().size());
  }
```
Directions API team may add new data formats for exclude. I think the best way would be to just ignore them during the parsing.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
VysotskiVadim: Why not. Added, thanks

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsExcludeTest.java
VysotskiVadim: I agree, that AAA brings more clarity to tests. Sometimes tests are so complex that it's not easy to understand the border between setup and an act 🙂 
I added empty lines to some tests from PR. But some of the tests I find too trivial and I'm not sure if highlighting of their AAA structure will bring some benefit. What do you think about AAA structure highlighting for trivial tests?

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
VysotskiVadim: Thank you. Now I understand where the inconsistency comes from. My Android Studio uses settings from Mapbox Navigation Android. 

As I see the majority of code uses 2 as indent and continuation indent. So to have less changes in PR for easier review I stick to 2 indent size formatting.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
```diff
@@ -0,0 +1,210 @@
...
+        } else if (excludeItem.contains("(") || excludeItem.contains(")")) {
+          throw new IllegalArgumentException("Can't parse parameter " + excludeItem);
```
LukasPaczos, line 118: Let's remove these 2 cases and let the backend respond with an appropriate errors. We should only handle what we expect.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
LukasPaczos: I wouldn't consider this a breaking change. It just reflects the state of the server-side API itself which wasn't considered breaking since the format is a generic string.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
LukasPaczos: Sorry, now I see it's the other way around. Regardless, since the server-side API format is generic string, we cannot throw an exception on the parsing level ourselves. We probably need to remove that if-check.

VysotskiVadim: this line is about a case when you put a comma instead of the space in point
```kotlin
  @Test(expected = IllegalArgumentException.class)
  public void parsePointWithCommaDelimiter() {
    String invalidPoint = "point(5.0,7.0)";
    Exclude.fromUrlQueryParameter(invalidPoint);
  }
``` 
Do you suggest to ignore everything we can't parse? Like:
```kotlin
  @Test
  public void parsePointWithCommaDelimiter() {
    String invalidPoint = "point(5.0,7.0)";
    Exclude exclude = Exclude.fromUrlQueryParameter(invalidPoint);
    assertNull(exclude);
  }
```

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
Guardiola31337: > As I see the majority of code uses 2 as indent and continuation indent. So to have less changes in PR for easier review I stick to 2 indent size formatting.

I'd thought it was the other way around, but yeah ultimately not a big deal 👍

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsExcludeTest.java
Guardiola31337: In such cases, I personally prefer to add it too, for consistency.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
Guardiola31337: Yeah, I agree with @LukasPaczos providing the right format should be managed by the end-user when integrating the API and check on the server side. I believe we should also remove the previous `if` https://github.com/mapbox/mapbox-java/pull/1362#discussion_r808503881 If in the future Directions API adds new data formats for exclude, we'll accommodate [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it).

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
VysotskiVadim: What if you build new route options from existing route options as on example ☝️

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
VysotskiVadim: resolving discussion as the code we're discussing has already been removed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java
VysotskiVadim: @LukasPaczos , @Guardiola31337 , updated parsing with a new concept: `Exclude` is just a type-safe way of working with known excludes.  All the unknown properties and flags are ignored now

Guardiola31337: Why even ignore it and return `null` from our side? Why can't we pass whatever we get from the client (assuming is well formed / formatted and let the server crash if not? https://github.com/mapbox/mapbox-java/blob/f2c30793a4054248bdb95bf511d53a3da2256662/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Exclude.java#L184

VysotskiVadim: @Guardiola31337 , the line 184 is for reading points, not for writing. You can write incorrect point like `exclude("point(test)")`, but you won't be able to read it as a `Point` because it can't be parsed

LukasPaczos (approved):

Issue #1344, opened 2021-12-14
roundtrip unrecognized JSON properties


LukasPaczos:
Refs https://github.com/mapbox/mapbox-directions-swift/issues/637.

We should find a way to keep and serialize back any, at the time of serializing, unrecognized properties to support forward compatibility.

cc @1ec5 @SiarheiFedartsou @Guardiola31337

1ec5:
In principle, persisting and roundtripping unrecognized properties would be beneficial to any models that correspond to server API responses. It’s particularly needed for the Directions and Map Matching APIs, which may return undocumented response properties that Navigation Native may want to consume. However, multiple APIs routinely introduce beta request options that result in beta response properties.

Similarly, GeoJSON has a provision for _foreign members_, acknowledging that GeoJSON authoring tools often extend GeoJSON objects with arbitrary properties. Sometimes foreign members are vendor-specific, while sometimes they’re defined by separate standards like GeoJSON-T and TopoJSON. mapbox/turf-swift#175 added Swift support for round-tripping foreign members (but not interpreting them); it would be nice to support them consistently across languages.

LukasPaczos:
Tagging @tobrun who is also looking into this.

VysotskiVadim:
We can add support of unrecognised properties to the generated type adapters. The type adapters we use are generated by [auto-value-gson library. ](https://github.com/rharter/auto-value-gson)What if we fork it and add support of unrecognised json properties. It shouldn't be too hard to generate handling of unrecognised properties in our type adapters.

Discussion with library's authors: https://github.com/rharter/auto-value-gson/issues/266. Spoiler: they didn't like the idea.

VysotskiVadim:
PR: https://github.com/mapbox/mapbox-java/pull/1394

VysotskiVadim:
Waiting for the official fork to be created + need to reconsider with team where to store binary artefacts from the fork.

1ec5:
[...]

PR #1365, opened 2022-02-17
experimental EV routing request parameters and data

LukasPaczos:
Introduces a set of experimental request and response parameters supporting EV routing capabilities. Not all parameters are yet supported, even on staging endpoints.

All the experimental parameters are subject to change and evolve, therefore, I opted to avoid documenting anything specific for now.


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/ExperimentalWaypointMetadata.java
```diff
@@ -0,0 +1,190 @@
...
+
+    /**
+     * Object representing experimental value.
+     * <p>
+     * All available experimental values are subject to change at any time.
+     *
+      @param powerKw powerKw
+     */
+    @NonNull
+    public abstract Builder powerKw(@Nullable Integer powerKw);
```
purew, line 181: Question regarding capitalization. Watt is usually symbolized with `W` (capital letter) and kilo by `k` which doesn't translate well to snake case :) Is `powerKiloWatt` too verbose?

I'm likely overanalyzing this. Feel free to ignore me :)

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -482,6 +487,22 @@ private RouteOptions routeOptions() {
...
       .enableRefresh(true)
       .metadata(true)
+      .experimental(
+        RouteOptions.Experimental.builder()
+          .engine("electric")
+          .evInitialCharge(80000)
+          .evMaxCharge(80000)
+          .evConnectorTypes("ccs_combo_type1,ccs_combo_type2")
+          .energyConsumptionCurve("0,300;20,160;80,140;120,180")
+          .auxiliaryConsumption("auxiliaryConsumption")
```
purew, line 497: The auxilliare consumption is currently not prioritized as that information can be passed through overloaded via the charging-curve parameter https://github.com/mapbox/nav-api-intake/issues/893#issuecomment-969241997

services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
```diff
@@ -153,7 +167,7 @@
     @Field("radiuses") String radiuses,
     @Field("steps") Boolean steps,
     @Field("bearings") String bearings,
-    @Query("avoid_maneuver_radius") Integer avoidManeuverRadius,
+    @Field("avoid_maneuver_radius") Integer avoidManeuverRadius,
```
purew, line 170: Unrelated change?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/ExperimentalWaypointMetadata.java
LukasPaczos: Yea, this is an unfortunate setup. On top of snake case, code style verification in this project forbids any acronyms in the API signature (more than 2 capital letters next to each other), so `powerKW` cannot be used and I don't think we need to necessarily revisit this now.

I like the `powerKiloWatt` suggestion :+1:

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -18,7 +18,8 @@
...
+    "https://api.mapbox.com/directions/v5/mapbox/driving/-122.4003312,37.7736941;-122.4187529,37.7689715;-122.4255172,37.7775835?access_token=pk.token&geometries=polyline6&alternatives=false&overview=full&radiuses=;unlimited;5.1&steps=true&avoid_maneuver_radius=200&bearings=0,90;90,0;&layers=-42;;0&continue_straight=false&annotations=congestion,distance,duration&language=ru&roundabout_exits=false&voice_instructions=true&banner_instructions=true&voice_units=metric&exclude=toll,ferry&include=hot,hov2&approaches=;curb;&waypoints=0;1;2&waypoint_names=;two;&waypoint_targets=;12.2,21.2;&enable_refresh=true&walking_speed=5.11&walkway_bias=-0.2&alley_bias=0.75&snapping_include_closures=;false;true&arrive_by=2021-01-01'T'01:01&depart_at=2021-02-02'T'02:02&max_height=1.5&max_width=1.4&metadata=true&engine=electric&ev_initial_charge=80000&ev_max_charge=80000&ev_connector_types=ccs_combo_type1,ccs_combo_type2&energy_consumption_curve=0,300;20,160;80,140;120,180&auxiliary_consumption=auxiliaryConsumption&ev_ascent=5.0&ev_descent=6.0&ev_charging_curve=0,100000;40000,70000;60000,30000;80000,10000&ev_max_ac_charging_power=1&ev_min_charge_at_destination=2&ev_min_charge_at_charging_station=3";
```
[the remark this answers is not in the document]
LukasPaczos: :+1: unrelated for this PR though since these are unit tests and we mock the result anyway.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
LukasPaczos: Would you recommend removing this param for now?

services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
LukasPaczos: Yes, I can extract it out to a separate commit.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
```diff
@@ -416,4 +416,47 @@ private DirectionsCriteria() {
   })
   public @interface ApproachesCriteria {
   }
+
+  // EXPERIMENTAL
```
VysotskiVadim, line 420: Have we always highlighted experimentation API by a comment? Maybe it's better to add an 'Unstable' or 'Experimental' annotation like [this](https://stackoverflow.com/a/32616239/3599375) ?

LukasPaczos: This particular comment is just for us to organize the code, unless you mean all the other actual javadoc comments?

There's no precedent for experimental mapbox-java APIs and I quickly tested out adding a custom annotation that uses `@Documented` but it didn't do the trick, javadoc still requires me to write documentation for the annotation method.

LukasPaczos: Or are you thinking about a supplement in form of annotation, next to javadoc comments saying that the method is experimental?

VysotskiVadim: I was thinking about an `Experimental` annotation which will be visible in java-docs just as annotation. It could be a standard way to mark an API as not stable instead of manually writing messages about stability

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1871,4 +1979,313 @@ public Builder snappingIncludeClosuresList(@Nullable List<Boolean> snappingClosu
...
+
+  /**
+   * Object representing experimental value.
+   * <p>
+   * All available experimental values are subject to change at any time.
+   */
+  @AutoValue
+  public abstract static class Experimental extends DirectionsJsonObject {
+
+    static String[] experimentalParameters = new String[] {
```
VysotskiVadim, line 1991: It seems like the `experimentalParameters` array is used only for lookup if a filed is experimental. Maybe it's easier to keep all the experimental fields name in a `HashSet`?

LukasPaczos: @VysotskiVadim what would be a practical difference?

VysotskiVadim: You won't need to implement custom lookup with HashSet. Instead of 
```kotlin
        String property = URLDecoder.decode(query.substring(0, idx), UTF_8);
        String value = URLDecoder.decode(query.substring(idx + 1), UTF_8);

        boolean isExperimental = false;
        for (int i = 0; i < Experimental.experimentalParameters.length; i++) {
          if (property.equals(Experimental.experimentalParameters[i])) {
            isExperimental = true;
            if (!optionsJson.has("experimental")) {
              optionsJson.add("experimental", new JsonObject());
            }
            JsonObject experimental = optionsJson.getAsJsonObject("experimental");
            experimental.addProperty(property, value);
            break;
          }
        }
```

you will be able to do:
```kotin
        String property = URLDecoder.decode(query.substring(0, idx), UTF_8);
        String value = URLDecoder.decode(query.substring(idx + 1), UTF_8);

        boolean isExperimental = false;
        if (experimentalParameters.contains(property)) {
          isExperimental = true;
          if (!optionsJson.has("experimental")) {
            optionsJson.add("experimental", new JsonObject());
          }
          JsonObject experimental = optionsJson.getAsJsonObject("experimental");
          experimental.addProperty(property, value);
        }
        
```

Also runtime complexity of lookup in hash table is O(1) vs O(n) in array. But I don't think performance is a good reason for this change, because array has small and constant size.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/ExperimentalWaypointMetadata.java
```diff
@@ -0,0 +1,190 @@
...
+  public abstract String plugType();
+
+  /**
+   * Object representing experimental value.
+   * <p>
+   * All available experimental values are subject to change at any time.
+   */
+  @Nullable
+  @SerializedName("power_kw")
+  public abstract Integer powerKiloWatt();
```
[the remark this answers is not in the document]
LukasPaczos: Thanks for bringing this up :+1: 

This response data will only be returned if the `electric engine` experimental parameter is used. Since it shouldn't be used in any type of production environment, we shouldn't need to worry about compatibility for our case. 

That said, I don't want to setup a dangerous precedent where an "experimental" property would be introduced and released in a stable version of the library and then potentially conflict with a server side response much, much later down the line, causing parsing exceptions.

I will take another pass at this, to make this safer. My current idea is to introduce a new annotation we discussed above but not only for documentation reasons but also to ensure that a stable version of `mapbox-java` cannot be released with an experimental property. Planning to use the annotation processing and `BuildConfig` to see if I can throw a build-time (preferably) or runtime exception in these cases. If this won't work, then maybe some alternative standalone script.

VysotskiVadim (approved):
I'm okay if this changes will be released in not stable version of `mapbox-java`. The comments I left could be a blocker (maybe not 😄 ) only for a stable release.

LukasPaczos:
I'm going ahead and merging here. I've been trying to figure out some automated way of verifying the existence of experimental features, or gracefully falling back when they fail to parse, but couldn't find any in the timeboxed exercise.

Similarly to https://github.com/mapbox/mapbox-java/issues/1344, we'd need to change how the Auto-Value-Gson lib works to easily fallback to `null` if a property fails to parse. If we wanted custom type adapters without modifying the generated code, we'd most-likely also have to have custom return types (even for primitives) which isn't ideal either.

Anyway, the plan would be to revert this PR before the next stable release and budget more time for looking for other options to maintain experimental properties.

PR #1376, opened 2022-03-09
fixed an issue where "avoid_maneuver_radius" was represented as integer instead of double

LukasPaczos:
This re-aligns the value with the API. Even though this is technically a breaking change, the integer representation was functionally wrong so we should treat this change as a bug fix.


RingerJK:
@LukasPaczos I see in docs https://docs.mapbox.com/api/navigation/directions/ number range of `avoid_maneuver_radius ` is [0, 1000]. Are you sure that we need so accurate value, looks like 1 meter should be enough? (Gson is mapping double to integer well)

LukasPaczos:
The API allows for a `number` which also accepts floats and doubles, so `mapbox-java` needs to reflect that. I found this issue out when trying to parse request params created by Nav Native which were a floating point number and led to a crash. So it's less about what makes practical sense and more about the compatibility with the API.

Guardiola31337 (approved):
Breaking SemVer, although we can consider it a bugfix :shipit:

RingerJK (approved):

PR #1377, opened 2022-03-10
add ability to handle encoded coordinates in RouteOptions#toUrl

LukasPaczos:
`RouteOptions` use [java.net.URI](https://docs.oracle.com/javase/7/docs/api/java/net/URI.html) (which follows https://www.ietf.org/rfc/rfc2396.txt) to encode the [URL](https://docs.oracle.com/javase/7/docs/api/java/net/URL.html) returned by `RouteOptions#toUrl` and expects the same format in `RouteOptions#fromUrl`. The way we use commas and semicolons to delimit and lon/lat arguments of a coordinate, as well as collections of coordinates, conforms to the URI spec and we shouldn't need to encode commas and semicolons on the URI path. However, some clients could still pass the path arguments encoded so this adds handling for these potential cases.


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -873,15 +873,15 @@ public static RouteOptions fromUrl(@NonNull URL url) {
...
-    optionsJson.addProperty("coordinates", pathElements[5]);
-
-    String[] queryElements = url.getQuery().split("&");
-    for (String query : queryElements) {
-      int idx = query.indexOf("=");
-      try {
+    try {
+      String[] pathElements = url.getPath().split("/");
+      optionsJson.addProperty("user", URLDecoder.decode(pathElements[3], UTF_8));
+      optionsJson.addProperty("profile", URLDecoder.decode(pathElements[4], UTF_8));
```
VysotskiVadim, line 879: Can `user` and `profile` contain semicolons?

VysotskiVadim (approved):
LGTM

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
LukasPaczos: Not really, I just added it to keep things consistent :upside_down_face:

PR #1385, opened 2022-03-25
Added route index to the MapMpatchingMatching

cafesilencio:
The MapMpatchingMatching class was missing the route index field for the `DirectionsRoute`. This would create an invalid route when transforming the `DirectionsRoute` to a `NavigationRoute` since the route index field is required by the `NavigationRoute`.


services-matching/src/main/java/com/mapbox/api/matching/v5/models/MapMatchingMatching.java
```diff
@@ -33,6 +33,15 @@ public static Builder builder() {
     return new AutoValue_MapMatchingMatching.Builder();
   }
 
+  /**
+   * The index of this route in the original network response.
+   *
+   * @return string of an int value representing the index
+   * @since 4.4.0
```
LukasPaczos, line 40: Let's drop the `since`.

LukasPaczos (approved):

services-matching/src/main/java/com/mapbox/api/matching/v5/models/MapMatchingMatching.java
cafesilencio: Copy paste gets me every time.

PR #1390, opened 2022-04-03
Codecov is back

RingerJK:


VysotskiVadim (approved):
Nice ❤️ 

Tested following cases:

- sha sum of `pgp_keys.asc` doesn't match the file
- `codecov.SHA256SUM` was modified 
- `codecov` binary was modified

PR #1394, opened 2022-04-04
Roundtrip of unrecognised JSON properties

VysotskiVadim:
## Goal
Implementation of https://github.com/mapbox/mapbox-java/issues/1344
Models from `mapbox-java` should not erase unrecognised json properties during `fromJson` -> `toJson`
#### example
If we have a model with a single property `a`
```java
public abstract Model {
    public abstract String a();
}
```
and a JSON with properties `a` and `b`
```json
{"a":"value", "b":"value"}
```
the model should let unrecognised fields to do a roundtrip
```java
String json = {"a":"value", "b":"value"}
assertEquals(json, Model.fromJson(json).toJson());
```

## Implementation

I forked [the library which generates gson type adapters ](https://github.com/rharter/auto-value-gson)for us. Now if there are some unrecognised fields a type adapter put them into `unrecognised` fields map. 
I will share a link when fork will be publicly available.


services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/DirectionsResponseTest.java
```diff
@@ -41,15 +49,27 @@ public void fromJson_correctlyBuildsFromJson() throws Exception {
   }
 
   @Test
-  public void testToFromJson() throws Exception {
-    String json = loadJsonFixture(DIRECTIONS_V5_PRECISION6_FIXTURE);
-    DirectionsResponse responseFromJson1 = DirectionsResponse.fromJson(json);
+  public void testToFromJsonWithRealResponse() throws Exception {
+    Gson gson = new GsonBuilder().create();
+    String originalJson = loadJsonFixture(DIRECTIONS_V5_PRECISION6_FIXTURE_ARTIFICIAL_FIELDS);
```
VysotskiVadim, line 54: Added `DIRECTIONS_V5_PRECISION6_FIXTURE_ARTIFICIAL_FIELDS` just because `DirectionsResponse` always adds `routeIndex` and `requestUuid`.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Bearing.java
```diff
@@ -3,12 +3,14 @@
...
 
+import java.util.Map;
+
 /**
  * A value pair of an {@link #angle} which is a clockwise value from true north between 0 and 360,
  * and a {@link #degrees()} which is the range of degrees by which the angle can deviate.
  */
 @AutoValue
-public abstract class Bearing {
+public abstract class Bearing extends DirectionsJsonObject {
```
VysotskiVadim, line 13: Some objects didn't extend `DirectionsJsonObject`. Now all models extend it

VysotskiVadim:
Did rough testing of performance. No microbenchmarks was involved, but I guess we can less or more understand performance impact.

#### Test 1
Deserialised and serialised directions repose 1000 times:
```java
   for (int i = 0; i < 1000; i++) {
      DirectionsResponse response = DirectionsResponse.fromJson(regularJsonString);
      String json = response.toJson();
    }
```
**2210 ms**

#### Test 2
Deserialised and serialised mutated directions repose 1000 times:
```java
   for (int i = 0; i < 1000; i++) {
      DirectionsResponse response = DirectionsResponse.fromJson(mutatedJsonString);
      String json = response.toJson();
    }
```
**3075 ms**
Mutated json is a directions response where to each property was added a few unrecognised fields. See 

#### Test 3
Tested mutated json serialisation and deserialisation 1000 times on main branch:
```java
   for (int i = 0; i < 1000; i++) {
      DirectionsResponse response = DirectionsResponse.fromJson(mutatedJson);
      String back = response.toJson();
    }
```
**2333 ms**

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
```diff
@@ -501,5 +502,8 @@ public abstract static class Builder {
      * @since 3.0.0
      */
     public abstract BannerComponents build();
+
+    @Nullable
```
RingerJK, line 506: why `Builder` is `Nullable`?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
```diff
@@ -501,5 +502,8 @@ public abstract static class Builder {
...
+    abstract Builder unrecognised(Map<String, Object> value);
```
RingerJK, line 507: we know that unrecognized fields are Strings, this way, let's make `Map<String, String>`

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Bearing.java
RingerJK: we're potentially expect that any object can have unrecognized fields, let's add `abstract Builder unrecognised(Map<String, Object> value);` to `DirectionsJsonObject`

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
VysotskiVadim: it's not only strings. It can also be `JsonObject`, `JsonArray`, etc.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
```diff
@@ -1,18 +1,22 @@
...
 import java.io.Serializable;
+import java.util.Map;
 
 /**
  * Provides a base class for Directions model classes.
  *
  * @since 3.4.0
  */
-public class DirectionsJsonObject implements Serializable {
+public abstract class DirectionsJsonObject implements Serializable {
```
LukasPaczos, line 19: This is a breaking change (not sure how impactful it would actually be in practice). Could we make the methods return empty and implement downstream, how would it look like?

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/DirectionsResponseTest.java
```diff
@@ -41,15 +49,27 @@ public void fromJson_correctlyBuildsFromJson() throws Exception {
...
+    JsonObject originalJsonObject = gson.fromJson(originalJson, JsonObject.class);
+
+    DirectionsResponse model = DirectionsResponse.fromJson(originalJson);
+    JsonObject jsonFromObject = gson.fromJson(model.toJson(), JsonObject.class);
+
+    assertEquals(originalJsonObject, jsonFromObject);
```
LukasPaczos, line 60:
```suggestion
    DirectionsResponse model = DirectionsResponse.fromJson(originalJson);
    String jsonFromModel = model.toJson();

    assertEquals(originalJson, jsonFromModel);
```
We want to ensure the resulting strings are correct, the object wrapping step seems unnecessary.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/utils/MutateJsonUtil.java
```diff
@@ -0,0 +1,36 @@
...
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+
+import java.util.Map;
+
+public class MutateJsonUtil {
+
+  public static void mutateJson(JsonElement element) {
```
LukasPaczos, line 12: Could you add a short comment on what's the goal for this chunk of code? It's not too hard to follow but would be good to have an overview for future refactors.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
RingerJK: why do you want to operate with these fields as JsonObject or JsonArrays? Looks like we only need to serialize/deserialize them. So no need to keep Objects

LukasPaczos: I was actually thinking whether we could consider making this map public (no in builder but in the resulting type), so that if someone would like to access the unrecognized fields, they don't need to serialize the type first. I believe this will be needed soon (for example for experimental EV properties).

RingerJK: @LukasPaczos Experimental EV and other options are allocated in RouteOptions. We don't extend RouteOptions with Unrecognized fields I think 🤔 Suppose, it's separate initiative, not really connected with this PR

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Bearing.java
VysotskiVadim: added, thank you for the suggestion

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
VysotskiVadim: nice catch 👍

VysotskiVadim: AutoValue won't generate fields for not abstract methods

VysotskiVadim: I see a few possible solutions right now. 

### Do breaking change

Maybe we can make ourself believe that it's not really a breaking change using following arguments: it more like a fix of a bug, nobody should have created an instance of `DirectionsJsonObject`, it doesn't make any sense to create an instance of `DirectionsJsonObject`. But `abstract` do really changes API.

### Workaround

I can create an abstract class `BaseDirectionsJsonObject` which is going to be a parent of `DirectionsJsonObject` and move `unrecognised` filed there. I find this solution ugly but it won't break API.

VysotskiVadim: I think that we need to decide if somebody could be affected by not being able to instantiate `DirectionsJsonObject`. If we're sure that we will affect nobody, and even if we will we're ready to reply in their issues on github, then I let's break the rule.
If we don't feel confident, let's workaround

VysotskiVadim: I feel confident. I can't imaging a case when somebody instantiate `DirectionsJsonObject` but if somebody does it, I'd like to learn their case when they open an issue on our github 🙂 
@LukasPaczos , @RingerJK , what do you think about it?

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/DirectionsResponseTest.java
VysotskiVadim: Strings won't the same. An order of fields changes when json gets serialised to java model and then gets deserialised back. And I think it's acceptable behaviour. `JsonObject` compares that 2 JSONs has the same set of fields with the same value and it doesn't care about order.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/utils/MutateJsonUtil.java
VysotskiVadim: Added

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
VysotskiVadim: `RouteOptions` extends `DirectionsJsonObject` so it already has unrecognised fields

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
LukasPaczos: @VysotskiVadim I agree. There's no practical reason to extend the `DirectionsJsonObject` for developers, I think we can break here. Let's clearly state this in the changelog and go from there.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/DirectionsResponseTest.java
LukasPaczos: That makes sense :+1: I guess the equality of a serialized form is just a side effect, not a guarantee.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/BannerComponents.java
VysotskiVadim: fixed, thank you

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
```diff
@@ -27,4 +32,13 @@ public String toJson() {
     gson.registerTypeAdapter(Point.class, new PointAsCoordinatesTypeAdapter());
     return gson.create().toJson(this);
   }
+
+  @Nullable
+  @UnrecognizedJsonProperties
+  abstract Map<String, SerializableJsonElement> unrecognised();
```
LukasPaczos, line 38:
```suggestion
  abstract Map<String, SerializableJsonElement> unrecognized();
```

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
```diff
@@ -27,4 +32,13 @@ public String toJson() {
...
+
+  abstract static class Builder<T extends Builder> {
+    @NonNull
+    abstract T unrecognised(Map<String, SerializableJsonElement> value);
```
LukasPaczos, line 42:
```suggestion
    abstract T unrecognized(Map<String, SerializableJsonElement> value);
```

services-directions/build.gradle
```diff
@@ -8,9 +8,9 @@ dependencies {
   compileOnly dependenciesList.supportAnnotation
 
   // AutoValue
-  compileOnly dependenciesList.autoValue
```
LukasPaczos, line 11: Why is this not needed anymore?

services-matrix/build.gradle
```diff
@@ -9,6 +9,7 @@ dependencies {
   // AutoValue
   compileOnly dependenciesList.autoValue
   compileOnly dependenciesList.autoValueGson
+    implementation dependenciesList.autoValueGsonRuntime
```
LukasPaczos, line 12: nit: formatting, and in other places

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -889,6 +890,9 @@ public static RouteOptions fromUrl(@NonNull URL url) {
         int idx = query.indexOf("=");
         String property = URLDecoder.decode(query.substring(0, idx), UTF_8);
         String value = URLDecoder.decode(query.substring(idx + 1), UTF_8);
+        if (property.equals(ACCESS_TOKEN_URL_PARAM_NAME)) {
+          continue;
+        }
```
LukasPaczos, line 895: Do we have a test that covers this already?

VysotskiVadim: yep, I added this line because one of tests failed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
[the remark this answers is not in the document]
VysotskiVadim: https://github.com/mapbox/mapbox-java/pull/1416

VysotskiVadim: renamed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
VysotskiVadim: renamed

services-matrix/build.gradle
VysotskiVadim: fixed, thank you

services-directions/build.gradle
VysotskiVadim: `dependenciesList.autovalue` is annotation processor, we don't need to connect it one more time as `compileOnly`. Docs: https://github.com/google/auto/blob/master/value/userguide/index.md#with-gradle.

VysotskiVadim: reverted for consistency. I don't remember why I removed this one 🙂 
I guess everything still works because the project gets auto value transitively

VysotskiVadim:
@RingerJK , @LukasPaczos , please review companion PR https://github.com/mapbox/auto-value-gson/pull/2

LukasPaczos (approved):
LGTM, nice work!

PR #1396, opened 2022-04-07
docs: added `directions-models` and `directions-refresh-models`

RingerJK:
cc @mikelmaron

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -25,10 +25,14 @@ javadoc:
 	mkdir documentation/libjava-geojson/
 	mkdir documentation/libjava-turf/
 	mkdir documentation/libjava-services/
+	mkdir documentation/libjava-services-directions-models/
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -20,15 +20,20 @@ build-cli:
 	./gradlew shadowJar
 
 javadoc:
+	rm -rf documentation
```
mikelmaron, line 23: somehow this scares me a little :) suggest add a `clean` option to Makefile, or adding `-p` to mkdir to not error if the directory exists

RingerJK: actually, `documentation` is the temporary dir, which creates a transit place between dirs where docs are generated and publisher-production. It was useful for local tests. I'm gonna revert this line

mikelmaron (approved):

mikelmaron:
@RingerJK approved

PR #1399, opened 2022-04-11
Docs v6.4.0-beta.4: `libjava-services-directions-models` and `libjava-services-directions-refresh-models` modules

RingerJK:


mikelmaron:
@RingerJK do you have an idea why circleci is failing?

RingerJK:
> @RingerJK do you have an idea why circleci is failing?

@mikelmaron it's fine, we don't have CI config for publisher-production branch

mikelmaron (approved):

Issue #1398, opened 2022-04-11
Inconsistent map-matching api documentation


RingerJK:
>There seems to be either missing or inconsistent api documentation for the map matching api
https://docs.mapbox.com/api/navigation/map-matching/
For example on the documentation page, there is no optional parameter for banner_instructions however requests likes these seem to work. On https://docs.mapbox.com/api/navigation/directions/, banner_instructions, is clearly displayed as an optional parameter.

ref https://github.com/mapbox/android-docs/issues/2044

PR #1400, opened 2022-04-13
mapmathing: - added to services docs; - fix api-ref docs

RingerJK:
Closes https://github.com/mapbox/mapbox-java/issues/1398

### DIFF
--- a/services/build.gradle
+++ b/services/build.gradle
@@ -6,7 +6,7 @@ sourceSets {
                          '../services-optimization/src/main/java', '../services-staticmap/src/main/java',
                          '../services-speech/src/main/java', '../services-tilequery/src/main/java',
                          '../services-route-tiles/src/main/java', '../services-directions-refresh/src/main/java',
-                         '../services-isochrone/src/main/java']
+                         '../services-isochrone/src/main/java', '../services-matching/src/main/java']
 }
 
 dependencies {
--- a/services-matching/src/main/java/com/mapbox/api/matching/v5/MapboxMapMatching.java
+++ b/services-matching/src/main/java/com/mapbox/api/matching/v5/MapboxMapMatching.java
@@ -422,6 +422,10 @@ public abstract class MapboxMapMatching extends
      * Setting this will determine whether to return steps and turn-by-turn instructions. Can be
      * set to either true or false to enable or disable respectively. null can also optionally be
      * passed in to set the default behavior to match what the API does by default.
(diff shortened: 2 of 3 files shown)


services/build.gradle
```diff
@@ -6,14 +6,15 @@ sourceSets {
                          '../services-optimization/src/main/java', '../services-staticmap/src/main/java',
                          '../services-speech/src/main/java', '../services-tilequery/src/main/java',
                          '../services-route-tiles/src/main/java', '../services-directions-refresh/src/main/java',
-                         '../services-isochrone/src/main/java']
+                         '../services-isochrone/src/main/java', '../services-matching/src/main/java']
```
RingerJK, line 9: include `mapmatching` to `services` api-ref docs  

cc @Guardiola31337

services/build.gradle
```diff
@@ -6,14 +6,15 @@ sourceSets {
...
 }
 
 dependencies {
     api project(":services-core")
     api project(":services-geojson")
     api project(":services-directions-models")
     api project(":services-directions-refresh-models")
+    api project(":services-matching")
```
Guardiola31337, line 17: Why are we adding this but `services-isochrone` or `services-speech` (to name a couple) are not in here? 🤔

RingerJK: removed, it's not affecting docs

Guardiola31337: At some point we should revisit this (what we're including as part of `services`). Could you cut a ticket when you have a chance so we don't forget? 🙏

Guardiola31337 (approved):

services/build.gradle
RingerJK: done https://github.com/mapbox/mapbox-java/issues/1401

Guardiola31337: Thanks 🙇‍♂️

services/build.gradle
zugaldia: @RingerJK @Guardiola31337  When will we see the new map matching docs on https://docs.mapbox.com/android/java/api-reference/?

Guardiola31337: They're already available under [mapbox-java-services](https://docs.mapbox.com/android/java/api/libjava-services/6.4.0/) e.g. [`MapMatchingService`](https://docs.mapbox.com/android/java/api/libjava-services/6.4.0/com/mapbox/api/matching/v5/MapMatchingService.html) 🚀

PR #1410, opened 2022-04-21
Fix waypointNames '&' char encoding

### LINKED ISSUE #1409: `RouteOptions::toUrl` doesn't encode '&' character in query parameters
### Description

One of our customers detected a crash caused by requesting routes with waypoint names containing '&' character. From the captured crash log we determined the root cause of the crash to be incorrect `waypoint_names` argument encoding by the `RouteOptions::toUrl` method.

```
DIRECTIONS_API?access_token=<token>&geometries=polyline6&alternatives=true&overview=full&steps=true&avoid_maneuver_radius=111.11111450195312&bearings=13.6232796,90;&layers=0;&continue_straight=true&annotations=distance,congestion,maxspeed&language=en_SG&roundabout_exits=true&voice_instructions=true&banner_instructions=true&voice_units=metric&waypoints=0;1&waypoint_names=;Serangoon%20Garden%20Market%20&%20Food%20Centre&enable_refresh=true
```

### Crash details

Investigation and support ticket: https://github.com/mapbox/ask-navigation/issues/94

**Configuration:**
Operating systems + versions: Android 11
Device/simulator/emulator models: Samsung Galaxy Note20 Ultra 5G

**Steps to reproduce**
1. set a destination that contains & in the name of destination

```kotlin
val waypointNamesList=listof("","Serangoon Garden Market & Food Centre")
var routeOptionsBuilder = RouteOptions.builder()
....
routeOptionsBuilder.waypointNamesList(waypointNamesList)
routeOptionsBuilder.build()
```

2. make reroute happen

**Captured Crash log:**
```
Fatal Exception: java.lang.StringIndexOutOfBoundsException: String index out of range: -1
       at java.lang.String.substring(String.java:2064)
       at com.mapbox.api.directions.v5.models.RouteOptions.fromUrl(RouteOptions.java:883)
       at com.mapbox.navigation.base.route.NavigationRoute$Companion.create(NavigationRoute.java:71)
       at com.mapbox.navigation.base.internal.utils.DirectionsResponseUtilsKt$parseDirectionsResponse$2.invokeSuspend(DirectionsResponseUtils.kt:19)
       at kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:33)
       at kotlinx.coroutines.DispatchedTask.run(DispatchedTask.kt:106)
       at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1167)
       at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:641)
       at java.lang.Thread.run(Thread.java:923)
```

korshaknn:
closes #1409 
`&` char in Utf-8 is `%26`.
we can't replace it in the next block:
```
if (waypointNames() != null) {
      sb.append(String.format("&waypoint_names=%s", waypointNames()));
}
```
because if we do something like this 
```
sb.append(String.format("&waypoint_names=%s", waypointNames().replace("&", "%26")));
```
we will get invalid chars while building a final Url `URL(encodedUri.toASCIIString())`
because `%26` will be encoded to `%2526` (`%` char is `%25` in Utf-8)
and we will get invalid `%2526` instead of `%26`.

That's why I replace the char on a final step
```
URL(encodedUri.toASCIIString().replace("%20&%20", "%20%26%20"));
```
I check for start+end space characters (`%20`) not to affect `&` chars that are used to connect url params.


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1104,7 +1104,7 @@ public URL toUrl(@NonNull String accessToken) {
         decodedUrl.getQuery(),
         decodedUrl.getRef()
       );
-      return new URL(encodedUri.toASCIIString());
+      return new URL(encodedUri.toASCIIString().replace("%20&%20", "%20%26%20"));
```
LukasPaczos, line 1107: This will only work for strings like "Me & you" but if the name is "My &cafe&" we would still crash.

Could you add a comment in code that this is just a specific workaround until we find a better solution to fix the encoding in general? I don't think we should close https://github.com/mapbox/mapbox-java/issues/1409 just yet.

korshaknn: done

LukasPaczos (approved):

PR #1416, opened 2022-04-25
API to access unrecognized fields

VysotskiVadim:
The PR adds API to access unrecognised fields

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,8 @@
 Mapbox welcomes participation and contributions from everyone.
 
 ### main
-- Added roundtrip of unrecognized JSON properties [#1394](https://github.com/mapbox/mapbox-java/pull/1394)
-- Added API to access unrecognized JSON properties [#1416](https://github.com/mapbox/mapbox-java/pull/1416)
+- Added roundtrip of unrecognized JSON properties. [#1394](https://github.com/mapbox/mapbox-java/pull/1394)
+- Added API to access unrecognized JSON properties. See `#getUnrecognizedProperty` and `#getUnrecognizedPropertiesNames` in each Directions API model object.  [#1416](https://github.com/mapbox/mapbox-java/pull/1416)
 
 ### v6.4.1 - April 22, 2022
 - Added correct escaping for `<white space>&<white space>` pattern in query parameters for `RouteOptions#toUrl`. [#1410](https://github.com/mapbox/mapbox-java/pull/1410).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 ### main
 - Added roundtrip of unrecognized JSON properties [#1394](https://github.com/mapbox/mapbox-java/pull/1394)
+- Added API to access unrecognized JSON properties [#1416](https://github.com/mapbox/mapbox-java/pull/1416)
 
 ### v6.4.1 - April 22, 2022
 - Added correct escaping for `<white space>&<white space>` pattern in query parameters for `RouteOptions#toUrl`. [#1410](https://github.com/mapbox/mapbox-java/pull/1410).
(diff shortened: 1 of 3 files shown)


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
```diff
@@ -33,6 +34,25 @@ public String toJson() {
...
+    JsonElement result = null;
+    Map<String, SerializableJsonElement> recognisedProperties = unrecognised();
+    if (recognisedProperties != null) {
+      SerializableJsonElement property = recognisedProperties.get(propertyName);
+      if (property != null) {
+        result = property.getElement();
+      }
+    }
+    return result;
+  }
```
LukasPaczos, line 54:
```suggestion
  /**
   * Access JSON property that wasn't recognized during JSON deserialization.
   * This may be useful to access experimental properties.
   * @param propertyName name of a json property
   * @return value of a requested property or null if the requested property doesn't exist.
   */
  @Nullable
  public JsonElement getUnrecognizedProperty(String propertyName) {
    JsonElement result = null;
    Map<String, SerializableJsonElement> recognizedProperties = unrecognized();
    if (recognizedProperties != null) {
      SerializableJsonElement property = recognizedProperties.get(propertyName);
      if (property != null) {
        result = property.getElement();
      }
    }
    return result;
  }
```
:upside_down_face:

LukasPaczos (commented):
This somewhat obfuscates what's available (one would have to attach the debugger and see the contents of the map) but maybe that's for the better. The fields should be access based on the Directions API docs.

I don't have a strong opinion on this so let's leave as is and see how it goes during pre-releases phase.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
VysotskiVadim: oh, again 🤦

VysotskiVadim:
> This somewhat obfuscates what's available (one would have to attach the debugger and see the contents of the map) but maybe that's for the better.

@LukasPaczos , I added API to get properties names.
Ready for one more round of review

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsJsonObject.java
```diff
@@ -33,6 +36,41 @@ public String toJson() {
...
   }
 
+  /**
+   * Use this method to access JSON property that wasn't recognized during JSON serialization.
+   * This may be useful to access experimental properties.
+   * @param propertyName name of a json property
+   * @return value of a requested property or null if the requested property doesn't exist.
+   */
+  @Nullable
+  public final JsonElement getUnrecognizedProperty(String propertyName) {
```
RingerJK, line 46: suppose I missed: why do we just not return the `map<String, JsonElement>`?

VysotskiVadim: Because under the hood they are `Map<String, SerializableJsonElement>`. From the other side, why not transform one map to another every time field is accessed? what do you think?

LukasPaczos: We need to expose unrecognized properties to support experimental API features (like EV routing waypoint properties or state of charge annotations).

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 ### main
 - Added roundtrip of unrecognized JSON properties [#1394](https://github.com/mapbox/mapbox-java/pull/1394)
+- Added API to access unrecognized JSON properties [#1416](https://github.com/mapbox/mapbox-java/pull/1416)
```
LukasPaczos, line 7:
```suggestion
- Added roundtrip of unrecognized JSON properties. [#1394](https://github.com/mapbox/mapbox-java/pull/1394)
- Added API to access unrecognized JSON properties. See `#getUnrecognizedProperty` and `#getUnrecognizedPropertiesNames` in each Directions API model object.  [#1416](https://github.com/mapbox/mapbox-java/pull/1416)
```

LukasPaczos (approved):
LGTM

PR #1421, opened 2022-04-27
RouteLegRefresh: added incidents

RingerJK:


VysotskiVadim:
@RingerJK , what do you think about adding an example of json with incidents from Directions API and test if it can be serialised to the model?

RingerJK:
> @RingerJK , what do you think about adding an example of json with incidents from Directions API and test if it can be serialised to the model?

that's a good point. Extended current jsons with tests, check the branch

LukasPaczos (approved):

LukasPaczos:
Ah, could you please also add a changelog to unreleased?

RingerJK:
my bad, I didn't know we started to add changelog for mapbox-java, will cut another PR

RingerJK:
changelog pull request https://github.com/mapbox/mapbox-java/pull/1435

PR #1433, opened 2022-05-06
improve RouteOptions query parameter encoding

LukasPaczos:
Closes https://github.com/mapbox/mapbox-java/issues/1409 by improving the query parameter encoding by aligning it fully with the specification described in [URLEncoder](https://docs.oracle.com/javase/8/docs/api/java/net/URLEncoder.html) class.

I tested the new encoded strings with the Directions API and didn't find any incompatibilities.


services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -101,20 +100,6 @@ public void testRouteOptionsFromUrl() throws IOException {
     );
   }
 
-  @Test
-  public void testRouteOptionsFromUrl_alreadyDecoded() throws IOException {
```
LukasPaczos, line 105: This test was wrong - if a URL is decoded in full, it means that it's not a valid URL anymore (it contains unescaped reserved characters), and it's expected that `RouteOptions.fromUrl` might fail.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
```diff
@@ -511,8 +514,8 @@ public void routeOptionWithExcludedPoint() {
 
     List<String> queryParameters = Arrays.asList(url.getQuery().split("&"));
     assertTrue(
-      "url doesn't contain excluded point: " + url.toString(),
-      queryParameters.contains("exclude=point(1.0%202.0)")
+      "url doesn't contain excluded point: " + url,
+      queryParameters.contains("exclude=point%281.0+2.0%29")
```
VysotskiVadim, line 518: Does Direction API recognise an encoded point? Even if it does, I find this URL isn't easily readable. Is it possible to handle excluded points so that it stays like "point(1.0%202.0)"?

VysotskiVadim, line 518: Does Direction API recognise an encoded point? Even if it does, I find this URL isn't easily readable. Is it possible to handle excluded points so that it stays like "point(1.0%202.0)"?

VysotskiVadim (approved):
Looks good to me except for points encoding. Anyway it's not critical if it works.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/RouteOptionsTest.java
LukasPaczos: At this point I don't want to over-complicate the setup by adding exclusions. I did start with that approach to not only exclude `(` and `)` but also `;` and `,` in correct places, to keep the URL as readable as possible. This starts to get very complicated if we dive into recursive checks for nested characters and just wasn't worth it in my opinion.

We can reconsider separately.

LukasPaczos: I did check the point exclusion query and it doesn't thrown `InvalidArgument` on the backend, so I assume it works correctly but tagging @purew @mandeepsandhu for visibility, in case we should change the setup anywhere.

PR #1439, opened 2022-05-17
RouteOptions: added `max_weight`

RingerJK:


services-directions-models/src/test/resources/route_options_v5.json
```diff
@@ -32,6 +32,7 @@
   "depart_at": "2021-02-02'T'02:02",
   "max_height": 1.5,
   "max_width": 1.4,
+  "max_weight": 2.9,
```
LukasPaczos, line 35: Could you make this same change in https://github.com/mapbox/mapbox-java/blob/main/services-directions/src/test/resources/route_options_v5.json?

RingerJK: nice catch. Added

services-directions/src/main/java/com/mapbox/api/directions/v5/DirectionsService.java
```diff
@@ -95,6 +95,7 @@ Call<DirectionsResponse> getCall(
     @Query("depart_at") String departAt,
     @Query("max_height") Double maxHeight,
     @Query("max_width") Double maxWidth,
+    @Query("max_weight") Double maxWeight,
```
LukasPaczos, line 98: We're missing a `MapboxDirections` test for this.

RingerJK: in progress...

RingerJK: added

LukasPaczos (approved):
LGTM

PR #1443, opened 2022-05-25
Updated changelog and readme for 6.5.0 release

VysotskiVadim:


CHANGELOG.md
```diff
@@ -4,6 +4,9 @@ Mapbox welcomes participation and contributions from everyone.
 
 ### main
 
+### v6.5.0 - May 25, 2022
+- No additional changes
```
VysotskiVadim, line 8: I noticed that we always write this if nothing has changed since the last release. 

If I were `mapbox-java`'s user, I would prefer to see all updates since the latest stable version in every stable version changelog entry. @mapbox/navigation-android , WDYT?

PR #1457, opened 2022-07-01
affected route names

VysotskiVadim:
A new field in the [incident object](https://docs.mapbox.com/api/navigation/directions/#incident-object)


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/Incident.java
```diff
@@ -466,6 +473,12 @@ public abstract static class Builder extends DirectionsJsonObject.Builder<Builde
      */
     public abstract Builder numLanesBlocked(@Nullable Integer numLanesBlocked);
 
+    /**
+     * Sets list of roads names affected by the incident.
+     * @param affectedRoadNames list of roads names affected by the incident.
+     */
+    public abstract Builder affectedRoadNames(@Nullable List<String> affectedRoadNames);
```
Guardiola31337, line 480:
```suggestion
     */
     @NonNull
    public abstract Builder affectedRoadNames(@Nullable List<String> affectedRoadNames);
```

We can also add the above missing annotations 👍

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/IncidentTest.java
```diff
@@ -17,31 +17,21 @@ public void sanity(){
...
-      .countryCodeAlpha2("US")
-      .countryCodeAlpha3("USA")
-      .build();
+    Incident incident = getFilledIncident();
     byte[] serialized = TestUtils.serialize(incident);
     assertEquals(incident, deserialize(serialized, Incident.class));
   }
 
+  @Test
+  public void testToAndFromJson() {
```
Guardiola31337, line 26: We should also remove `test` prefix from tests names, it's implicit with the usage of `@Test` annotation.

Same thing applies to other tests above and below.

Guardiola31337 (commented):
Left a couple of minor comments.

services-directions-models/src/test/java/com/mapbox/api/directions/v5/models/IncidentTest.java
VysotskiVadim: I don't see anything bad in having a `test` word in the beginning in a test name. But if you have strong opinion about this, I don't mind changing the names.
Updated.

RingerJK:
please extend `RouteLegRefreshTest` and add  `affected_road_names` to `directions_refresh_v1.json`, `directions_refresh_v1_obj.json`,  `directions_refresh_v1_obj_passed_first_leg.json` where it needs

RingerJK (approved):

PR #1463, opened 2022-07-21
correctly assign MapMatchingMatching#routeIndex

LukasPaczos:
Fixes an issues where the `MapMatchingMathcing#routeIndex` wasn't correctly assigned which could've resulted in exceptions when used together with Nav SDK compatibility APIs, for example:
```
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: com.mapbox.navigation.examples, PID: 13969
    java.lang.IllegalArgumentException: Provided DirectionsRoute has to have #routeIndex property set.
    If the route was generated independently of Nav SDK,
    rebuild the object and assign the index based on the position in the response collection.
        at com.mapbox.navigation.base.route.NavigationRouteEx.toNavigationRoute(NavigationRoute.kt:463)
        at com.mapbox.navigation.base.route.NavigationRouteEx.toNavigationRoute(NavigationRoute.kt:443)
        at com.mapbox.navigation.base.route.NavigationRouteEx.toNavigationRoutes(NavigationRoute.kt:370)
        at com.mapbox.navigation.core.MapboxNavigation.setRoutes(MapboxNavigation.kt:775)
        at com.mapbox.navigation.core.MapboxNavigation.setRoutes$default(MapboxNavigation.kt:774)
```


Guardiola31337 (approved):
Great catch 🚀

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Mapbox welcomes participation and contributions from everyone.
 
 ### main
 - Added `Incident.affectedRoadNames`. [#1457](https://github.com/mapbox/mapbox-java/pull/1457)
+- Fixes an issues where the `MapMatchingMathcing#routeIndex` wasn't correctly assigned. [#1463](https://github.com/mapbox/mapbox-java/pull/1463)
```
dzinad, line 7: an issue?

services-matching/src/test/java/com/mapbox/api/matching/v5/MapboxMapMatchingTest.java
```diff
@@ -224,6 +224,19 @@ public void mapMatchingToDirectionsRoute() throws Exception {
...
+  @Test
+  public void toDirectionsRoute_routeIndex() throws Exception {
+    MapboxMapMatching mapMatching = MapboxMapMatching.builder()
+      .profile(PROFILE_DRIVING)
+      .coordinates(coordinates)
+      .baseUrl(mockUrl.toString())
+      .accessToken(ACCESS_TOKEN)
+      .build();
+    assertEquals(
+      "0",
```
dzinad, line 236: Don't you also want to test when it's non zero? I imagine the situation when such a test would have caught a bug: instead of using the for with i you leave forEach, introduce i and forget to increment it:
```
int i = 0;
for (MapMatchingMatching matching : matchings) {
   ...routeIndex(String.valueOf(i))
}
```

dzinad (approved):

PR #1469, opened 2022-07-28
expose snapping_include_static_closures

dzinad:


samples/src/main/java/com/mapbox/samples/test.json
```diff
@@ -0,0 +1,122 @@
+{
```
VysotskiVadim, line 1: how is this file used? Can't find any usages 🤔

dzinad: Whoops.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1858,6 +1897,51 @@ public Builder snappingIncludeClosuresList(@Nullable List<Boolean> snappingClosu
...
+    @NonNull
+    public Builder snappingIncludeStaticClosuresList(
+            @Nullable List<Boolean> snappingStaticClosures
+    ) {
+      String result = FormatUtils.join(";", snappingStaticClosures);
+      if (result != null) {
+        snappingIncludeStaticClosures(result);
+      }
+      return this;
+    }
```
VysotskiVadim, line 1943: What does happen if you reset parameters from existing route options? 
```
RouteOptions.builder().
    // all all required parameters
    .build()
    .toBuilder()
    .snappingIncludeStaticClosuresList(null)
    .build()
```
or
```
RouteOptions.builder().
    // all all required parameters
    .build()
    .toBuilder()
    .snappingIncludeStaticClosuresList(emptyList())
    .build()
```
?
Do you think it makes sense to add tests for this?

VysotskiVadim (approved):
LGTM 👍
Please check the comment if you get a chance

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
```diff
@@ -1858,6 +1897,51 @@ public Builder snappingIncludeClosuresList(@Nullable List<Boolean> snappingClosu
...
+     *  However, you can skip a coordinate and show its position in the list with the null.
+     *  If null, this parameter defaults to false.
+     * <p>
+     * Only available with the {@link DirectionsCriteria#PROFILE_DRIVING_TRAFFIC}.
+     *
+     * @param snappingStaticClosures a list of booleans
+     * @return this builder for chaining options together
+     */
+    @NonNull
+    public Builder snappingIncludeStaticClosuresList(
```
RingerJK, line 1935: could you also add `snappingIncludeStaticClosuresList(null)` test

RingerJK (commented):
LGTM with 1 test

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
dzinad: I added tests for null and empty list but without previous `toBuilder` invocation. I also added tests that nothing breaks after `toBuilder` + `build` (is this redundant?). I'm not sure tho why it would make sense to check resetting.

VysotskiVadim: resetting from a value to null looks suspicious in the code. It seems that in case of `null` `snappingIncludeStaticClosures` won't be called from `snappingIncludeStaticClosuresList ` and state won't change. As I see, coordinates work in the same way, so I'm not sure it's a bug. I think it would be nice to document that behaviour in the test. WDYT?

dzinad: Great catch! I'd say it's a bug. @LukasPaczos added this code, WDYT?

LukasPaczos: That does look like a bug.

LukasPaczos:
Should this feature receive the same treatment [as `snapping_include_closures`](https://github.com/mapbox/mapbox-navigation-android/issues/5664)? Should we append it by default for re-routes' origin? I would imagine so but wanted to double check @mandeepsandhu @mskurydin.

mandeepsandhu:
> Should this feature receive the same treatment [as `snapping_include_closures`](https://github.com/mapbox/mapbox-navigation-android/issues/5664)? Should we append it by default for re-routes' origin? I would imagine so but wanted to double check @mandeepsandhu @mskurydin.

I believe so, yes. @nuttert can you confirm?

LukasPaczos:
Tracking in https://github.com/mapbox/mapbox-navigation-android/issues/6119.

LukasPaczos (approved):
LGTM

PR #1458, opened 2022-07-06
add toll cost parameter

dzinad:

PR #1472, opened 2022-08-10
Add `SuppressWarnings` annotation that would allow for adding type safe bindings without documentation.

abhishek1508:
Add `SuppressWarnings` annotation that would allow for adding type safe bindings without documentation.

Usage of annotations:

For suppressing `method` documentation
```
@SuppressWarnings("checkstyle:javadocmethod")
@NonNull
public abstract Builder guideMap(@Nullable String guideMap);
```

For suppressing `variable` documentation
```
@SuppressWarnings("checkstyle:javadocvariable")
public static final String SAPAGUIDEMAP = "sapaguidemap";
```

cc @mapbox/navigation-android

### DIFF
--- a/config/checkstyle/checkstyle.xml
+++ b/config/checkstyle/checkstyle.xml
@@ -45,12 +45,18 @@
   <!--################ Non-AST checks ################-->
   <!--################################################-->
 
+  <module name="SuppressWarningsFilter" />
+
   <module name="TreeWalker">
 
     <!--#################################################################-->
     <!--################ NAMING CHECKS ##################################-->
     <!--#################################################################-->
 
+    <module name="SuppressWarningsHolder" />
+    <module name="SuppressWarnings">
+      <property name="id" value="checkstyle:suppresswarnings"/>
+    </module>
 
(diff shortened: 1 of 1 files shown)


LukasPaczos (approved):

dzinad:
Thanks! What about classes? They can also go undocumented. For example: https://github.com/mapbox/mapbox-java/pull/1458/files#diff-d5e8ea08168649a9383db3b5daa40f71bf5d6428df4b008a0a165f78af40f188

LukasPaczos:
Let's add that as well @dzinad :+1:

abhishek1508:
> Thanks! What about classes? They can also go undocumented. For example: https://github.com/mapbox/mapbox-java/pull/1458/files#diff-d5e8ea08168649a9383db3b5daa40f71bf5d6428df4b008a0a165f78af40f188

Yep let's do it. I only added the ones I needed for SAPA and set the infrastructure, so now we know what needs to be done for others.

dzinad:
Added in https://github.com/mapbox/mapbox-java/pull/1458.

PR #1486, opened 2022-08-31
introduce TurfSimplify

### LINKED ISSUE #1205: Implement turf-simplify
This library already contains a port of simplify-js: https://github.com/hgoebl/simplify-java

dzinad:
Closes #1205.
We could add a method that simplifies `LineString` instead of `List<Point>` (probably in turf module as well), but I'm not sure if it's needed for the first iteration. The example of how to simplify `LineString` is present in `BasicTurfSimplify`. Let me know what you think.


LukasPaczos (approved):
🚢

LukasPaczos:
@dzinad I just found out that `mapbox-java` already offered simplify-js port in https://github.com/mapbox/mapbox-java/blob/v6.8.0-beta.4/services-geojson/src/main/java/com/mapbox/geojson/utils/PolylineUtils.java (added in https://github.com/mapbox/mapbox-java/commit/ed4a1ff888a706c7e049ef29950a17eefeeec3e6). I don't know why it was added to a `Polyline` class in a `geojson` module of all places but looks like no-one was aware of its existence.

We should join the test cases as needed, and deprecate the `PolylineUtils#simplify`. Before the changes go stable we should also move the functions from `TurfSimplify` to `TurfTransformation` (to align with https://turfjs.org/docs) and update https://github.com/mapbox/mapbox-java/blob/main/docs/turf-port.md.

Thanks for the work on this one!

PR #1500, opened 2022-10-19
Support unrecognized properties for EV features

dzinad:


CHANGELOG.md
```diff
@@ -3,6 +3,9 @@
 Mapbox welcomes participation and contributions from everyone.
 
 ### main
+- Added `unrecognized()` to `DirectionsRefreshJsonObject` so that unrecognized properties can be received from refresh response.
+- Added `unrecognized()` to `DirectionsRefreshJsonObject` so that unrecognized properties can be received from refresh response. [#1500](https://github.com/mapbox/mapbox-java/pull/1500)
```
LukasPaczos, line 7: Duplicate entries.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
```diff
@@ -219,6 +223,28 @@ public abstract static class Builder extends DirectionsJsonObject.Builder<Builde
...
 
+    /**
+     * Use this method to add parameters which are not present in the model yet but are supported
+     * on the Directions API side in the response.
+     * Use it for experimental parameters.
+     *
+     * @param unrecognizedProperties parameters to add to request
+     */
+    @NonNull
+    public Builder unrecognizedProperties(
```
LukasPaczos, line 234: Why not make the abstract `unrecognized` method public in `DirectionsJsonObject.Builder`? We definitely see a use-case for this method being public. Or add `unrecognizedProperties` as public function to the abstract builder if we don't want to expose the `SerializableJsonElement`.

cc @VysotskiVadim

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/UnrecognizedPropertiesUtils.java
```diff
@@ -0,0 +1,57 @@
...
+import com.mapbox.auto.value.gson.SerializableJsonElement;
+
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
+
+/**
+ * Provides utility methods to work with unrecognized properties.
+ */
+public final class UnrecognizedPropertiesUtils {
```
LukasPaczos, line 15: Should this be public? `DirectionsJsonObject` already exposes these accessors.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
```diff
@@ -25,4 +34,51 @@ public String toJson() {
...
+    return UnrecognizedPropertiesUtils.getUnrecognizedProperty(propertyName, unrecognized());
+  }
+
+  @Nullable
+  @UnrecognizedJsonProperties
+  abstract Map<String, SerializableJsonElement> unrecognized();
+
+  abstract static class Builder<T extends Builder> {
+    @NonNull
+    abstract T unrecognized(@Nullable Map<String, SerializableJsonElement> value);
```
LukasPaczos, line 82: Same as above, why not making this public?

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
```diff
@@ -25,4 +34,51 @@ public String toJson() {
...
+   * @return value of a requested property or null if the requested property doesn't exist.
+   */
+  @Nullable
+  public final JsonElement getUnrecognizedProperty(String propertyName) {
+    return UnrecognizedPropertiesUtils.getUnrecognizedProperty(propertyName, unrecognized());
+  }
+
+  @Nullable
+  @UnrecognizedJsonProperties
+  abstract Map<String, SerializableJsonElement> unrecognized();
```
LukasPaczos, line 78: We could consider making this public, refs `NAVAND-778`. Doesn't have to happen in this PR though.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
```diff
@@ -219,6 +223,28 @@ public abstract static class Builder extends DirectionsJsonObject.Builder<Builde
...
+      @Nullable Map<String, JsonElement> unrecognizedProperties
```
LukasPaczos, line 235: `RouteOptions.Builder` exposes this map as `Map<String, String>`, we should align this one way or the other. For the general purpose `JsonElement` does look like a better candidate, but again only if we don't want to expose the `SerializableJsonElement`. @VysotskiVadim I remember you had some thoughts about it in the past.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
VysotskiVadim, line 82:
```suggestion
    internal abstract T unrecognized(@Nullable Map<String, SerializableJsonElement> value);
```
Let's keep it hidden from the end user

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
```diff
@@ -1,16 +1,25 @@
...
 import java.io.Serializable;
+import java.util.Collections;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * Provides a base class for Directions model classes.
  */
-public class DirectionsRefreshJsonObject implements Serializable {
+public abstract class DirectionsRefreshJsonObject implements Serializable {
```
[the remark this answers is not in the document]
VysotskiVadim: even if it's a breaking change we can consider the previous implementation as a bug. I believe we already did it once 🙂

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
VysotskiVadim, line 234:
```suggestion
    internal Builder unrecognizedProperties(
```
I believe users are not supposed to use it, aren't they?

dzinad: 1) It wasn't public so I thought there was a reason for it. So we should only make it public when needed, a white list, not black list;
2) The `unrecognized()` method accepts `Map<String, SerializableJsonElement>` which is not very convenient for the end user, so yes, I'd prefer adding `unrecognizedProperties` with `Map<String, JsonElement>` instead.

So is there a reason why it's not public?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/UnrecognizedPropertiesUtils.java
dzinad: I want to access it from the `service-directions-refresh-models` module. To avoid code duplicating.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
dzinad: How will making this public help `NAVAND-778`? For the case described there it's more like adding a method `addUnrecognizedProperties` instead of the current one that resets them.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
dzinad: Yes, here we definitely need `JsonElement`. The structure will be complex.
For `RouteOptions` it's only about get parameters, that's why strings are used.
These are different use cases so I'm not sure if we should complicate `RouteOptions` usage with converting it all to `JsonElement`. But we could have 2 methods for `RouteOptions` to keep both consistency and usability.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
dzinad: Absent modifier means package-private here. I did it like this to keep it consistent with `DirectionsJsonObject`.
But the question is why it should be hidden at all.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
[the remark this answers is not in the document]
dzinad: > TBH I don't remember which visibility default constructor from java has

It's public with no arguments.

> it's a breaking change, isn't it?

You are right. Good catch! I see the following options here:
1) Commit the breaking change and reduce the constructor visibility for the future. Why would anyone create an empty `DirectionsRefreshJsonObject`?.. But that would still break semver;
2) Up the major version. I'm not sure about the policy on that. @LukasPaczos could you chime in here?
3) Add unrecognized properties directly to `DirectionsRouteRefresh`. That's the only place they are needed right now. But it's not very flexible;
4) Add compile-time support for EV features. Not really consistent with the feature being in beta;

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
VysotskiVadim: >  @VysotskiVadim I remember you had some thoughts about it in the past.

the initial idea was to expose as little as possible to simplify future changes. 

> For RouteOptions it's only about get parameters, that's why strings are used.

that's right.

> These are different use cases so I'm not sure if we should complicate RouteOptions usage with converting it all to JsonElement. 

I agree with @dzinad. I don't see value in aligning unrecognised properties API between request and response models.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
VysotskiVadim: it's a fair point

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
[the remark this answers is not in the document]
VysotskiVadim: Actually I had the same problem adding unrecognized properties: https://github.com/mapbox/mapbox-java/pull/1394/files#r848311760 😄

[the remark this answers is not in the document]
dzinad: Haha, nice
Then maybe we should think about reducing the visibility of everything that should not be used directly? These are mostly constructors but it makes sense to double-check. That's a separate issue though.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
LukasPaczos: I don't fully agree. Whether we look at request or response object, they are all `DirectionsJsonObject`s (or refresh) and support the unrecognized fields in exactly the same way, and that's why I think they should also expose the same API to mutate or reset the properties. 

We started with `RouteOptions` needing some form of unrecognized properties setter, now we need to add the same to `LegAnnotation` and this could be true at any point in time for any `DirectionsJsonObject`, depending on what's `experimental` in the Directions API at a certain time.

Couldn't we expose `Map<String, JsonElement>` getter and setters for all `DirectionsJsonObject`s?

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
LukasPaczos: A public getter would allow us to first get the properties map, mutate them, and set them back. As we would do with any other builder object parameter. The getter and setter methods in this case need to work on the same `map` type so that developers don't need to do any parsing themselves, that's why I was hoping to expose a single function in the abstract class instead of adding explicit getter and setters manually everywhere.

dzinad: OK, makes sense but it's not very convenient to use with `SerializableJsonElement`. But I guess we could add convenience methods directly to `DirectionsJsonObject` and `DirectionsRefreshJsonObject`. WDYT?

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
dzinad: We can if we leave the convenience method for `RouteOptions`.

LukasPaczos: Why couldn't we use only `Map<String, JsonElement>` for `RouteOptions` as well?

The only change developers would have to make is wrap their `string` values in `JsonPrimitive(value)`. That's actually even more flexible because other than `string` parameters are also acceptable by the Directions API. Of course we can leave the existing function in place but I'd rather deprecate it.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
LukasPaczos: Sure, as we're discussing in https://github.com/mapbox/mapbox-java/pull/1500#discussion_r1000562709 `JsonElement` might be better.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/LegAnnotation.java
dzinad: 1) It's a breaking change and that's something that would definitely affect the users;
2) In the majority of the cases users only need `Map<String, String>`. Making them use `Map<String, JsonElement>` will force them to write more boiler-plate code on their side. 

I'm strongly in favour of leaving the convenience method as-is. We'll provide a more flexible options: if they want something other than strings, they can use the new method. And I'm not suggesting to add `Map<String, String>` everywhere: only leave it where it already is and where it makes sense to have it.

LukasPaczos: > It's a breaking change and that's something that would definitely affect the users;

We'd not be removing the old function, only deprecating it.

> In the majority of the cases users only need Map<String, String>. Making them use Map<String, JsonElement> will force them to write more boiler-plate code on their side.

`Map<String, String>` is limiting and actually somewhat incorrect. For example, look at `ev_max_charge` parameter in https://docs.mapbox.com/api/navigation/directions/#parameters-for-electric-vehicle-routing - it's an integer but we'd recommend (and today require) developers pass it as a `string`. `JsonPrimitive(integer)` would be more accurate here.

In any case, we can discuss and consider the deprecation separately, the old function has to stay in place regardless.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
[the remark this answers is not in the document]
LukasPaczos: Let's make the change as last time around :+1: 

> Then maybe we should think about reducing the visibility of everything that should not be used directly?

Good ticket to have for the next major version :+1:

services-directions-models/src/main/java/com/mapbox/api/directions/v5/utils/UnrecognizedPropertiesUtils.java
LukasPaczos: I imagine we don't need this if we expose the map getters.

services-directions-refresh-models/src/main/java/com/mapbox/api/directionsrefresh/v1/models/DirectionsRefreshJsonObject.java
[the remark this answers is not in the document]
dzinad: `NAVAND-802`

LukasPaczos (approved):
LGTM

PR #1509, opened 2022-11-09
Return back waypoints per route

korshaknn:
return back #1503


services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsResponse.java
```diff
@@ -66,10 +66,16 @@ public static Builder builder() {
...
+   * @deprecated the waypoints are returned in the root of the response only if
+   *   {@link RouteOptions#waypointsPerRoute()} is false. Otherwise they are returned in
+   *   {@link DirectionsRoute#waypoints()}. You can continue to use this and it will work as-is.
+   *   However, it's recommended to use the new approach to distinguish waypoints between routes
+   *   (for example, this is necessary when asking for an EV-optimized route with alternatives).
    *
    * @return list of {@link DirectionsWaypoint} objects ordered from start of route till the end
    * @since 1.0.0
    */
+  @Deprecated
```
LukasPaczos, line 78: The API will actually only be available in 2 weeks or so, so let's skip the deprecation notice for now.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsResponse.java
```diff
@@ -263,12 +269,18 @@ public abstract static class Builder extends DirectionsJsonObject.Builder<Builde
...
+     *   {@link DirectionsRoute#waypoints()}. You can continue to use this and it will work as-is.
+     *   However, it's recommended to use the new approach to distinguish waypoints between routes
+     *   (for example, this is necessary when asking for an EV-optimized route with alternatives).
      *
      * @param waypoints list of {@link DirectionsWaypoint} objects ordered from start of route till
      *                  the end
      * @return this builder for chaining options together
      * @since 3.0.0
      */
+    @Deprecated
```
LukasPaczos, line 283: Same as above.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsResponse.java
korshaknn: should i also remove a docs part?
```
 @deprecated the waypoints are returned in the root of the response only if
   *   {@link RouteOptions#waypointsPerRoute()} is false. Otherwise they are returned in
   *   {@link DirectionsRoute#waypoints()}. You can continue to use this and it will work as-is.
   *   However, it's recommended to use the new approach to distinguish waypoints between routes
   *   (for example, this is necessary when asking for an EV-optimized route with alternatives).
```

LukasPaczos: Yes, we can re-add it later.

korshaknn:
@LukasPaczos ptal

LukasPaczos (approved):

services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/DirectionsResponse.java
```diff
@@ -66,16 +66,10 @@ public static Builder builder() {
    * List of {@link DirectionsWaypoint} objects. Each {@code waypoint} is an input coordinate
    * snapped to the road and path network. The {@code waypoint} appear in the list in the order of
    * the input coordinates.
-   * @deprecated the waypoints are returned in the root of the response only if
```
dzinad: Why did you remove `@Deprecated`? It's deprecated on backend.

LukasPaczos: Context in https://github.com/mapbox/mapbox-java/pull/1509#discussion_r1017895480.

korshaknn: https://github.com/mapbox/mapbox-java/pull/1509#discussion_r1017895480

LukasPaczos: Nav API team mentioned earlier today that these changes are not yet available. Unless the situation is different @dzinad?

dzinad: When did they mention that?
It's working for me. See
`https://api.mapbox.com/directions/v5/mapbox/driving-traffic/138.4282743,34.9665031;138.42594,34.9665461?access_token=<TOKEN>&geometries=polyline6&overview=full&steps=true&compute_toll_cost=true`
vs 
`https://api.mapbox.com/directions/v5/mapbox/driving-traffic/138.4282743,34.9665031;138.42594,34.9665461?access_token=<TOKEN>&geometries=polyline6&overview=full&steps=true&compute_toll_cost=true&waypoints_per_route=true`.

Or do you mean EV waypoints? Then probably yes, it's still in development but this parameter is not about EV. Anyways, I'm not against not deprecating it at all.

PR #1522, opened 2022-12-01
Routable points support for services-geocoding

DzmitryFomchyn:
Add [routable points](https://docs.mapbox.com/api/search/geocoding/) support for `services-geocoding`

### DIFF
--- a/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/MapboxGeocoding.java
+++ b/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/MapboxGeocoding.java
@@ -97,7 +97,8 @@ public abstract class MapboxGeocoding extends MapboxService<GeocodingResponse, G
       limit(),
       languages(),
       reverseMode(),
-      fuzzyMatch());
+      fuzzyMatch(),
+      routing());
   }
 
   private Call<List<GeocodingResponse>> getBatchCall() {
@@ -214,6 +215,9 @@ public abstract class MapboxGeocoding extends MapboxService<GeocodingResponse, G
 
   @Nullable
   abstract Boolean fuzzyMatch();
+
+  @Nullable
+  abstract Boolean routing();
 
   @Nullable
   abstract String clientAppName();
@@ -559,6 +563,24 @@ public abstract class MapboxGeocoding extends MapboxService<GeocodingResponse, G
      * @since 4.9.0
      */
     public abstract Builder fuzzyMatch(Boolean fuzzyMatch);
+
+    /**
+     * Specify whether to request additional metadata about the recommended navigation destination
+     * corresponding to the feature (true) or not (false, default).
+     * Only applicable for address features.
+     *
+     * For example, if routing=true the response could include data about a point
+     * on the road the feature fronts.
+     * Response features may include an array containing one or more routable points.
+     * Routable points cannot always be determined. Consuming applications should fall back to
+     * using the feature's normal geometry for routing if a separate routable point
+     * is not returned.
+     *
+     * @param routing optionally set whether to request
+     *                additional metadata about the recommended navigation destination
+     * @return this builder for chaining options together
+     */
+    public abstract Builder routing(Boolean routing);
 
     /**
      * Required to call when this is being built. If no access token provided,
--- a/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/CarmenFeature.java
+++ b/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/CarmenFeature.java
@@ -262,6 +262,15 @@ public abstract class CarmenFeature implements GeoJson {
   public abstract String matchingPlaceName();
 
   /**
+   * An object with the routable points for the {@link CarmenFeature}.
+   *
+   * @return an object with the routable points for the {@link CarmenFeature}
+   */
+  @Nullable
+  @SerializedName("routable_points")
+  public abstract RoutablePoints routablePoints();
+
+  /**
    * A string of the IETF language tag of the query's primary language.
    *
    * @return string containing the query's primary language
@@ -481,6 +490,14 @@ public abstract class CarmenFeature implements GeoJson {
      * @since 2.2.0
      */
     public abstract Builder language(@Nullable String language);
+
+    /**
+     * An object with the routable points for the {@link CarmenFeature}.
+     *
+     * @param routablePoints an object with the routable points for the {@link CarmenFeature}
+     * @return this builder for chaining options together
+     */
+    public abstract Builder routablePoints(@Nullable RoutablePoints routablePoints);
 
     /**
      * Build a new {@link CarmenFeature} object.
--- a/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
+++ b/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
@@ -1,0 +1,95 @@
+package com.mapbox.api.geocoding.v5.models;
+
+import androidx.annotation.Nullable;
+
+import com.google.auto.value.AutoValue;
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.annotations.SerializedName;
+import com.mapbox.geojson.Point;
+
+/**
+ * Object representing {@link CarmenFeature}'s routable point.
+ */
+@AutoValue
+public abstract class RoutablePoint {
+
+  /**
+   * A string representing the routable point name.
+   *
+   * @return routable point name
+   */
+  @Nullable
+  @SerializedName("name")
+  public abstract String name();
+
+  /**
+   * A {@link Point} object which represents the routable point location.
+   *
+   * @return a GeoJson {@link Point} which defines the routable point location
+   */
+  @Nullable
+  public Point coordinate() {
+    final double[] coordinate = rawCoordinate();
+    if (coordinate != null && coordinate.length == 2) {
+      return Point.fromLngLat(coordinate[0], coordinate[1]);
+    }
+    return null;
+  }
+
+  // No public access thus, we lessen enforcement on mutability here.
+  @Nullable
+  @SerializedName("coordinates")
+  @SuppressWarnings("mutable")
+  abstract double[] rawCoordinate();
+
+  /**
+   * Convert current instance values into another Builder to quickly change one or more values.
+   *
+   * @return a new instance of {@link Builder}
+   */
+  @SuppressWarnings("unused")
+  public abstract Builder toBuilder();
+
+  /**
+   * Gson type adapter for parsing Gson to this class.
+   *
+   * @param gson the built {@link Gson} object
+   * @return the type adapter for this class
+   */
+  public static TypeAdapter<RoutablePoint> typeAdapter(Gson gson) {
+    return new AutoValue_RoutablePoint.GsonTypeAdapter(gson);
+  }
+
+  /**
+   * This builder can be used to set the values describing the {@link RoutablePoint}.
+   */
+  @AutoValue.Builder
+  @SuppressWarnings("unused")
+  public abstract static class Builder {
+
+    /**
+     * A string representing the routable point name.
+     *
+     * @param name routable point name
+     * @return this builder for chaining options together
+     */
+    public abstract Builder name(@Nullable String name);
+
+    /**
+     * Raw coordinates (longitude and latitude, order matters)
+     * that represent the routable point location.
+     *
+     * @param coordinate raw coordinates that represent the routable point location
+     * @return this builder for chaining options together
+     */
+    public abstract Builder rawCoordinate(@Nullable double[] coordinate);
+
+    /**
+     * Build a new {@link RoutablePoint} object.
+     *
+     * @return a new {@link RoutablePoint} using the provided values in this builder
+     */
+    public abstract RoutablePoint build();
+  }
+}
--- a/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
+++ b/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
@@ -1,0 +1,70 @@
+package com.mapbox.api.geocoding.v5.models;
+
+import androidx.annotation.Nullable;
+
+import com.google.auto.value.AutoValue;
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.annotations.SerializedName;
+
+import java.util.List;
+
+/**
+ * An object with the routable points for the {@link CarmenFeature}.
+ */
+@AutoValue
+public abstract class RoutablePoints {
+
+  /**
+   * A list of routable points for the {@link CarmenFeature}, or null if no points were found.
+   *
+   * @return a list of routable points for the {@link CarmenFeature},
+   *               or null if no points were found
+   */
+  @Nullable
+  @SerializedName("points")
+  public abstract List<RoutablePoint> points();
+
+  /**
+   * Convert current instance values into another Builder to quickly change one or more values.
+   *
+   * @return a new instance of {@link Builder}
+   */
+  @SuppressWarnings("unused")
+  public abstract Builder toBuilder();
+
+  /**
+   * Gson type adapter for parsing Gson to this class.
+   *
+   * @param gson the built {@link Gson} object
+   * @return the type adapter for this class
+   */
+  public static TypeAdapter<RoutablePoints> typeAdapter(Gson gson) {
+    return new AutoValue_RoutablePoints.GsonTypeAdapter(gson);
+  }
+
+  /**
+   * This builder can be used to set the values describing the {@link RoutablePoints}.
+   */
+  @AutoValue.Builder
+  @SuppressWarnings("unused")
+  public abstract static class Builder {
+
+    /**
+     * A list of routable points for the {@link CarmenFeature},
+     * or null if no points were found.
+     *
+     * @param points a list of routable points for the {@link CarmenFeature},
+     *               or null if no points were found
+     * @return this builder for chaining options together
+     */
+    public abstract Builder points(@Nullable List<RoutablePoint> points);
+
+    /**
+     * Build a new {@link RoutablePoints} object.
+     *
+     * @return a new {@link RoutablePoints} using the provided values in this builder
+     */
+    public abstract RoutablePoints build();
+  }
+}
--- a/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/GeocodingService.java
+++ b/services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/GeocodingService.java
@@ -35,6 +35,8 @@ public interface GeocodingService {
    * @param language     The locale in which results should be returned.
    * @param reverseMode  Set the factors that are used to sort nearby results.
    * @param fuzzyMatch   Set whether to allow the geocoding API to attempt exact matching or not.
+   * @param routing      Set whether to request additional metadata
+   *                     about the recommended navigation destination.
    * @return A retrofit Call object
    * @since 1.0.0
    */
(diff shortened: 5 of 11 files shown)


services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/MapboxGeocoding.java
```diff
@@ -560,6 +564,25 @@ public abstract Builder reverseMode(
...
+     * on the road the feature fronts.
+     * Response features may include an array containing one or more routable points.
+     * Routable points cannot always be determined. Consuming applications should fall back to
+     * using the feature's normal geometry for routing if a separate routable point
+     * is not returned.
+     *
+     * @param routing optionally set whether to request
+     *                additional metadata about the recommended navigation destination
+     * @return this builder for chaining options together
+     * @since 6.10.0
```
Guardiola31337, line 582:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/CarmenFeature.java
```diff
@@ -261,6 +261,16 @@ public Point center() {
   @SerializedName("matching_place_name")
   public abstract String matchingPlaceName();
 
+  /**
+   * An object with the routable points for the {@link CarmenFeature}.
+   *
+   * @return an object with the routable points for the {@link CarmenFeature}
+   * @since 6.10.0
```
Guardiola31337, line 268:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/CarmenFeature.java
```diff
@@ -482,6 +492,15 @@ public abstract static class Builder {
      */
     public abstract Builder language(@Nullable String language);
 
+    /**
+     * An object with the routable points for the {@link CarmenFeature}.
+     *
+     * @param routablePoints an object with the routable points for the {@link CarmenFeature}
+     * @return this builder for chaining options together
+     * @since 6.10.0
```
Guardiola31337, line 500:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+import com.google.auto.value.AutoValue;
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.annotations.SerializedName;
+import com.mapbox.geojson.Point;
+
+/**
+ * Object representing {@link CarmenFeature}'s routable point.
+ *
+ * @since 6.10.0
```
Guardiola31337, line 14:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+ */
+@AutoValue
+public abstract class RoutablePoint {
+
+  /**
+   * A string representing the routable point name.
+   *
+   * @return routable point name
+   * @since 6.10.0
```
Guardiola31337, line 23:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+  @Nullable
+  @SerializedName("coordinates")
+  @SuppressWarnings("mutable")
+  abstract double[] rawCoordinate();
+
+  /**
+   * Convert current instance values into another Builder to quickly change one or more values.
+   *
+   * @return a new instance of {@link Builder}
+   * @since 6.10.0
```
Guardiola31337, line 54:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  @Nullable
+  @SerializedName("name")
+  public abstract String name();
+
+  /**
+   * A {@link Point} object which represents the routable point location.
+   *
+   * @return a GeoJson {@link Point} which defines the routable point location
+   * @since 6.10.0
```
Guardiola31337, line 33:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  @SuppressWarnings("unused")
+  public abstract Builder toBuilder();
+
+  /**
+   * Gson type adapter for parsing Gson to this class.
+   *
+   * @param gson the built {@link Gson} object
+   * @return the type adapter for this class
+   * @since 6.10.0
```
Guardiola31337, line 64:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+   */
+  public static TypeAdapter<RoutablePoint> typeAdapter(Gson gson) {
+    return new AutoValue_RoutablePoint.GsonTypeAdapter(gson);
+  }
+
+  /**
+   * This builder can be used to set the values describing the {@link RoutablePoint}.
+   *
+   * @since 6.10.0
```
Guardiola31337, line 73:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+  @AutoValue.Builder
+  @SuppressWarnings("unused")
+  public abstract static class Builder {
+
+    /**
+     * A string representing the routable point name.
+     *
+     * @param name routable point name
+     * @return this builder for chaining options together
+     * @since 6.10.0
```
Guardiola31337, line 84:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public abstract Builder name(@Nullable String name);
+
+    /**
+     * Raw coordinates (longitude and latitude, order matters)
+     * that represent the routable point location.
+     *
+     * @param coordinate raw coordinates that represent the routable point location
+     * @return this builder for chaining options together
+     * @since 6.10.0
```
Guardiola31337, line 94:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+     */
+    public abstract Builder rawCoordinate(@Nullable double[] coordinate);
+
+    /**
+     * Build a new {@link RoutablePoint} object.
+     *
+     * @return a new {@link RoutablePoint} using the provided values in this builder
+     * @since 6.10.0
```
Guardiola31337, line 102:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
```diff
@@ -0,0 +1,79 @@
...
+import com.google.gson.Gson;
+import com.google.gson.TypeAdapter;
+import com.google.gson.annotations.SerializedName;
+
+import java.util.List;
+
+/**
+ * An object with the routable points for the {@link CarmenFeature}.
+ *
+ * @since 6.10.0
```
Guardiola31337, line 15:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
```diff
@@ -0,0 +1,79 @@
...
+ */
+@AutoValue
+public abstract class RoutablePoints {
+
+  /**
+   * A list of routable points for the {@link CarmenFeature}, or null if no points were found.
+   *
+   * @return a list of routable points for the {@link CarmenFeature},
+   *               or null if no points were found
+   * @since 6.10.0
```
Guardiola31337, line 25:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  @Nullable
+  @SerializedName("points")
+  public abstract List<RoutablePoint> points();
+
+  /**
+   * Convert current instance values into another Builder to quickly change one or more values.
+   *
+   * @return a new instance of {@link Builder}
+   * @since 6.10.0
```
Guardiola31337, line 35:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  @SuppressWarnings("unused")
+  public abstract Builder toBuilder();
+
+  /**
+   * Gson type adapter for parsing Gson to this class.
+   *
+   * @param gson the built {@link Gson} object
+   * @return the type adapter for this class
+   * @since 6.10.0
```
Guardiola31337, line 45:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
```diff
@@ -0,0 +1,79 @@
...
+   */
+  public static TypeAdapter<RoutablePoints> typeAdapter(Gson gson) {
+    return new AutoValue_RoutablePoints.GsonTypeAdapter(gson);
+  }
+
+  /**
+   * This builder can be used to set the values describing the {@link RoutablePoints}.
+   *
+   * @since 6.10.0
```
Guardiola31337, line 54:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
```diff
@@ -0,0 +1,79 @@
...
+  public abstract static class Builder {
+
+    /**
+     * A list of routable points for the {@link CarmenFeature},
+     * or null if no points were found.
+     *
+     * @param points a list of routable points for the {@link CarmenFeature},
+     *               or null if no points were found
+     * @return this builder for chaining options together
+     * @since 6.10.0
```
Guardiola31337, line 67:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoints.java
```diff
@@ -0,0 +1,79 @@
...
+     */
+    public abstract Builder points(@Nullable List<RoutablePoint> points);
+
+    /**
+     * Build a new {@link RoutablePoints} object.
+     *
+     * @return a new {@link RoutablePoints} using the provided values in this builder
+     * @since 6.10.0
```
Guardiola31337, line 75:
```suggestion
```

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+@AutoValue
+public abstract class RoutablePoint {
+
+  /**
+   * A string representing the routable point name.
+   *
+   * @return routable point name
+   * @since 6.10.0
+   */
+  @Nullable
```
Guardiola31337, line 25: Is this really nullable? I mean, what I'd expect is `RoutablePoints` to be nullable but If they're not, any `RoutablePoint` from the list to have valid values. Does that make sense?

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+  @SerializedName("name")
+  public abstract String name();
+
+  /**
+   * A {@link Point} object which represents the routable point location.
+   *
+   * @return a GeoJson {@link Point} which defines the routable point location
+   * @since 6.10.0
+   */
+  @Nullable
```
Guardiola31337, line 35: Same as above re: Is this really nullable?

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+  public Point coordinate() {
+    final double[] coordinate = rawCoordinate();
+    if (coordinate != null && coordinate.length == 2) {
+      return Point.fromLngLat(coordinate[0], coordinate[1]);
+    }
+    return null;
+  }
+
+  // No public access thus, we lessen enforcement on mutability here.
+  @Nullable
```
Guardiola31337, line 45: Same as above re: Is this really nullable?

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+  public abstract static class Builder {
+
+    /**
+     * A string representing the routable point name.
+     *
+     * @param name routable point name
+     * @return this builder for chaining options together
+     * @since 6.10.0
+     */
+    public abstract Builder name(@Nullable String name);
```
Guardiola31337, line 86: Do we want to accept `null` here? See above.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
```diff
@@ -0,0 +1,106 @@
...
+
+    /**
+     * Raw coordinates (longitude and latitude, order matters)
+     * that represent the routable point location.
+     *
+     * @param coordinate raw coordinates that represent the routable point location
+     * @return this builder for chaining options together
+     * @since 6.10.0
+     */
+    public abstract Builder rawCoordinate(@Nullable double[] coordinate);
```
Guardiola31337, line 96: Same as above re: Do we want to accept `null` here?

Guardiola31337 (commented):
Left some comments.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
kalbaxa: @Guardiola31337 Do you mean by duplicating e.g. the rooftop location in `RoutablePoints`? If yes, I disagree.

I assume the correct distinction would be to leave `RoutablePoints` null if there are none and let the app logic decide how to behave in that case (e.g. fallback to original location). Otherwise it's becoming ambiguous whether the data is actually available for the specific search result.

kalbaxa (approved):
When we align with @DzmitryFomchyn and @Guardiola31337 on the routable points approach, I'm good with the merge.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
DzmitryFomchyn: @Guardiola31337 in addition to what @kalbaxa said, I think it's better to leave it as nullable because all the `service-geocoder` relies on `AutoValue/GSON` serialisation/deserialisation. If we declare any field as non-nullable and for some reason backend don't send the value, the SDK will crash. I don't want this. I think it's better to check for null on customer's side and use alternative fallback. Also, nullable fields are consistent with the current approach.

Guardiola31337: @kalbaxa What I meant was that routable points could be `null` (that's totally fine) but if the server emits a routable point, I would expect not to be `null` (having `null` fields), that's the contract I'd expect from the backend.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/MapboxGeocoding.java
DzmitryFomchyn: Thanks, I've removed `@since` note everywhere in this PR

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
Guardiola31337: @DzmitryFomchyn IMO errors in the backend should fail-fast, this approach would hide this type of issues as they'd go unnoticed.

Guardiola31337:
Discussed internally, let's ship as is as `services-geocoding` is going to be deprecated soon.

cc @DzmitryFomchyn @kalbaxa

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
DzmitryFomchyn: A can agree with that, but we should also be consistent here. All the current code relies on nullability, it won't make any better if we make one field as non-null while everything else remain nullable. Given everything that's written above, I think we should leave as is now and discuss another approach in case of SDK refactoring, if it happen. More likely, `service-geocoder` will be deprecated soon.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
DzmitryFomchyn: Discussed in another thread.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
DzmitryFomchyn: Discussed in another thread.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
DzmitryFomchyn: Discussed in another thread.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
DzmitryFomchyn: Discussed in another thread.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v5/models/RoutablePoint.java
dzinad: > If we declare any field as non-nullable and for some reason backend don't send the value, the SDK will crash. I don't want this.

+++

Event if it's not an error on the backend side, it can be a logic change. For example, we accidentally made waypoints nullable but that played well because after some time they did become nullable because a case was introduced that they can be returned in a different response level.

LukasPaczos: > it can be a logic change

A logic change like this should mean a major version upgrade of an endpoint. All customers (not only Nav SDK) would be impacted by such an intentional change.

dzinad: Not necessarily. With waypoints they wouldn't (and weren't). Because it depends on a new request parameter which falls back to old behaviour.

LukasPaczos: We would have to provide backwards compatibility if that wasn't the case (either on backend or mapbox-java level). We were lucky and needed less work for that specific situation but that shouldn't be the reason to make contracts less strict going forward.

PR #1538, opened 2023-02-16
Index url path parameters from end to allow APIs to be nested under a path

dudeuter:
Currently, the `baseUrl` needs to be 'only' a base URL.

This change will allow you to add a baseUrl with a path at the end without conflicting with the path parameters.

Passes all of the route options tests.

This can potentially break functionality for users who have additional path parameters appended after the coordinates. But I doubt anyone is doing that.

The convention is that the last path in a URL is the resource. So the last path ought to always be a coordinate. Also, coordinates are the most brittle value to be parsed. In this case, if you are using a non-standard base URL, `user` and `profile` are more likely to not have as much meaning.

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/RouteOptions.java
@@ -1004,9 +1004,9 @@ public abstract class RouteOptions extends DirectionsJsonObject {
 
     try {
       String[] pathElements = url.getPath().split("/");
-      optionsJson.addProperty("user", URLDecoder.decode(pathElements[3], UTF_8));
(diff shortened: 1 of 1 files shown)


dzinad:
The following test will fail:
```
    RouteOptions routeOptions = RouteOptions.fromUrl(new URL("https://api.mapbox.com/somepath/directions/v5/mapbox/driving-traffic/1.1,2.2;3.3,4.4?overview=full"));
    assertEquals("https://api.mapbox.com/somepath/", routeOptions.baseUrl());
```
with
```
Expected :https://api.mapbox.com/somepath/
Actual   :https://api.mapbox.com
```

dudeuter:
I still think this is a good idea, but I'm closing this because I don't want to change the behavior of the API.

"Don't fix what isn't broken..." or something.

PR #1542, opened 2023-04-10
add RouteLeg#notifications

dzinad:


LukasPaczos:
Flagging that backend API design is still in review so we shouldn't merge until that's finalized.

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
RingerJK, line 1: Could you remove "shift" changes?

services-directions/src/test/java/com/mapbox/api/directions/v5/MapboxDirectionsTest.java
```diff
@@ -360,6 +360,16 @@ public void include_doesGetFormattedInUrlCorrectly() throws Exception {
     assertTrue(directions.cloneCall().request().url().toString().contains("include=hot"));
   }
 
+  @Test
+  public void notifications_doesGetFormattedInUrlCorrectly() throws Exception {
```
RingerJK, line 364: Why do we need this test? What do we test? 🤔

dzinad:
Also, this will be a breaking change for those who used notifications through unrecognized properties.
I believe this is fine since unrecognized properties are supposed to be "beta" aka "unstable and subject to change", but we should issue a warning when we release this and maybe even explicitly tag our customers who we know use it (if there are any).

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
```diff
@@ -695,25 +868,91 @@ private DirectionsCriteria() {
...
+  }
+
+
+  /**
+   * Supported notification refresh types. See {@link Notification#refreshType()}.
+   */
+  @Retention(RetentionPolicy.CLASS)
+  @StringDef({
+          NOTIFICATION_EV_STATION_OUT_OF_ORDER,
+          NOTIFICATION_REFRESH_TYPE_DYNAMIC
```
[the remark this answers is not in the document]
yuryybk: wow, fixed

services-directions-models/src/main/java/com/mapbox/api/directions/v5/DirectionsCriteria.java
yuryybk: fixed

yuryybk (approved):

PR #1577, opened 2024-04-16
Geocoding V6 Support

ahmedaly16:
Geocoding V6 is going to be GA on Monday, and it's going to be the default Geocoding API service listed under https://docs.mapbox.com/api/search/geocoding/ instead of V5.

This PR is reopening the original PR by @DzmitryFomchyn: https://github.com/mapbox/mapbox-java/pull/1550

Also I included all changes that we're aware of that happened to Geocoding V6 since the original PR was closed, I updated the tests as well.


trescube (approved):
Looks great!  Very clean, well-structured, and maintainable.

services-geocoding/src/main/java/com/mapbox/api/geocoding/v6/models/V6Context.java
```diff
@@ -0,0 +1,113 @@
...
+public abstract class V6Context extends V6JsonObject {
+
+  /**
+   * Address context element.
+   *
+   * @return address context element
+   */
+  @SerializedName("address")
+  @Nullable
+  public abstract V6ContextElement address();
```
DzmitryFomchyn, line 28: With `V6ContextElement` users will have direct access only to `id` and `name` whereas [documentation](https://docs.mapbox.com/api/search/geocoding/#the-context-object) say that 

> Address features will also include an address sub-object, which additionally contains street_name and address_number properties

I'd recommend make sure public documentation is up to date and then compare all the v6 API with what's in PR

ahmedaly16: I added this one, which wasn't something new, these fields were always there. Anyway, I found other differences between the context in the PR and in the public documentation, but it's all some fields which are missing. I spoke to Robert and we're okay with having some differences or missing features for now. We can always add that later as long as nothing is wrong now, the top priority is to have a release by end of this week.

ahmedaly16: I added this one.

PR #1585, opened 2024-07-06
Optimize Memory Usage by Interning Frequently Occurring Strings in JSON

DzmitryFomchyn:
https://mapbox.atlassian.net/browse/NAVAND-3254

In the directions response format, there are string elements that can have a limited number of values but frequently occur in the JSON response. Some of them are: `IntersectionLanes.indications()`, `IntersectionLanes.validIndication()`, `StepManeuver.type()`, and `StepManeuver.modifier()`. The GSON parser does not support string interning, which leads to new string allocations every time. Consequently, the parsed route object contains many duplicate strings. In this PR, we introduce custom TypeAdapters that intern selected strings to save memory.

Some analysis for `IntersectionLanes` objects:

- Long route from Minsk to Oslo, through Munich, Stuttgart, and Paris (4.5K km)
    - 7561 indications, 4773 valid indications, only 6 unique strings.
    - Approximately 567 MB of memory saved.
- Munich to Stuttgart (232 km)
    - 500 indications, 295 valid indications, only 5 unique strings.
    - Approximately 36 MB of memory saved.
- Short route inside Stuttgart (6 km)
    - 126 indications, 42 valid indications, only 5 unique strings.
    - Approximately 7 MB of memory saved.

### DIFF
--- a/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/IntersectionLanes.java
+++ b/services-directions-models/src/main/java/com/mapbox/api/directions/v5/models/IntersectionLanes.java
@@ -109,7 +109,7 @@ public abstract class IntersectionLanes extends DirectionsJsonObject {
    * @since 3.0.0
    */
   public static TypeAdapter<IntersectionLanes> typeAdapter(Gson gson) {
-    return new AutoValue_IntersectionLanes.GsonTypeAdapter(gson);
+    return new IntersectionLanesTypeAdapter(new AutoValue_IntersectionLanes.GsonTypeAdapter(gson));
   }
 
   /**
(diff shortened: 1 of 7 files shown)


RingerJK (approved):

dmitry-novikov (approved):

dmitry-novikov:
Is it possible to configure GSON to use `String.intern()` by default?

DzmitryFomchyn:
> Is it possible to configure GSON to use `String.intern()` by default?

Unfortunately, no, GSON doesn't support it. And actually, we don't want every string to be interned because many fields have arbitrary data (not like enum values) and it doesn't make sense to intern value that exists only once, this will degrade performance.

Ideal solution would be to choose which fields we want to intern, for example, by marking them with annotation like `@StringIntern`

```
  @StringIntern
  @SerializedName("valid_indication")
  public abstract String validIndication();
```

But it's not possible without patching GSON SDK and I'm not sure if it makes sense to do it now given our pans on changing parsing mechanism.

PR #1588, opened 2024-08-27
Adding `country` to `V6StructuredInputQuery`

ahmedaly16:
The `country` field is part of the `Forward geocoding with structured input`: https://docs.mapbox.com/api/search/geocoding#forward-geocoding-with-structured-input

So I'm adding it to our object `V6StructuredInputQuery`.

### DIFF
--- a/samples/src/main/java/com/mapbox/samples/BasicV6BatchGeocoding.java
+++ b/samples/src/main/java/com/mapbox/samples/BasicV6BatchGeocoding.java
@@ -36,11 +36,11 @@ public class BasicV6BatchGeocoding {
       .street("Pennsylvania Avenue NW")
       .postcode("20500")
       .place("Washington, DC")
+      .country("us")
       .build();
 
     final V6ForwardGeocodingRequestOptions structuredInputOptions = V6ForwardGeocodingRequestOptions
       .builder(structuredInputQuery)
-      .country("us")
       .build();
 
     final V6ReverseGeocodingRequestOptions reverseOptions = V6ReverseGeocodingRequestOptions
--- a/services-geocoding/src/main/java/com/mapbox/api/geocoding/v6/V6GeocodingService.java
+++ b/services-geocoding/src/main/java/com/mapbox/api/geocoding/v6/V6GeocodingService.java
@@ -66,7 +66,7 @@ interface V6GeocodingService {
    * @param permanent     {@link MapboxV6Geocoding#permanent()}
    * @param autocomplete  {@link V6ForwardGeocodingRequestOptions#autocomplete()}
    * @param bbox          {@link V6ForwardGeocodingRequestOptions#bbox()}
-   * @param country       {@link V6ForwardGeocodingRequestOptions#country()}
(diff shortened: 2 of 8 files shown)


ahmedaly16:
@DzmitryFomchyn do you recall if the `country` field was left out of `V6StructuredInputQuery` intentionally? Or was it just missed?

samples/src/main/java/com/mapbox/samples/BasicV6BatchGeocoding.java
```diff
@@ -36,11 +36,11 @@ private static List<V6RequestOptions> requestOptions() {
       .street("Pennsylvania Avenue NW")
       .postcode("20500")
       .place("Washington, DC")
+      .country("us")
       .build();
 
     final V6ForwardGeocodingRequestOptions structuredInputOptions = V6ForwardGeocodingRequestOptions
       .builder(structuredInputQuery)
-      .country("us")
```
DzmitryFomchyn, line 43: Country in `V6ForwardGeocodingRequestOptions` is still supported, right? So you can leave country in this example

services-geocoding/src/main/java/com/mapbox/api/geocoding/v6/V6GeocodingService.java
```diff
@@ -66,7 +66,7 @@ Call<V6Response> forwardGeocoding(
    * @param permanent     {@link MapboxV6Geocoding#permanent()}
    * @param autocomplete  {@link V6ForwardGeocodingRequestOptions#autocomplete()}
    * @param bbox          {@link V6ForwardGeocodingRequestOptions#bbox()}
-   * @param country       {@link V6ForwardGeocodingRequestOptions#country()}
+   * @param country       {@link V6StructuredInputQuery#country()}
```
DzmitryFomchyn, line 69: But effectively country is still taken from `V6ForwardGeocodingRequestOptions`. Is it expected?

ahmedaly16: I assume you mean this: https://github.com/mapbox/mapbox-java/blob/05a7bfe9b3313f4c5d3a2a6f2c77922c97efc171/services-geocoding/src/main/java/com/mapbox/api/geocoding/v6/MapboxV6Geocoding.java#L87-L107

But isn't everything there taken from `V6ForwardGeocodingRequestOptions`? Why does it say other fields above taken from `V6StructuredInputQuery` like `V6StructuredInputQuery#neighborhood()`?

I guess because it assumes the `V6ForwardGeocodingRequestOptions` will be created by the builder which takes `V6StructuredInputQuery`.

If the right thing is to revert this line I changed, should we also change the others above?

DzmitryFomchyn: Ah, ok, I didn't notice at the first that it's a function dedicated to structure input `structureInputForwardGeocoding()`. I think we can leave as is.

DzmitryFomchyn (approved):

PR #1622, opened 2025-08-28
Expose step intersection `duration` as Double

jush:
Expose `StepIntersection#duration` (previously available through `StepIntersection#getUnrecognizedProperty("duration")`)


RingerJK (commented):
I suppose it should also be exposed `distance` filed

RingerJK (approved):

